PR #3, opened 2016-01-20
LocalMedia#removeStream() and LocalMedia#removeTrack() do not properly stop and remove tracks

kainosnoema:
Because `LocalMedia#removeTrack()` is [simply calling `Media#_removeTrack()`](https://github.com/twilio/twilio-conversations.js/blob/7259a9b4066a63c2c12890f5b75910cc72fdd565/lib/media/localmedia.js#L69-L78), it doesn't properly stop and remove the track from the media stream, as is being done correctly in `LocalMedia#_removeTrack()`.

My suggestion is to remove the current definition of `LocalMedia#removeTrack()`, and replace it with the current definition of `LocalMedia#_removeTrack()`, [which overrides the underlying `Media#_removeTrack()` properly](https://github.com/twilio/twilio-conversations.js/blob/7259a9b4066a63c2c12890f5b75910cc72fdd565/lib/media/localmedia.js#L208-L220). This will by side-affect fix the behavior of `LocalMedia#removeStream()`, which calls into the current broken definition. There may be some other methods that are correctly calling the current `LocalMedia#removeTrack()` though, and those would need to be changed to call the underlying `Media#_removeTrack()` instead.

### DIFF
--- a/lib/media/localmedia.js
+++ b/lib/media/localmedia.js
@@ -109,7 +109,7 @@ LocalMedia.prototype.removeMicrophone = function removeMicrophone(stop) {
   if (microphone) {
     return this.removeTrack(microphone, stop);
   }
-  return null;
+  return microphone;
 };
 
 /**
@@ -155,7 +155,7 @@ LocalMedia.prototype.removeCamera = function removeCamera(stop) {
   if (camera) {
     return this.removeTrack(camera, stop);
   }
-  return null;
+  return camera;
 };
 
 /**
--- a/lib/media/localmedia.js
+++ b/lib/media/localmedia.js
@@ -67,17 +67,6 @@ inherits(LocalMedia, Media);
 LocalMedia.prototype.addTrack = Media.prototype._addTrack;
 
 /**
- * Removes a {@link LocalTrack} from the {@link LocalMedia} object, if it was added.
- * @method
- * @param {LocalTrack} track - The {@link LocalTrack} to remove
- * @param {?boolean} [stop=true] - Whether or not to call
- *   {@link LocalTrack#stop}
- * @returns {this}
- * @fires Media#trackRemoved
- */
-LocalMedia.prototype.removeTrack = Media.prototype._removeTrack;
-
-/**
  * Adds a {@link LocalAudioTrack} representing your browser's microphone to the
  * {@link LocalMedia} object, if not already added.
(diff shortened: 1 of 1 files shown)


kainosnoema:
The build in Travis seems to be failing because of missing environment variables that are no longer being exported (`ACCOUNT_SID`, etc).

markandrus:
Hi @kainosnoema, thanks for the report! I should have some time to look at it tomorrow. Regarding the Travis build, it is using `SIGNING_KEY_SECRET` which is "secret", so Travis will not include them on PR builds in order not to leak the credential. I'll try building your branch manually when I review it.

markandrus:
@kainosnoema I've reviewed your change and it looks good. It also suggests to me that it might be a good idea to support the optional `stop` parameter that `removeStream`/`removeTrack` accept in both `removeCamera` and `removeMicrophone`.

kainosnoema:
@markandrus great thought. I've pushed two commits: one adding the optional `stop` param to `removeMicrophone`/`removeCamera`, and another to return the tracks when removed (as is mentioned in the docs already).

I haven't extensively tested these changes. I'd be happy to add unit tests, but there doesn't seem to be much existing to go off of. :grimacing:

markandrus:
+1, thanks @kainosnoema! I've built and tested this locally, and also identified a bug with `isEnded` that I'll take up in a separate PR.

> I'd be happy to add unit tests, but there doesn't seem to be much existing to go off of. :grimacing:

You are right. We may want to build out the WebRTC mocks so that we can write some media unit tests.

kainosnoema:
@markandrus sure thing.

By the way, you should be able to add encrypted environment variables to the TravisCI project and hide them from the build output so they're not leaked. Having passing builds/badge on the project would be really nice.

markandrus:
@kainosnoema thanks for the tip—I will have to investigate that. I believe it may still be unsafe, because what if some code in a pull request explicitly calls `console.log(process.env.SIGNING_KEY_SECRET)` or similar? Are you familiar with a workaround for this?

Perhaps I could rewrite the `gulp` task to simply skip integration tests on pull requests...

kainosnoema:
@markandrus no, you're right. It's still a risk. Disabling integration tests in Travis sounds good, as long as we can flesh out the unit tests. :)

PR #6, opened 2016-02-15
Callstats api integration

karthikbr82:
callstats.io monitors and manages the performance of video calls in an WebRTC application. 
This PR integrates the callstats APIs in twilio-conversations.js

Pending -

reportError has to be integrated for catching the getUserMediaError - in the file lib/media/localmedia.js
(For integration conferenceID is necessary)
The fabric events such as mute/unmute pause/play has to be integrated as well.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -66,6 +66,10 @@ function Client(accessManager, options) {
   var UserAgent = options.userAgent;
   var userAgent = new UserAgent(accessManager, options);
 
+  /* global callStats appId appSecret csInitCallback */
+  /* eslint no-undef: 2 */
+  callStats.initialize(appId, appSecret, accessManager.identity, csInitCallback, null, null);
+
   userAgent.on('invite', function onInviteServerTransaction(inviteServerTransaction) {
     var cookie = inviteServerTransaction.cookie;
     var outgoingInvite = outgoingInvites.get(cookie) || canceledOutgoingInvites.get(cookie);
--- a/lib/client.js
+++ b/lib/client.js
@@ -68,7 +68,7 @@ function Client(accessManager, options) {
 
   /* global callStats appId appSecret csInitCallback */
   /* eslint no-undef: 2 */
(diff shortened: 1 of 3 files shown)


vr000m:
@rfbrazier Attached is the corresponding screenshot of the dashboard. To recreate the integration, 
1.  need an appID and AppSecret from callstats.io, and 
2. run the twilio video test app

fireshot capture 6 - apps callstats io_ - https___dashtest callstats io_search

cummack:
Hi Callstats folks,

Thanks for the PR. You've given us some great ideas on how we can enable developers to build this kind of third-party integration using only public APIs from our SDK.  I'm going to close this PR for now, but we'll be in touch via email to get some input on the proposed APIs mentioned.

jamesgraham:
@cummack did anything happen with those public APIs or are we forced to pay for Insights or similar?

markandrus:
@jamesgraham the APIs have been on the back-burner, unfortunately. Regarding Insights, I don't think we plan to lock anyone in—gathering stats, listening to ICE events, etc., should all be achievable client-side, without Insights.

There are a few questions about what such generic APIs should look like, e.g.

* Should we actually bubble up events like "peerConnectionCreated", "createOffer", "setLocalDescription", etc.
* Should we make it easy to provide a PeerConnection "subclass" that you could instrument to listen for the events you care about?

Are you interested in the APIs in particular, or specifically callstats.io integration?

Best,
Mark

jamesgraham:
Hi Mark,

I noted the events are part of the getStats spec. It would be great to have some events fired that we can listen to and notify the user of issues -for example high latency during a session- without delving into that spec. Programmable Voice have these but I think they are only fired if you have Insights enabled (I could be wrong about that).

We would also log these for our own statistics and customer support purposes. Either implementation looks ok but I would always strive for consistency with your other SDKs (Voice as mentioned above).


Thanks!

PR #12, opened 2016-06-15
rename function to prevent a `strict mode` error

ksmth:
The `conversationInfo` function is shadowed by the `conversationInfo` parameter which results in a SyntaxError.


markandrus:
yikes! thank you @ksmth!

markandrus:
This is now released in twilio-conversations.js 0.13.7

PR #36, opened 2016-11-27
addCamera and addMicrophone allow constraints

mattapperson:
This commit allows a constraints object to be passed in as the first param to allow specific cameras or mics to be selected.

### DIFF
--- a/lib/media/localmedia.js
+++ b/lib/media/localmedia.js
@@ -130,10 +130,11 @@ LocalMedia.prototype.addTrack = function addTrack(track) {
  * @returns {Promise<LocalAudioTrack>}
  * @fires Media#trackAdded
  */
-LocalMedia.prototype.addMicrophone = function addMicrophone() {
+LocalMedia.prototype.addMicrophone = function addMicrophone(audioConstraints) {
   var self = this;
   var microphone = null;
   var log = this._log;
+  audioConstraints = audioConstraints || true;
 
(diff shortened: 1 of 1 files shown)


ZuSe:
It should also work with all webrtc-contraint attributes, shouldn't it?

mattapperson:
It should, yes. I'm using it to switch between shared desktop and a camera in an electron app

ZuSe:
Ok. What I miss atm are the resolution and frame rate constraints that were possible to put in when initializing the localMedia object with the conversations api. This option should be back with your commit :)

mattapperson:
yup 😊

mattapperson:
@markandrus is there a CLA I need to sign?

markandrus:
@mattapperson thanks for the PR! There's no CLA to sign (I'm not sure if we need one or not...). The team will take a look at your changes later today.

ryan-rowland:
LGTM 👍 

Not an action item for this PR, but seeing this change duplicated for audio & video prompted me to compare `addCamera` and `addMicrophone`. Since these are essentially doing the same thing for two different types of tracks, maybe we could consider DRYing it up. @markandrus

mattapperson:
@ryan-rowland I was just thinking the same when creating this PR

manjeshbhargav:
LGTM

mattapperson:
Woot! Thanks guys!

PR #58, opened 2017-02-23
wait after OPTIONS check; add per-Framework timeouts

markandrus:
@manjeshbhargav this PR adds

* Per-Framework Test timeouts (since we've since Meteor can take much longer),
* Waits between HTTP OPTIONS checks (in case we are spamming the server and slowing down its startup), and
* Prints stdout from the Framework Test "start" command (so that we can see if something is going wrong).

### DIFF
--- a/test/framework/twilio-video-no-framework/app.js
+++ b/test/framework/twilio-video-no-framework/app.js
@@ -19,7 +19,7 @@ function getQueryParameters(location) {
 
 const token = (getQueryParameters(location).get('token') || [])[0] || '';
 
-new Twilio.Video.Client().connect({ token: token }).then(room => {
+new Twilio.Video.connect({ token: token }).then(room => {
   root.innerHTML = `<p>Connected to Room ${room.sid}.</p>`;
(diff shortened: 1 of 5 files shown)


test/framework/twilio-video-no-framework/app.js
```diff
@@ -19,7 +19,7 @@ function getQueryParameters(location) {
 
 const token = (getQueryParameters(location).get('token') || [])[0] || '';
 
-new Twilio.Video.Client().connect({ token: token }).then(room => {
+new Twilio.Video.connect({ token: token }).then(room => {
```
manjeshbhargav: @markandrus This might conflict with the same change that I made in `client:media-api-simplification`.

manjeshbhargav:
@markandrus +1. Just one comment about conflict.

test/framework/twilio-video-no-framework/app.js
markandrus: We can rebase that branch as necessary

PR #70, opened 2017-03-30
Fix incorrect parameters being passed to LocalTrack

ElliotChong:
The API for LocalTrack is `LocalTrack(Track, mediaStreamTrack, options)`.

The current `LocalVideoTrack` implementation passes a `mediaStream` object instead of `mediaStreamTrack` as the 2nd parameter which causes an error to be thrown.

### DIFF
--- a/lib/media/track/localvideotrack.js
+++ b/lib/media/track/localvideotrack.js
@@ -18,7 +18,7 @@ function LocalVideoTrack(mediaStream, mediaStreamTrack, options) {
   if (!(this instanceof LocalVideoTrack)) {
     return new LocalVideoTrack(mediaStream, mediaStreamTrack, options);
   }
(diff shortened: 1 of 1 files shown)


manjeshbhargav:
@ElliotChong Thanks for bringing this to our notice. We actually have the fix for this lined up for `beta6`.

lib/media/track/localvideotrack.js
```diff
@@ -18,7 +18,7 @@ function LocalVideoTrack(mediaStream, mediaStreamTrack, options) {
   if (!(this instanceof LocalVideoTrack)) {
```
manjeshbhargav: @ElliotChong `mediaStream` should also be removed from the constructor of `LocalVideoTrack`.

ElliotChong:
Thanks @manjeshbhargav, looking forward to `beta6`!

PR #76, opened 2017-04-05
Stop LocalTracks on disconnect

### LINKED ISSUE #75: beta 5 - removetrack does not close webcam
Hi,
I'm trying to stop local audio/video(not to disconnect from a room but just to stop the local audio/video).
I tried the following code and it does stop sending the audio/video, the problem is that the webcam light is still on, like it is still broadcasting the video.

    var tracks = Array.from(room.localParticipant.tracks.values());
    tracks.forEach(function(track) {
        room.localParticipant.removeTrack(track);
    });

markandrus:
@manjeshbhargav @ryan-rowland This behavior got lost in the transition from 1.0.0-beta4 to 1.0.0-beta5. This commit adds it back with some unit tests to prevent future regression. See [here](https://github.com/twilio/twilio-video.js/blob/1.0.0-beta4/lib/room.js#L302) how it worked before. Should fix #75.

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -26,8 +26,13 @@ function LocalParticipant(signaling, localTracks, options) {
   options = Object.assign({
     LocalAudioTrack: LocalAudioTrack,
     LocalVideoTrack: LocalVideoTrack,
+    shouldStopLocalTracks: false,
     tracks: localTracks
   }, options);
+
+  var tracksToStop = options.shouldStopLocalTracks
+    ? new Set(localTracks)
+    : new Set();
 
   Participant.call(this, signaling, options);
   Object.defineProperties(this, {
@@ -36,6 +41,9 @@ function LocalParticipant(signaling, localTracks, options) {
     },
     _LocalVideoTrack: {
       value: options.LocalVideoTrack
+    },
+    _tracksToStop: {
+      value: tracksToStop
     }
   });
 }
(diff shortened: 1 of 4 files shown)


lib/localparticipant.js
```diff
@@ -92,6 +100,14 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
       signaling.removeListener('stateChanged', stateChanged);
       self.removeListener('trackAdded', localTrackAdded);
       self.removeListener('trackRemoved', localTrackRemoved);
+
+      if (self._tracksToStop.size) {
```
manjeshbhargav: @markandrus I don't think we need to check for `size` here:
```javascript
log.info('LocalParticipant disconnected.');
self._tracksToStop.forEach(track => {
  log.debug('Stopping automatically-acquired LocalTrack:', track);
  track.stop();
});
```

markandrus: I think a single log message if there are any LocalTracks to stop might be nicer. Each call to `LocalTrack#stop` is going to log, anyway. So the difference is between

```
[LocalParticipant] LocalParticipant disconnected. Stopping automatically-acquired LocalTracks
[LocalTrack 1] Stopping
[LocalTrack 2] Stopping
```

And

```
[LocalParticipant] LocalParticipant disconnected.
[LocalParticipant] Stopping automatically-acquired LocalTrack: [LocalTrack 1]
[LocalTrack 1] Stopping
[LocalParticipant] Stopping automatically-acquired LocalTrack: [LocalTrack 2]
[LocalTrack 2] Stopping
```

WDYT?

ryan-rowland:
Not blocking on the PR, and I know we just came off another similar discussion, so we can table for later. But after seeing it for a while and mulling it over, I’m not sure I like how we’re passing the same options object throughout all of our components. We’re passing it through opaquely, so the reader has no idea by looking at where its passed in which of its properties the consuming constructor is looking for / going to read.

lib/localparticipant.js
manjeshbhargav: @markandrus Then we can do this:
```javascript
log.info('LocalParticipant disconnected.' + self._tracksToStop.size ?
  ' Stopping automatically-acquired LocalTracks.' ? '');
self._tracksToStop.forEach(track => track.stop());
```

PR #83, opened 2017-04-13
Calling new LocalTrack errors out

randallb:
Without an options object, this errors out.


randallb:
I'm not sure if this is the best way to fix, but it's one way to fix. :)

markandrus:
Thanks @randallb. @manjeshbhargav ran with this and landed it in the recent 1.0.0-beta6 release.

PR #86, opened 2017-04-18
Spawn Karma for each integration test module

markandrus:
@manjeshbhargav

### DIFF
--- a/scripts/karma.js
+++ b/scripts/karma.js
@@ -1,0 +1,46 @@
+#!/usr/bin/env node
+'use strict';
+
+const { readdirSync, statSync } = require('fs');
+const { Server, stopper } = require('karma');
+const { parseConfig } = require('karma').config;
+const { join } = require('path');
+
+const configFile = join(__dirname, '..', 'integration.conf.js');
+const integrationTests = join(__dirname, '..', 'test', 'integration', 'spec');
+
+
+function getTestPaths(path) {
+  var stat = statSync(path);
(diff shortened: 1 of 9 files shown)


scripts/karma.js
```diff
@@ -0,0 +1,39 @@
...
+
+const configFile = join(__dirname, '..', 'integration.conf.js');
+const integrationTests = join(__dirname, '..', 'test', 'integration', 'spec');
+
+// NOTE(mroberts): We have a memory leak, either in twilio-video.js or in
+// Firefox, that causes Firefox to slow down after running a bunch of tests that
+// exercise WebRTC APIs. To workaround this, we spawn Karma for each integration
+// test module.
+async function main() {
+  for (const file of [
```
manjeshbhargav: @markandrus Can this list be generated from the folder structure using `require('fs').readDir`?

manjeshbhargav:
@markandrus Just one comment

scripts/karma.js
manjeshbhargav: @markandrus Did some research, it's fairly straightforward:
```javascript
const fs = require('fs');

function getTestPaths(path) {
  var stat = fs.statSync(path);
  if (stat && stat.isDirectory()) {
    return fs.readdirSync(path).reduce((files, file) => {
      return files.concat(testgetTestPaths(`${path}/${file}`));
    }
  }
  return [path];
}

const files = getTestPaths('test/integration/spec').filter(path => !/webrtc/.test(path));
```

PR #90, opened 2017-04-25
Setup travis-multirunner under node_modules

markandrus:
@manjeshbhargav travis-multirunner's `setup.sh` drops a Firefox tar.bz2 as well as the `browsers/` directory in wherever it is called from. This causes problems with our release-tool script for publishing to NPM. This change simply cds into `node_modules/travis-multirunner` before calling the `setup.sh` script (that way we don't pollute the working directory).


manjeshbhargav:
+1

PR #139, opened 2017-07-21
Workaround adapter.js MSID behavior

markandrus:
This sucks, but let's see if it works. adapter.js has shimmed `addTrack`/`addStream` in a way that changes MediaStream IDs. We can try to workaround its shims by getting access to a "fresh" RTCPeerConnection from an iframe's `contentWindow`. EDIT: We can also re-enable Chrome Beta in `.travis.yml` this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,6 @@ env:
     - DBUS_SESSION_BUS_ADDRESS=/dev/null
     - DISPLAY=:99.0
     - PATH=$HOME/.meteor:$PATH
-matrix:
-  allow_failures:
-    - env: BROWSER=chrome BVER=beta
 before_script:
   - cd node_modules/travis-multirunner
   - BROWSER=chrome ./setup.sh
--- a/lib/webrtc/rtcpeerconnection/chrome.js
+++ b/lib/webrtc/rtcpeerconnection/chrome.js
@@ -12,6 +12,22 @@ var PeerConnection = typeof RTCPeerConnection !== 'undefined'
   ? RTCPeerConnection
   : webkitRTCPeerConnection;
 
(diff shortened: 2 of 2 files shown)


anand-io:
Does this mean, it is still not recommended to use adapter.js along with twilio-video.js?

markandrus:
@anand-io I would not recommend using adapter.js unless your application needs to handle the WebRTC APIs directly (even then, you may not need it). The issue is that adapter.js will override the browser's WebRTC APIs, potentially introducing bugs or behavior changes. I'm not trying to knock adapter.js here—I think its goals are noble, it's just all software introduces risk of bugs...

Anyway, if there _are_ issues with adapter.js's changes, it's hard for libraries like twilio-video.js to get access to the native browser APIs without using hacks like the &lt;iframe&gt; trick in this PR.

To give a little more context on this PR, I noticed [an issue with adapter.js](https://github.com/webrtc/adapter/issues/634), caught by our [RTCPeerConnection tests](https://github.com/twilio/twilio-video.js/blob/master/test/integration/spec/webrtc/rtcpeerconnection.js) (which we run both with and without adapter.js). There is [a PR open on adapter.js to fix this](https://github.com/webrtc/adapter/pull/637). If they land and release that fix soon, I'd rather close this PR.

In practice, until the fix in adapter.js lands, if you have two or more Chrome Participants in a P2P Room and add, remove, and add a Track again, this could cause other Tracks to stop.

anand-io:
@markandrus  thanks for the info. As it was removed from COMMON_ISSUES.md, I thought it was safe to use. 
I was using it for getUserMedia, but I think we don't need to use adapter.js for that anymore as  ```mediaDevices.getUserMedia``` and ```MediaDevices.enumerateDevices``` become standard.

markandrus:
adapter.js has been updated.

Issue #140, opened 2017-07-25
Detached Participant audio can still be heard


ghost:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [ ] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.

**Code to reproduce the issue:**
```js
participantOnOffVolume (instanceId) {
    if (document.getElementById(`participant-${instanceId}-micToggle`).querySelector('#buttonVolume').classList.contains('zmdi-volume-off')) this.attachTracks(Array.from(this.videos[instanceId].audioTracks.values()), document.getElementById(`participant-${instanceId}-video`), this.videos[instanceId]);
    else this.detachTracks(Array.from(this.videos[instanceId].audioTracks.values()), this.videos[instanceId]);
  }

detachTracks (tracks, participant) {
    tracks.forEach((track) => {
      track.detach()
        .forEach((detachedElement) => {
          detachedElement.remove();
          this.adjustButtons(participant._instanceId);
        });
    });
  }

```

**Expected behavior:**
Functionality is meant to mute(detach) or unmute(reattach) userA's audio track on userB's browser during a group call.  Upon pressing the mute audio button, userA's audio track should no longer be heard after detached and removed.

**Actual behavior:**
UserA's audio track can be heard by userB even after detached and removed.  HTML has been inspected and found no audio track but still heard audio.

**Software versions:**

 - [x] Browser(s): Chrome 59.0.3071.115
 - [x] Operating System: Win10
 - [x] twilio-video.js: 1.2.0
 - [x] Third-party libraries (e.g., Angular, React, etc.): Angular 1.6.5

Samenai:
Video: https://streamable.com/xlwpk

UserB has camera and mic turned off, userA has camera turned off.  You can hear userA's audio track (music I played on a bluetooth speaker and my typing) even though the track was detached and removed.

Samenai:
Quick workaround:

```js
participantOnOffVolume (instanceId) {
[...]

PR #141, opened 2017-07-28
Fix detach behavior in Chrome (and Safari 11 Tech Preview)

markandrus:
Currently, in Chrome, it appears that an HTMLMediaElement's `srcObject` setter takes a "snapshot" of the MediaStream's MediaStreamTracks at the time at which it is set. This means, calls to `removeTrack` don't effect playback. Stranger, if a MediaStreamTrack is removed from the `srcObject`, and then the HTMLMediaElement is detached from the DOM, audio playback will continue. This bug is captured [here](https://bugs.chromium.org/p/chromium/issues/detail?id=749928).

A sufficient workaround is to reset the `srcObject` attribute after making changes to the MediaStream. So this PR introduces that change to `detach`. It's worth mentioning that this `srcObject` behavior also effects _adding_ MediaStreamTracks to a `srcObject`, as reported [here](https://bugs.chromium.org/p/chromium/issues/detail?id=720258); however, we've "accidentally" been working around this by always resetting the `srcObject` attribute in `attach`.

Fixes https://github.com/twilio/twilio-video.js/issues/140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Bug Fixes
 
 - In Firefox, we were raising a `peerIdentity` TypeError in the console.
   (JSDK-1372)
+- `detach`ing a Track in Chrome could result in that Track continuing to
+  playback, even if the &lt;audio&gt; or &lt;video&gt; elements it was
+  `attach`ed to were removed from the DOM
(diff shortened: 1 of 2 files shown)


manjeshbhargav:
+1

PR #142, opened 2017-08-01
fix RTCSessionDescription and RTCPeerConnection not defined errors on…

mmaloon:
when using twilio-video on Firefox, you get a 'RTCSessinDescription not defined' or 'RTCPeerConnection not defined' error in the console.  This fixes that by prefering the 'mox' prefixed versions of those interfaces

### DIFF
--- a/lib/webrtc/rtcpeerconnection/firefox.js
+++ b/lib/webrtc/rtcpeerconnection/firefox.js
@@ -8,7 +8,7 @@ var MediaStream = require('../mediastream');
 var updateTracksToSSRCs = require('../../util/sdp').updateUnifiedPlanTrackIdsToSSRCs;
 var util = require('../../util');
 
-var PeerConnection = RTCPeerConnection || mozRTCPeerConnection;
+var PeerConnection = mozRTCPeerConnection || RTCPeerConnection;
 
 // NOTE(mroberts): This class wraps Firefox's RTCPeerConnection implementation.
 // It provides some functionality not currently present in Firefox, namely the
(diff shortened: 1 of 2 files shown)


markandrus:
Hi @mmaloon,

Thanks for the PR. Can you please share what version of Firefox you are using?

I think the checks should be tweaked a little bit, e.g.

```js
var PeerConnection = typeof RTCPeerConnection !== 'undefined' ? RTCPeerConnection : mozRTCPeerConnection;
```

```js
module.exports = typeof RTCSessionDescription !== 'undefined' ? RTCSessionDescription : mozRTCSessionDescription;
```

i.e., we should still prefer the unprefixed versions and fallback to the `moz`-prefixed versions (using the `moz`-prefixed versions will raise deprecation warnings in newer Firefox versions).

Thanks,
Mark

mmaloon:
Hi Mark,

Of course right after I made the pull request I realized that I was using pretty old version (~40) of Firefox.  Once I updated the code worked fine with the original code.

The reason I didn’t include those checks is because those same checks are made it the respective ‘index.js’ files which include that browser specific version of the interface, however I understand your point and fully agree.  Perhaps the webrtc/rtcsessiondescription/index.js and webrtc/rtcpeerconnection.index.js files should have the order of their check changed to start with the unprefixed version, then fall back to the prefixed ones.

Thanks for the prompt reply, I look forward to integrating Twilio’s video API into our website (https://wizedemy.com <https://wizedemy.com/> - launching end of month)

Regards,
Mike

PR #148, opened 2017-08-10
1.2.1-rc1 Fixes

markandrus:
@manjeshbhargav would you take a look? As you mentioned earlier, the "dummy" MediaStreamTrack shows up in `getStats`. We'll need to fix that before releasing; however, Deniss also mentioned seeing the dummy MediaStreamTrack in recordings, too. There may be some more work to do here, so here is my proposal: leave this in, but disabled. We can iterate on it in future sprints and try, e.g., filtering the MSID from the SDP, filtering the result of `getStats`, etc. Would like to get these in for 1.2.1-rc2.

### DIFF
--- a/lib/signaling/v2/peerconnectionmanager.js
+++ b/lib/signaling/v2/peerconnectionmanager.js
@@ -1,11 +1,12 @@
 'use strict';
 
-var AudioContextFactory = require('../../webaudio/audiocontext');
 var inherits = require('util').inherits;
 var PeerConnectionV2 = require('./peerconnection');
 var MediaStream = require('../../webrtc/mediastream');
 var QueueingEventEmitter = require('../../queueingeventemitter');
 var util = require('../../util');
+
+// var isFirefox = util.guessBrowser() === 'firefox';
 
 /**
  * Construct {@link PeerConnectionManager}.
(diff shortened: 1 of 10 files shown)


lib/webaudio/audiocontext.js
```diff
@@ -1,16 +1,79 @@
...
-  NativeAudioContext = webkitAudioContext;
+/**
+ * {@link AudioContextFactory} ensures we construct at most one AudioContext
+ * at a time, and that it is eventually closed when we no longer need it.
+ * @param {AudioContextFactoryOptions} [options]
+ * @property {AudioContextFactory} AudioContextFactory - The
+ *   {@link AudioContextFactory} constructor
+ */
+function AudioContextFactory(options) {
+  var NativeAudioContext = options && options.AudioContext
```
manjeshbhargav: @markandrus Nitpick: I would do this outside the constructor.

markandrus: If we move the binding `NativeAudioContext` out, what would you want to name the binding inside the constructor? Because we still need to choose between the native `AudioContext` and the one passed in via `options`

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -25,12 +26,31 @@ function PeerConnectionManager(iceServerSource, options) {
   QueueingEventEmitter.call(this);
 
   options = Object.assign({
-    audioContext: audioContext,
+    AudioContextFactory: require('../../webaudio/audiocontext'),
```
manjeshbhargav: @markandrus Nitpick: Since this is a singleton instance, can we camel-case the name to `audioContextFactory`.

markandrus: Good idea 👍

lib/webaudio/audiocontext.js
manjeshbhargav: @markandrus I would do something like this:
```javascript
var NativeAudioContext = typeof AudioContext !== 'undefined'
  ? AudioContext
  : typeof webkitAudioContext !== 'undefined'
    ? webkitAudioContext
    : null;

function AudioContextFactory(options) {
  options = Object.assign({
    AudioContext: NativeAudioContext
  }, options);

    _AudioContext: {
        value: options.AudioContext
    }
  });
}
```

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -25,12 +26,31 @@ function PeerConnectionManager(iceServerSource, options) {
...
     MediaStream: MediaStream,
     PeerConnectionV2: PeerConnectionV2
   }, options);
 
+  // NOTE(mroberts): By default, only construct an AudioContext (and therefore
+  // trigger the "dummy" MediaStreamTrack behavior) in Firefox.
+  var audioContext = options.audioContext ? options.audioContext : null;
```
manjeshbhargav: @markandrus Shouldn't this be `var audioContext = options.AudioContextFactory.getOrCreate(this);` since you've converted `audiocontext.js` to a factory ?

markandrus: This still allows for `audioContext` to be provided as an option; however, perhaps that's not useful. Will remove.

manjeshbhargav:
@markandrus Looks good. I've posted a couple of questions. Also, I think we should update the `PeerConnectionManager` unit test to account for the refactored `AudioContextFactory`.

manjeshbhargav:
+1

PR #151, opened 2017-08-17
JSDK-1449 Removing WebRTC shims

manjeshbhargav:
@mroberts @rrowland

Removing webrtc shims from the project.
Adding '' as a dependency.


markandrus:
@manjeshbhargav LGTM. Let's actually publish @twilio/webrtc 1.0.0 before merging.

lib/signaling/v2/peerconnection.js
```diff
@@ -634,6 +634,25 @@ function getConfiguration(configuration) {
...
+ * @returns {Array<string>} mediaSections
+ */
+function getMediaSections(sdp, kind, direction) {
+  kind = kind || '.*';
+  direction = direction || '.*';
+  return sdp.split('\r\nm=').slice(1).map(function(mediaSection) {
+    return 'm=' + mediaSection;
+  }).filter(function(mediaSection) {
+    var kindPattern = new RegExp('m=' + kind, 'gm');
+    var directionPattern = new RegExp('a=' + direction, 'gm');
```
ryan-rowland: Can these be done outside of the loop?

manjeshbhargav: @ryan-rowland No, because we have to filter out only those `m=` sections that match the `kind` and `direction`. Doing it outside the loop just tells you whether the SDP has at least one `m=` section of the given `kind` and `direction`.

ryan-rowland: Not sure I follow. `kind` and `direction` don't change between iterations, do they?. Couldn't we change the lines:
```
kind = kind || '.*';
direction = direction || '.*';
```

to the following?

```
var kindPattern = new RegExp('m=' + (kind || '.*'), 'gm');
var directionPattern = new RegExp('a=' + (direction || '.*'), 'gm');
```

manjeshbhargav: @ryan-rowland I get it now. You mean moving those regexes out of the loop?

ryan-rowland: Yep, otherwise we're creating them multiple times when they only need to be created once.

manjeshbhargav: @ryan-rowland I agree. I'll make a note to refactor this. But for now I don't want to introduce code changes to this function because I would also have to update the corresponding method in the webrtc shim project and submit a PR for that as well. I will do this as part of a future PR.

ryan-rowland: Ah, I see it was moved. Hard to tell with all the red in the PR.

manjeshbhargav:
Closing this and opening the same PR between two branches on `twilio/twilio-webrtc`: https://github.com/twilio/twilio-video.js/pull/152.

PR #157, opened 2017-08-22
Disable Node integration tests

markandrus:
@manjeshbhargav with the new Group Rooms deploy, we're actually checking SDPs on the backend, so our "fake SDPs" we've been sending around in our Node-based integration tests won't work anymore. Since we already re-run these same tests as part of our browser-based integration tests, I think we can disable the Node-based ones.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,11 +62,9 @@
   "scripts": {
     "lint": "eslint ./lib",
     "test:unit": "mocha ./test/unit/index.js",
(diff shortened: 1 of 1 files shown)


manjeshbhargav:
+1

manjeshbhargav:
@markandrus Sorry about that! Clicked on "close and comment" by mistake.

PR #159, opened 2017-08-24
Update CHANGELOG.md

markandrus:
@manjeshbhargav @ktoraskartwilio @rfbrazier


markandrus:
I'm going to merge for now so we can get an RC out, but let's revisit before launch.

PR #166, opened 2017-09-11
JSDK-1505 Modifying connect codec+bitrate test to work with SFU

manjeshbhargav:
@markandrus 

The "connect" fixed bitrate codec + bitrate test was failing in SFU mode as the codec map did not contain "pcma" and "pcmu". So the modified test passing condition is:

The test should pass either if the selected codec is not in the fixed payload types, or if it is, then the b= line should not be present.

### DIFF
--- a/test/integration/spec/connect.js
+++ b/test/integration/spec/connect.js
@@ -419,11 +419,11 @@ describe('connect', function() {
         return getMediaSections(pc.remoteDescription.sdp);
       }).forEach(section => {
         const codecMap = createCodecMap(section);
-        const expectedCodecIds = /m=audio/.test(section)
+        const expectedPayloadTypes = /m=audio/.test(section)
           ? flatMap(testOptions.preferredAudioCodecs, codecName => codecMap.get(codecName.toLowerCase()) || [])
           : flatMap(testOptions.preferredVideoCodecs, codecName => codecMap.get(codecName.toLowerCase()) || []);
-        const actualCodecIds = getCodecIds(section);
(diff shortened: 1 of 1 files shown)


test/integration/spec/connect.js
```diff
@@ -453,7 +453,14 @@ describe('connect', function() {
         assert(pc.remoteDescription.sdp);
         return getMediaSections(pc.remoteDescription.sdp, 'audio');
       }).forEach(section => {
-        assert(!/b=(AS|TIAS):([0-9]+)/.test(section));
+        const codecMap = createCodecMap(section);
+        const codecIds = getCodecIds(section);
```
markandrus: I thought we started referring to these as payload types everywhere?

test/integration/spec/connect.js
```diff
@@ -453,7 +453,14 @@ describe('connect', function() {
...
+        const fixedBitrateCodecIds = new Set([
+          ...(codecMap.get('pcma') || []),
+          ...(codecMap.get('pcmu') || [])
+        ]);
+        assert(!fixedBitrateCodecIds.has(codecIds[0])
+          || !/b=(AS|TIAS):([0-9]+)/.test(section));
```
markandrus: Can we explode this into an if-statement? It would be more readable, IMO.

test/integration/spec/connect.js
manjeshbhargav: @markandrus In the source, yes. I can change them in this test as well.

markandrus: If you don't mind, that would be nice 👍

markandrus:
+1

PR #171, opened 2017-09-17
LocalTrackPublication#unpublish should call the underlying ._unpublish...

manjeshbhargav:
...callback with its underlying LocalTrack.

@markandrus ,

Today, while trying to smoke-test the Track publishing APIs with simpler-signaling, I noticed that `LocalTrackPublication#unpublish` was calling it's underlying `._unpublish` callback with itself instead of its underlying `LocalTrack`, which is the expected behavior. So this PR contains the fix for it along with an integration test.

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -209,6 +209,10 @@ LocalParticipant.prototype._getOrCreateLocalTrackPublication = function _getOrCr
       + ' cannot be published'));
   }
 
+  function unpublish(publication) {
+    self.unpublishTrack(publication.track);
+  }
+
   return trackSignaling.getSid().then(function(sid) {
     if (!self.tracks.has(localTrack.id)) {
       throw new Error('The ' + localTrack + ' was unpublished');
@@ -230,7 +234,6 @@ LocalParticipant.prototype._getOrCreateLocalTrackPublication = function _getOrCr
     localTrackPublication = getTrackPublication(self.trackPublications, localTrack);
 
     if (!localTrackPublication) {
-      var unpublish = self.unpublishTrack.bind(self);
       localTrackPublication = util.asLocalTrackPublication(localTrack, sid, unpublish, options);
       self.trackPublications.set(sid, localTrackPublication);
       publishedKindTracks.set(sid, localTrackPublication);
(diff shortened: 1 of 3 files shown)


test/unit/spec/media/track/localtrackpublication.js
```diff
@@ -74,8 +74,8 @@ const sinon = require('sinon');
         ret = localTrackPublication.unpublish();
       });
 
-      it('should call the unpublish callback (with the LocalTrackPublication) that is passed to the constructor', () => {
-        sinon.assert.calledWith(unpublish, localTrackPublication);
+      it('should call the unpublish callback (with the LocalTrack) that is passed to the constructor', () => {
```
markandrus: Hmm, should we keep this as-is and just let the component that handles the `unpublish` command reach into the `track` property?

markandrus:
@manjeshbhargav thanks for identifying this and adding the tests. When you say

> `LocalTrackPublication#unpublish` was calling it's underlying `._unpublish` callback with itself instead of its underlying `LocalTrack`, which is the expected behavior.

what do you mean "expected" behavior? It seems to me this could work either way (either with a reference to the underlying LocalTrack or the LocalTrackPublication itself). If we look at the `@param` we included in [lib/meda/track/localtrackpublication.js](https://github.com/twilio/twilio-video.js/blob/master/lib/media/track/localtrackpublication.js#L15), it shows

```
 * @param {function(LocalTrackPublication): void} unpublish - The callback
 *   that unpublishes the {@link LocalTrackPublication}
```

But then I notice, in [lib/localparticipant.js](https://github.com/twilio/twilio-video.js/blob/master/lib/localparticipant.js#L234), we pass `unpublishTrack` as the callback, which is supposed to accept a LocalTrack (or MediaStreamTrack). I think we should use a function that accepts a LocalTrackPublication instead.

manjeshbhargav:
@markandrus I think what you're suggesting is fine too. Before the fix, `validateLocalTrack` was throwing inside `LocalParticipant#unpublishTrack` because we were passing a `LocalTrackPublication` to it.

manjeshbhargav:
+1

PR #172, opened 2017-09-18
{add,remove}Track{,s} should delegate to {,un}publishTrack{,s}

markandrus:
Otherwise LocalTrackPublications won't be created/updated.

***

@manjeshbhargav per our Slack conversation.

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -259,13 +259,19 @@ LocalParticipant.prototype._getOrCreateLocalTrackPublication = function _getOrCr
 LocalParticipant.prototype.addTrack = function addTrack(track) {
   this._log.deprecated('LocalParticipant#addTrack has been deprecated. '
     + 'Use LocalParticipant#publishTrack instead.');
-  return this._addTrack(util.asLocalTrack(track, {
-    log: this._log,
+  util.validateLocalTrack(track, {
     LocalAudioTrack: this._LocalAudioTrack,
     LocalDataTrack: this._LocalDataTrack,
     LocalVideoTrack: this._LocalVideoTrack,
     MediaStreamTrack: this._MediaStreamTrack
-  }));
+  });
+  if (this.tracks.has(track.id)) {
+    return null;
+  }
+  this.publishTrack(track).catch(function() {
+    // Do nothing.
+  });
+  return this.tracks.get(track.id);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/localparticipant.js
```diff
@@ -259,13 +259,19 @@ LocalParticipant.prototype._getOrCreateLocalTrackPublication = function _getOrCr
...
     MediaStreamTrack: this._MediaStreamTrack
-  }));
+  });
+  if (this.tracks.has(track.id)) {
+    return null;
+  }
+  this.publishTrack(track).catch(function() {
+    // Do nothing.
+  });
+  return track;
```
manjeshbhargav: @markandrus I think this should be: `return this.tracks.get(track.id);` because, `track` can also be a `MediaStreamTrack`.

lib/localparticipant.js
```diff
@@ -288,12 +294,13 @@ LocalParticipant.prototype.addTracks = function addTracks(tracks) {
...
-    var addedTrack = self.addTrack(track);
-    return addedTrack ? addedTracks.concat(addedTrack) : addedTracks;
-  }, []);
+  tracks = tracks.filter(function(track) {
+    return !this.tracks.has(track.id);
+  }, this);
+  this.publishTracks(tracks).catch(function() {
+    // Do nothing.
+  });
+  return tracks;
```
manjeshbhargav: @markandrus This should be `return tracks.map(track => this.tracks.get(track.id));`

manjeshbhargav:
@markandrus Why are we removing the unit test cases?

lib/localparticipant.js
markandrus: Ah, very good point 👍

lib/localparticipant.js
markandrus: Good catch!

markandrus:
Unfortunately the tests aren't really valid. They tested implementation, e.g., "does `addTrack` call `_addTrack` with certain arguments?" and "does `addTracks` delegate to `addTrack`". Rather than completely remove them, they ought to be replaced with tests that test the actual behavior rather than implementation. But we're a little pressed for time right now. OK to continue without them and add new ones at a later time? We may not even need to add some of them back once we remove the deprecated methods.

manjeshbhargav:
+1

PR #178, opened 2017-09-28
Small update to explain how to use preferred codec with Group Rooms

lifeofzero:


markandrus:
Closing in favor of #178. I think we should consider our CHANGELOG.md entry for a particular version "fixed" once released. I also think we may want a different place to surface this information than JSDoc... e.g., some guides like how React does it. But maybe JSDoc is OK for now.

Issue #184, opened 2017-10-05
Unhandled rejection on Safari (iPad retina iOS 8.0.2 and etc)


av-k:
On mobile devices (Safari), the request to receive user media data don't return anything.

![Image of bug](https://raw.githubusercontent.com/av-k/pub/master/images/twilio-video/ipad8-usermedia-bug-o.jpg)

markandrus:
Thanks for the report @av-k—I think we should remove that `return` statement, so that, on platforms that don't support `getUserMedia`, like iOS 8, we invoke `onFailure` with the expected error. WDYT @manjeshbhargav?

markandrus:
@av-k we release twilio-webrtc.js@1.0.2 with a fix for this. `getUserMedia` should now reject instead of hang indefinitely. If you install twilio-video.js from NPM, could you please try re-installing? Otherwise, if you use the CDN, our next release will have the latest fix.

PR #186, opened 2017-10-05
Update CHANGELOG.md for 1.5.0

markandrus:
@manjeshbhargav mind taking a look? I also combed through some of the `@example`s and

* Where the Room `name` deviated, standardized on "my-cool-room".
* Used "microphone" and "camera" for the LocalAudioTrack and LocalVideoTrack names.

My feeling is we should be prescriptive in the names, here. It also lets us standardize on future language we might use (for example, the `name` "screenshare") when we update the screensharing guide. ~~We might also want to fix https://github.com/twilio/twilio-video.js/issues/184 for this release since it looks pretty small.~~ EDIT: Actually, looks like that is now in twilio-webrtc.js.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,81 @@
+1.5.0 (in progress)
+===================
+
+New Features
+------------
+
+- You can now specify Track names. Refer to the Track name guide below for more
+  information.
+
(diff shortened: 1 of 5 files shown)


manjeshbhargav:
+1

PR #191, opened 2017-10-12
Use RTCRtpReceivers where available

markandrus:
@manjeshbhargav please take a look. Fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1405893. The `filter` should not really be necessary in spec-compliant implementations, but there _was_ [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=751785), so perhaps worth keeping the check?


manjeshbhargav:
+1

PR #193, opened 2017-10-12
Add LocalDataTrack example to JSDoc

markandrus:
@manjeshbhargav not quite sure about this example--it's rather large, but seems like we really want to explain both halves (local and remote).

screen shot 2017-10-12 at 3 00 43 pm

### DIFF
--- a/lib/media/track/localdatatrack.js
+++ b/lib/media/track/localdatatrack.js
@@ -17,6 +17,34 @@ var didPrintDataTrackWarning = false;
  * @extends {Track}
  * @param {LocalTrackOptions} [options] - {@link LocalTrack} options
  * @property {Track.Kind} kind - "data"
+ * @example
+ * var Video = require('twilio-video');
+ *
+ * var localDataTrack = new Video.LocalDataTrack();
+ * window.addEventListener('mousemove', function(event) {
+ *   localDataTrack.send({
+ *     x: e.clientX,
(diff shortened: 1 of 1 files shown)


manjeshbhargav:
@markandrus How about this example:

```javascript
var Video = require('twilio-video');
var localDataTrack = new Video.LocalDataTrack();

Video.connect('my-token', {
  name: 'my-room',
  tracks: [localDataTrack]
}).then(function(room) {
  room.on('participantConnected', function(participant) {
    localDataTrack.send('Welcome!');
  });
});

Video.connect('other-token', {
  name: 'my-room'
}).then(function(room) {
  room.on('trackSubscribed', function(track, participant) {
    track.on('message', function(message) {
      console.log(message); // Welcome!
    });
  });
});
```

markandrus:
@manjeshbhargav looks good--i like that it's a bit more symmetric and has a nice kind of sub-narrative... one concern though: the DataTracks aren't completely reliable/timing sensitive. If you wrote a test for sending a single message, I'm not sure it would succeed. Nevertheless, maybe this is sufficient for an example?

manjeshbhargav:
@markandrus You could delay `localDataTrack.send()` by a second or two. I think that would work.

markandrus:
@manjeshbhargav i updated to your example

manjeshbhargav:
+1

Issue #149, opened 2017-08-15
Is there a quick way to check if the browser is supported without actually connecting to a call?


dannycohn:
I want to provide a custom "test" page that my users can load in order to check if their browser is supported.  I do not need to check network conditions, just browser support.  Is there a way to perform this check with the SDK without actually connecting to a call?

manjeshbhargav:
Hi @dannycohn ,

You can use the following snippet:
```javascript
function isRTCPeerConnectionSupported() {
  return typeof RTCPeerConnection !== 'undefined'
    || typeof webkitRTCPeerConnection !== 'undefined'
    || typeof mozRTCPeerConnection !== 'undefined';
}

function isGetUserMediaSupported() {
  return !!(navigator.mediaDevices && navigator.mediaDevices.getUserMedia)
    || !!(navigator.getUserMedia)
    || !!(navigator.webkitGetUserMedia)
    || !!(navigator.mozGetUserMedia);
}

function isWebRTCSupported() {
  return isGetUserMediaSupported() && isRTCPeerConnectionSupported();
}
```

manjeshbhargav:
@dannycohn ,

Do you think that it will be useful for you if our SDK provided a way to check for browser support. Do you have any suggestions on how that might look? Please feel free to respond with your thoughts.

Thanks,

Manjesh Malavalli

dannycohn:
Yes I think it would be extremely valuable. I'm not even sure that the code snippet you provided would provide enough confidence.  That snippet returns true on Safari 11, but with 1.2.0, Safari 11 doesn't work. I would think that a check that is provided by the SDK that can confirm specifically that the loaded SDK supports the current browser would be very helpful.

I want to test for browser support as soon as the page loads, and notify the user immediately if it is not supported, rather than wait until a call starts.

I guess I just imagined something along the lines of `Video.isBrowserSupported()`.  It could either return a boolean, or if you want to indicate experimental support, an enum value indicating the level of support.

manjeshbhargav:
@dannycohn ,

Thank you for the detailed response. Regarding Safari, we just released `twilio-video.js@1.2.1` that has experimental Safari support ([CHANGELOG](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md)).

Regarding your suggestion, we were thinking of something similar, although it would just be a `boolean` property instead of a `function`. We will get back to you on implementing that check soon.

Thanks,

Manjesh Malavalli

[...]

PR #196, opened 2017-10-16
Twilio.Video.isSupported Flag

manjeshbhargav:
@markandrus


markandrus:
The goal of this PR is to provide a solution for https://github.com/twilio/twilio-video.js/issues/149, namely, allowing users to write code like,

```js
const { connect, isSupported } = require('twilio-video')
if (isSupported) {
  connect(token, { /* ... */ })
} else {
  console.error('Sorry, but video chat is not supported on this platform')
}
```

PR #200, opened 2017-10-23
Add isSupported to CHANGELOG.md

markandrus:
@manjeshbhargav how does this look?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ New Features
   the LocalParticipant with an informative error. Similarly, if publication
   succeeds, we will raise a "trackPublished" event with the resulting
   LocalTrackPublication.
+- Added a new top-level boolean property, `isSupported`, that indicates whether
+  or not the browser/environment/platform currently running twilio-video.js
+  contains the necessary APIs (for example, WebRTC) to participate in video
+  chats. Depending on how you use twilio-video.js, you can use one of the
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,19 @@ New Features
   the LocalParticipant with an informative error. Similarly, if publication
   succeeds, we will raise a "trackPublished" event with the resulting
   LocalTrackPublication.
+- Added a new boolean property, `isSupported`, that indicates whether or not the
```
manjeshbhargav: @markandrus `Added a new boolean property to the Video namespace`?

CHANGELOG.md
```diff
@@ -18,6 +18,19 @@ New Features
...
+  Depending on how you use twilio-video.js, you can use one of the methods below
+  to access it:
+
+  ```js
+  // Option 1: Using `require`
+  const { isSupported } = require('twilio-video');
+
+  // Option 2: Using browser global
+  const { isSupported } = Twilio.Video;
+  ```
```
manjeshbhargav: @markandrus Maybe the example is more descriptive if the object deconstruction feature is not used?

markandrus: We are using it elsewhere, e.g., we have a sample below where we do

```
const { LocalDataTrack } = require('twilio-video');
```

so I think it is OK to continue.

PR #201, opened 2017-10-23
JSDK-1580 Cannot send data on republished LocalDataTracks

manjeshbhargav:
@markandrus 

In order to fix this bug, we have to close any RTCDataChannels associated with the LocalDataTrack while unpublishing.


markandrus:
LGTM--let's see what Travis says

test/integration/spec/localparticipant.js
```diff
@@ -354,7 +354,15 @@ const isSafari = guess === 'safari';
             thoseTracks.forEach(thatTrack => assert.equal(thatTrack.name, thisLocalTrackPublication.trackName));
           });
 
-          if (kind !== 'data') {
+          if (kind === 'data') {
+            it('should transmit any data sent through the LocalDataTrack to the Room to each RemoteDataTrack', async () => {
+              const thoseTracks = thoseTracksMap[event];
+              const thoseTracksReceivedData = thoseTracks.map(track => new Promise(resolve => track.once('message', resolve)));
+              thisTrack.send('foo');
```
markandrus: Actually, just remembered—this won't be a reliable test. We need to do something like

```js
interval = setInterval(() => {
  thisTrack.send('foo');
}, 100);
```

and then

1. Check that at least 1 message gets through
2. `clearInterval` in `after`/`afterEach`

markandrus:
I see some timeouts in Travis CI logs for the new test in SFU topology.

lib/data/sender.js
```diff
@@ -46,7 +46,9 @@ DataTrackSender.prototype.addDataChannel = function addDataChannel(dataChannel)
  * @returns {this}
  */
 DataTrackSender.prototype.removeDataChannel = function removeDataChannel(dataChannel) {
-  this._dataChannels.delete(dataChannel);
+  if (this._dataChannels.delete(dataChannel)) {
+    dataChannel.close();
+  }
```
markandrus: I was thinking about this some more—can we move this responsibility to PeerConnectionV2's `removeDataTrackSender`? e.g.,

```
PeerConnectionV2.prototype.removeDataTrackSender = function removeDataTrackSender(dataTrackSender) {
  var dataChannel = this._dataTrackSenders.get(dataTrackSender);
  if (dataChannel) {
    dataTrackSender.removeDataChannel(dataChannel);
    this._dataTrackSenders.delete(dataTrackSender);
  }
};
```

markandrus: Just to elaborate on my thinking here: PeerConnectionV2 currently creates the RTCDataChannel, but if we let DataTrackSender `close` the RTCDataChannel, now we have the responsibility for its lifecycle split over two classes. Would rather let PeerConnectionV2 manage its lifecycle and DataTrackSender just call `send`.

manjeshbhargav: @markandrus I was thinking that once PeerConnectionV2 creates the RTCDataChannel, it transfers the ownership to DataTrackSender. This design pattern is pretty common in C++ for ex: `scoped_ptr`, which is assigned some memory, but when execution goes out of the scope in which it was created, the `scoped_ptr` now automatically de-allocates the memory associated with it.

manjeshbhargav:
@markandrus Just confirmed that this fix works with the `rtc-sfu-service@10.2.0.0` deployed to stage. I think we can merge this to master now.

PR #202, opened 2017-10-23
Add back the dummy audio MediaStreamTrack fix

markandrus:
_Not ready to merge._

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -27,8 +27,22 @@ function RoomV2(localParticipant, initialState, transport, peerConnectionManager
     _peerConnectionManager: {
       value: peerConnectionManager
     },
+    _published: {
+      value: new Map()
+    },
+    _publishedRevision: {
+      value: 0,
+      writable: true
+    },
     _RemoteParticipantV2: {
       value: options.RemoteParticipantV2
+    },
+    _subscribed: {
+      value: new Map()
+    },
+    _subscribedRevision: {
+      value: 0,
+      writable: true
     },
     _transport: {
       value: transport
@@ -163,8 +177,23 @@ RoomV2.prototype._update = function _update(roomState) {
     this.recording.update(roomState.recording);
   }
 
-  if (roomState.published) {
+  if (roomState.published && roomState.published.revision > this._publishedRevision) {
+    this._publishedRevision = roomState.published.revision;
+    roomState.published.tracks.forEach(function(track) {
+      if (track.sid) {
+        this._published.set(track.id, track.sid);
+      }
+    }, this);
     this.localParticipant.update(roomState.published);
+  }
+
+  if (roomState.subscribed && roomState.subscribed.revision > this._subscribedRevision) {
+    this._subscribedRevision = roomState.subscribed.revision;
+    roomState.subscribed.tracks.forEach(function(track) {
+      if (track.sid) {
+        this._subscribed.set(track.id, track.sid);
+      }
+    }, this);
   }
 
   if (roomState.participant) {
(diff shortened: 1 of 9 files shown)


lib/signaling/v2/room.js
```diff
@@ -166,10 +180,25 @@ RoomV2.prototype._update = function _update(roomState) {
     this.recording.update(roomState.recording);
   }
 
-  if (roomState.published) {
+  if (roomState.published && roomState.published.revision > this._publishedRevision) {
+    this._publishedRevision = roomState.published.revision;
+    roomState.published.tracks.forEach(function(track) {
```
manjeshbhargav: @markandrus Shouldn't we also remove any tracks that have been unpublished?

lib/signaling/v2/room.js
```diff
@@ -166,10 +180,25 @@ RoomV2.prototype._update = function _update(roomState) {
...
+      if (track.sid) {
+        this._published.set(track.id, track.sid);
+      }
+    }, this);
     this.localParticipant.update(roomState.published);
   }
 
+  if (roomState.subscribed && roomState.subscribed.revision > this._subscribedRevision) {
+    this._subscribedRevision = roomState.subscribed.revision;
+    roomState.subscribed.tracks.forEach(function(track) {
```
manjeshbhargav: @markandrus Shouldn't we also remove any tracks that have been unsubscribed?

lib/signaling/v2/room.js
markandrus: No, because `getStats` provides historical data, even after MediaStreamTracks have been removed. We should keep this set up to date in order to 1. filter, e.g., for this use case, but also 2. to be able to insert Track SIDs (in a future PR).

lib/signaling/v2/room.js
markandrus: Same rationale as above

lib/signaling/v2/room.js
```diff
@@ -184,7 +213,26 @@ RoomV2.prototype._update = function _update(roomState) {
  * @returns {Promise.<Array<StatsReport>>}
  */
 RoomV2.prototype.getStats = function getStats() {
-  return this._peerConnectionManager.getStats();
+  var self = this;
+
+  function filterLocalTrackStats(trackStats) {
+    return self._published.has(trackStats.trackId);
```
manjeshbhargav: @markandrus Can't we get this from `[...this.localParticipant.tracks.values()].filter(track => !!track.sid)`?

manjeshbhargav: @markandrus Can't we do this instead?
```javascript
const publishedTracks = flatMap(this.localParticipant.tracks).filter(track => !!track.sid);
return publishedTracks.has(trackStats.trackId);
```

lib/signaling/v2/room.js
```diff
@@ -184,7 +213,26 @@ RoomV2.prototype._update = function _update(roomState) {
...
+  }
+
+  function filterRemoteTrackStats(trackStats) {
+    return self._subscribed.has(trackStats.trackId);
```
manjeshbhargav: @markandrus Can't we do this instead?
```javascript
const subscribedTracks = flatMap(this.participants, participant => flatMap(participant.tracks));
return subscribedTracks.has(trackStats.trackId);
```

lib/signaling/v2/room.js
manjeshbhargav: @markandrus Are you sure, because we are filtering `getStats()` results based on the current set of `MediaStreamTracks` in the `PeerConnection`.

markandrus: @manjeshbhargav yes, check out [4.3 Lifetime of stats objects](https://www.w3.org/TR/webrtc-stats/#lifetime-of-stats-objects):

> The general rule is that stats objects, once created, exist for the lifetime of the PeerConnection that contains them, even when the underlying object they report on is stopped or deleted. This is important in order to report consistently on short-lived objects and to be able to report totals over the lifetime of a PeerConnection.
>
> When an object is closed or deleted, the timestamp member of the stats object stops updating; it is frozen at the time when the object stopped.

As for filtering results based on the current set of MediaStreamTracks in the PeerConnection, we probably need to revisit this decision.

lib/signaling/v2/room.js
markandrus: The `localParticipant.tracks` and RemoteParticipant `tracks` may not include the full set of Tracks, since we will remove from those collections as Tracks are removed.

PR #204, opened 2017-11-02
Test improvements

markandrus:
There are some changes to `lib/` in #202 that aren't quite ready. This PR includes only the test improvements. EDIT: Notes on some of these changes:

1. Updates ESLint
2. Lints `test/`
3. Removes `_sidDeferred`/`getSid` from `lib/signaling/track.js` as they were unused
4. Gates some of our `dummyEl` and related code in `lib/media/tracks` on presence of `document`
5. Adds logging support to PeerConnectionV2—we can see, for example, if applying a local or remote description fails (esp. useful in Travis CI logs)
6. Removes an unused property on RoomV2
7. Adds `getStats` integration tests

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -78,7 +78,11 @@ rules:
   semi: [2, always]
   keyword-spacing: 2
   space-before-blocks: [2, always]
-  space-before-function-paren: [2, never]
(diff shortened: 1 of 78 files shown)


manjeshbhargav:
+1

PR #206, opened 2017-11-06
JSDK-1579 Removing options items which are explicitly "undefined".

manjeshbhargav:
@markandrus ,

We now remove any keys in `ConnectOptions` that are explicitly set to `undefined`. That way,
in this specific bug, `wsServer` is properly constructed.


markandrus:
LGTM

PR #207, opened 2017-11-08
JSDK-1626 Should always pass SDP dict to ._setLocalDescription()

manjeshbhargav:
@markandrus 

In some places we were passing `RTCSessionDescription` to `._setLocalDescription`, where we were trying to apply the sdp transform for simulcast. So this PR now passes only the `RTCSessionDescriptionInitDict`.

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -314,8 +314,7 @@ PeerConnectionV2.prototype._handleDataChannelEvent = function _handleDataChannel
 PeerConnectionV2.prototype._handleGlare = function _handleGlare(offer) {
   var self = this;
   return Promise.resolve().then(function() {
-    var rollback = new self._RTCSessionDescription({ type: 'rollback' });
-    return self._setLocalDescription(rollback);
+    return self._setLocalDescription({ type: 'rollback' });
   }).then(function() {
     return self._answer(offer);
   }).then(function() {
@@ -423,22 +422,22 @@ PeerConnectionV2.prototype._offer = function _offer() {
       offer.sdp,
       self._preferredAudioCodecs,
       self._preferredVideoCodecs);
-    var updatedOffer = new RTCSessionDescription({
-      type: 'offer',
-      sdp: updatedSdp
-    });
+
     self._shouldOffer = false;
     if (!self._negotiationRole) {
       self._negotiationRole = 'offerer';
       self._needsInitialAnswer = true;
     }
-    return self._setLocalDescription(updatedOffer);
+    return self._setLocalDescription({
+      type: 'offer',
+      sdp: updatedSdp
+    });
   });
 };
 
 /**
  * Set a local description on the {@link PeerConnectionV2}.
- * @param {RTCSessionDescriptionInit} description
+ * @param {RTCSessionDescription|RTCSessionDescriptionInit} description
  * @returns {Promise<void>}
  */
 PeerConnectionV2.prototype._setLocalDescription = function _setLocalDescription(description) {
@@ -450,9 +449,15 @@ PeerConnectionV2.prototype._setLocalDescription = function _setLocalDescription(
 
   return Promise.resolve().then(function() {
     if (description.sdp) {
-      description.sdp = isChrome && vp8SimulcastRequested
-        ? self._setSimulcast(description.sdp, self._trackIdsToAttributes)
-        : description.sdp;
+      // NOTE(mmalavalli): We do not directly modify "description.sdp" here as
+      // "description" might be an RTCSessionDescription, in which case its
+      // properties are immutable.
+      description = {
+        type: description.type,
+        sdp: isChrome && vp8SimulcastRequested
+          ? self._setSimulcast(description.sdp, self._trackIdsToAttributes)
+          : description.sdp
+      };
     }
     description = new self._RTCSessionDescription(description);
     if (description.type === 'answer') {
--- a/test/integration/spec/localparticipant.js
+++ b/test/integration/spec/localparticipant.js
@@ -66,7 +66,7 @@ describe('LocalParticipant', function() {
         }
       ],
       [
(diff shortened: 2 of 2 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -263,7 +263,10 @@ PeerConnectionV2.prototype._answer = function _answer(offer) {
   }).then(function() {
     return self._peerConnection.createAnswer();
   }).then(function(answer) {
-    return self._setLocalDescription(answer);
+    return self._setLocalDescription({
+      type: answer.type,
+      sdp: answer.sdp
+    });
```
markandrus: This doesn't need to change, does it?

manjeshbhargav: @markandrus Yes it does, because `.createAnswer()` returns an `RTCSessionDescription`, whose members are read-only in Safari.

lib/signaling/v2/peerconnection.js
```diff
@@ -314,8 +317,7 @@ PeerConnectionV2.prototype._handleDataChannelEvent = function _handleDataChannel
 PeerConnectionV2.prototype._handleGlare = function _handleGlare(offer) {
   var self = this;
   return Promise.resolve().then(function() {
-    var rollback = new self._RTCSessionDescription({ type: 'rollback' });
-    return self._setLocalDescription(rollback);
+    return self._setLocalDescription({ type: 'rollback' });
```
markandrus: This is probably fine if we rewrap in an RTCSessionDescription inside `_setLocalDescription`. The WebRTC 1.0 APIs are supposed to accept a plain JSON object in the future (e.g., RTCSessionDescriptionInit).

manjeshbhargav: @markandrus We do rewrap it. If you look at the JSDoc for `._setLocalDescription`, then it is expecting to accept a Dict.

lib/signaling/v2/peerconnection.js
```diff
@@ -419,20 +421,18 @@ PeerConnectionV2.prototype._offer = function _offer() {
...
       self._needsInitialAnswer = true;
     }
-    return self._setLocalDescription(updatedOffer);
+    return self._setLocalDescription({
+      type: 'offer',
+      sdp: self._setCodecPreferences(
+        offer.sdp,
+        self._preferredAudioCodecs,
+        self._preferredVideoCodecs)
+    });
```
markandrus: What's the difference here? Is the issue that we called

```js
new RTCSessionDescription
```

instead of

```js
new self._RTCSessionDescription
```

?

manjeshbhargav: @markandrus `._setLocalDescription` accepts a Dict. So we should pass the dict directly instead of wrapping it in `self._RTCSessionDescription`.

lib/signaling/v2/peerconnection.js
markandrus: Is `_setLocalDescription` mutating its input argument? If so, that might explain why I saw the read-only warnings in Firefox again recently after we had fixed them. Instead of letting `_setLocalDescription` mutate its argument, I think it should copy it, if necessary.

lib/signaling/v2/peerconnection.js
markandrus: It is weird that it would accept _only_ the dict form though, and I guess that's because it mutates it?

lib/signaling/v2/peerconnection.js
markandrus: I think anywhere the RTCSessionDescriptionInit is acceptable, the RTCSessionDescription should be acceptable, too.

lib/signaling/v2/peerconnection.js
```diff
@@ -314,8 +314,7 @@ PeerConnectionV2.prototype._handleDataChannelEvent = function _handleDataChannel
 PeerConnectionV2.prototype._handleGlare = function _handleGlare(offer) {
   var self = this;
   return Promise.resolve().then(function() {
-    var rollback = new self._RTCSessionDescription({ type: 'rollback' });
-    return self._setLocalDescription(rollback);
+    return self._setLocalDescription({ type: 'rollback' });
```
markandrus: 👍

lib/signaling/v2/peerconnection.js
```diff
@@ -423,22 +422,22 @@ PeerConnectionV2.prototype._offer = function _offer() {
...
     self._shouldOffer = false;
     if (!self._negotiationRole) {
       self._negotiationRole = 'offerer';
       self._needsInitialAnswer = true;
     }
-    return self._setLocalDescription(updatedOffer);
+    return self._setLocalDescription({
+      type: 'offer',
+      sdp: updatedSdp
+    });
```
markandrus: 👍

lib/signaling/v2/peerconnection.js
```diff
@@ -450,9 +449,15 @@ PeerConnectionV2.prototype._setLocalDescription = function _setLocalDescription(
...
-        : description.sdp;
+      // NOTE(mmalavalli): We do not directly modify "description.sdp" here as
+      // "description" might be an RTCSessionDescription, in which case its
+      // properties are immutable.
+      description = {
+        type: description.type,
+        sdp: isChrome && vp8SimulcastRequested
+          ? self._setSimulcast(description.sdp, self._trackIdsToAttributes)
+          : description.sdp
+      };
```
markandrus: LGTM—should we do the same in `_setRemoteDescription`? I see it modifies `description.sdp`

markandrus:
One comment, otherwise +1

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus In `_setRemoteDescription` we only get the dict because usually we are passing the contents of the RSP message. I don't think there is any case where we would be passing an RTCSessionDescription to it.

PR #208, opened 2017-11-08
JSDK-1627 Set DataTrackReceiver's RTCDataChannel .binaryType to "arraybuffer"

manjeshbhargav:
@markandrus In Firefox, The `DataTrackReceiver`'s `RTCDataChannel` has it's `.binaryType` set to "blob", where as in Chrome/Safari it is set to "arraybuffer". So we explicitly set it.

### DIFF
--- a/test/integration/spec/localparticipant.js
+++ b/test/integration/spec/localparticipant.js
@@ -204,7 +204,6 @@ describe('LocalParticipant', function() {
         x => `that has ${x} been published`
       ]
     ], ([isEnabled, kind, withName, when]) => {
-      let dataChannelSendInterval;
       let thisRoom;
       let thisParticipant;
       let thisLocalTrackPublication;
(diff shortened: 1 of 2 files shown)


markandrus:
@manjeshbhargav LGTM

markandrus:
I think we could add a test for this, but otherwise fine to merge.

test/integration/spec/localparticipant.js
```diff
@@ -335,13 +331,16 @@ describe('LocalParticipant', function() {
...
-              const data = await Promise.all(thoseTracksReceivedData);
-              data.forEach(item => assert.equal(item, 'foo'));
+            ['string', 'arraybuffer'].forEach(dataType => {
+              it(`should transmit any ${dataType} data sent through the LocalDataTrack to the Room to each RemoteDataTrack`, async () => {
+                const data = dataType === 'string' ? 'foo' : new Uint32Array([1, 2, 3]);
+                const thoseTracks = thoseTracksMap[event];
+                const thoseTracksReceivedData = thoseTracks.map(track => new Promise(resolve => track.once('message', resolve)));
+                dataChannelSendInterval = setInterval(() => thisTrack.send(dataType === 'string' ? data : data.buffer), 1000);
+                const receivedData = await Promise.all(thoseTracksReceivedData);
+                clearInterval(dataChannelSendInterval);
```
markandrus: If we moved the `clearInterval` from the `after` block, I don't think we need to define `dataChannelSendInterval` outside this anymore, do we?

PR #214, opened 2017-11-14
JSDK-1640, JSDK-1641, JSDK-1642 Consuming WebRTC 1.0 APIs.

manjeshbhargav:
@markandrus 

Let me know what you think. I've basically opened this PR to run travis for FF/Safari. Chrome integration tests are working when I ran it locally (for both P2P and SFU).

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -388,7 +388,9 @@ PeerConnectionV2.prototype._maybeReoffer = function _maybeReoffer(offer) {
   // re-introducing 'offerToReceiveAudio' to the default RTCOfferOptions with a
   // value > 1.
   if (isFirefox) {
-    var senders = this._peerConnection.getSenders();
+    var senders = this._peerConnection.getSenders().filter(function(sender) {
+      return sender.track;
+    });
     shouldReoffer = ['audio', 'video'].reduce(function(shouldOffer, kind) {
       var mediaSections = getMediaSections(offer.sdp, kind, '(sendrecv|recvonly)');
       var sendersOfKind = senders.filter(isSenderOfKind.bind(null, kind));
@@ -749,7 +751,9 @@ PeerConnectionV2.prototype.removeDataTrackSender = function removeDataTrackSende
  */
 PeerConnectionV2.prototype.removeMediaStream = function removeMediaStream(mediaStream) {
   if (this._peerConnection.signalingState !== 'closed') {
-    var senders = new Map(this._peerConnection.getSenders().map(function(sender) {
+    var senders = new Map(this._peerConnection.getSenders().filter(function(sender) {
+      return sender.track;
+    }).map(function(sender) {
       return [sender.track.id, sender];
     }));
     mediaStream.getTracks().forEach(function(track) {
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -4,7 +4,6 @@ var WebRTC = require('@twilio/webrtc');
 var DefaultRTCIceCandidate = WebRTC.RTCIceCandidate;
 var DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
 var DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
-var flatMap = require('../../util').flatMap;
 var getStatistics = WebRTC.getStats;
 var getMediaSections = require('../../util/sdp').getMediaSections;
 var guessBrowser = require('../../util').guessBrowser;
@@ -639,7 +638,9 @@ PeerConnectionV2.prototype.addDataTrackSender = function addDataTrackSender(data
  */
 PeerConnectionV2.prototype.addMediaStream = function addMediaStream(mediaStream) {
   if (this._peerConnection.signalingState !== 'closed') {
-    this._peerConnection.addStream(mediaStream);
+    mediaStream.getTracks().forEach(function(track) {
+      this._peerConnection.addTrack(track, mediaStream);
+    }, this);
   }
 };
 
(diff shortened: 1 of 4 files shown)


lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -323,16 +323,17 @@ function applyMediaStreamTrackAndDataTrackSenderChanges(peerConnectionManager, c
     || changes.data.remove.size
     || changes.media.add.size
     || changes.media.remove.size) {
+    peerConnectionManager._peerConnections.forEach(function(peerConnection) {
```
manjeshbhargav: @markandrus I've separated out the removal of tracks from their addition that happens later because we need to remove the tracks in the local stream collection before changes are applied to it.

lib/signaling/v2/peerconnection.js
```diff
@@ -389,7 +388,9 @@ PeerConnectionV2.prototype._maybeReoffer = function _maybeReoffer(offer) {
   // re-introducing 'offerToReceiveAudio' to the default RTCOfferOptions with a
   // value > 1.
   if (isFirefox) {
-    var senders = this._peerConnection.getSenders();
+    var senders = this._peerConnection.getSenders().filter(function(sender) {
+      return sender.track;
+    });
```
markandrus: Why do we need to filter here?

lib/signaling/v2/peerconnection.js
```diff
@@ -677,15 +680,10 @@ PeerConnectionV2.prototype.getRemoteMediaStreamTracksAndDataTrackReceivers = fun
...
-      return track;
-    });
-  }
-  return flatMap(this._peerConnection.getRemoteStreams(), function(mediaStream) {
-    return mediaStream.getTracks();
+  return this._peerConnection.getReceivers().map(function(receiver) {
+    return receiver.track;
+  }).filter(function(track) {
+    return track;
   });
```
markandrus: Probably worth including a comment that explains that, the reason we filter here, is because there was a short-lived bug in Chrome where this could be null. I should've included such a comment the first time around.

lib/signaling/v2/peerconnection.js
```diff
@@ -753,7 +751,16 @@ PeerConnectionV2.prototype.removeDataTrackSender = function removeDataTrackSende
  */
 PeerConnectionV2.prototype.removeMediaStream = function removeMediaStream(mediaStream) {
   if (this._peerConnection.signalingState !== 'closed') {
-    this._peerConnection.removeStream(mediaStream);
+    var senders = new Map(this._peerConnection.getSenders().filter(function(sender) {
+      return sender.track;
+    }).map(function(sender) {
+      return [sender.track.id, sender];
```
markandrus: I think we should use the MediaStreamTrack itself as the key, not the ID.

markandrus: Also, technically, because of APIs like `replaceTrack`, I think the most appropriate data structure might be `Map<MediaStreamTrack, Set<RTCRtpSender>>`. WDYT?

lib/signaling/v2/peerconnectionmanager.js
markandrus: > we need to remove the tracks in the local stream collection before changes are applied to it.

Why is that? Also, this suggests to me we ought to move to a MediaStreamTrack- rather than MediaStream-based API soon...

manjeshbhargav: @markandrus I agree, however, we do need a MediaStream to pass to `addTrack()`.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus Because the spec for getSenders() says that there might be senders whose tracks are null.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus Please refer to this comment: https://github.com/twilio/twilio-webrtc.js/pull/19#discussion_r151266957

lib/signaling/v2/peerconnection.js
markandrus: Right, but is this a problem for the way we use this collection? I think we want to ensure that there is at least an RTCRtpSender available—not necessarily that it has a `track` present. Perhaps it's also sufficient to check for RTCRtpReceivers, since there should be the same number.

lib/signaling/v2/peerconnectionmanager.js
markandrus: Yeah, probably we can move the MediaStream yet another level down in the future (and let each PeerConnectionV2 have it's own MediaStream instance; or perhaps pass the same MediaStream to each PeerConnectionV2's constructor). `addTrack` is supposed to let you omit the MediaStream(s) argument completely, but I saw that was unsupported in at least one of the browsers so far.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus But aren't we basing a decision to reoffer based on the number of senders. So shouldn't we count only those senders which are active, ie whose track is not null?

lib/signaling/v2/peerconnectionmanager.js
manjeshbhargav: @markandrus I wouldn't create a new MediaStream for PeerConnectionV2 because I think we just fixed a bug where too many MediaStreams was causing audio problems in FF.

lib/signaling/v2/peerconnection.js
markandrus: Ah, OK—yeah, thinking about it some more, the logic here is intended to ensure we can match up MediaStreamTracks to m=lines, so we should be checking RTCRtpSenders with non-null `track` properties. 👍

markandrus:
This LGTM. IMO, we should flatten/squash commits on merging.

PR #215, opened 2017-11-16
Update ws package to 3.x

vincentwoo:
3.3.1 also contains a fix for a DOS attack for ws servers that most packages will be updating to. Keeping twilio-video.js on 2.x forces consumers to duplicate packages


markandrus:
Thanks, @vincentwoo 👍

PR #219, opened 2017-11-22
MSID Independence

manjeshbhargav:
@markandrus ,

This PR as of now contains the implementation of the `TrackTransceiver` based classes, and their use in the SDK. I've broken down the commits to conform to the task breakdown provided [here](https://paper.dropbox.com/doc/MSID-Independence-Implementation-Plan-XydIbCUd7yiW7YL2hD364).

TODO
------
* Move ordered TrackMatcher to twilio-video.
* Implement MID and Identity TrackMatchers.

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -143,8 +143,7 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
   var signaling = this._signaling;
 
   function localTrackAdded(localTrack) {
-    var trackSender = localTrack._dataTrackSender || localTrack._mediaTrackSender;
-    signaling.addTrack(trackSender, localTrack.name);
+    signaling.addTrack(localTrack._trackSender, localTrack.name);
     log.info('Added a new ' + util.trackClass(localTrack, true) + ':', localTrack.id);
     log.debug(util.trackClass(localTrack, true) + ':', localTrack);
   }
@@ -162,8 +161,7 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
   }
 
   function localTrackRemoved(localTrack) {
-    var trackSender = localTrack._dataTrackSender || localTrack._mediaTrackSender;
-    signaling.removeTrack(trackSender);
+    signaling.removeTrack(localTrack._trackSender);
     log.info('Removed a ' + util.trackClass(localTrack, true) + ':', localTrack.id);
     log.debug(util.trackClass(localTrack, true) + ':', localTrack);
   }
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -143,8 +143,8 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
   var signaling = this._signaling;
 
   function localTrackAdded(localTrack) {
-    var mediaStreamTrackOrDataTrackSender = localTrack._dataTrackSender || localTrack.mediaStreamTrack;
-    signaling.addTrack(mediaStreamTrackOrDataTrackSender, localTrack.name);
+    var trackSender = localTrack._dataTrackSender || localTrack._mediaTrackSender;
+    signaling.addTrack(trackSender, localTrack.name);
     log.info('Added a new ' + util.trackClass(localTrack, true) + ':', localTrack.id);
     log.debug(util.trackClass(localTrack, true) + ':', localTrack);
   }
@@ -162,7 +162,8 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
   }
 
   function localTrackRemoved(localTrack) {
-    signaling.removeTrack(localTrack._dataTrackSender || localTrack.mediaStreamTrack);
+    var trackSender = localTrack._dataTrackSender || localTrack._mediaTrackSender;
+    signaling.removeTrack(trackSender);
     log.info('Removed a ' + util.trackClass(localTrack, true) + ':', localTrack.id);
     log.debug(util.trackClass(localTrack, true) + ':', localTrack);
   }
--- a/lib/signaling/v2/cancelableroomsignalingpromise.js
+++ b/lib/signaling/v2/cancelableroomsignalingpromise.js
@@ -22,12 +22,12 @@ function createCancelableRoomSignalingPromise(token, ua, localParticipant, iceSe
 
   var peerConnectionManager = new PeerConnectionManager(iceServerSource, encodingParameters, preferredCodecs, options);
 
-  var mediaStreamTracksAndDataTrackSenders = flatMap(localParticipant.tracks, function(track) {
-    return [track.mediaStreamTrackOrDataTrackTransceiver];
+  var trackSenders = flatMap(localParticipant.tracks, function(trackV2) {
+    return trackV2.trackTransceiver;
   });
 
   peerConnectionManager.setConfiguration(options);
-  peerConnectionManager.setMediaStreamTracksAndDataTrackSenders(mediaStreamTracksAndDataTrackSenders);
+  peerConnectionManager.setTrackSenders(trackSenders);
 
   var cancelationError = new Error('Canceled');
 
--- a/lib/signaling/v2/cancelableroomsignalingpromise.js
+++ b/lib/signaling/v2/cancelableroomsignalingpromise.js
@@ -23,7 +23,7 @@ function createCancelableRoomSignalingPromise(token, ua, localParticipant, iceSe
   var peerConnectionManager = new PeerConnectionManager(iceServerSource, encodingParameters, preferredCodecs, options);
 
   var trackSenders = flatMap(localParticipant.tracks, function(trackV2) {
-    return trackV2.trackTransceiver;
+    return [trackV2.trackTransceiver];
   });
 
   peerConnectionManager.setConfiguration(options);
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -20,6 +20,9 @@ var DataTrackReceiver = require('../../data/receiver');
 var MediaTrackReceiver = require('../../media/track/receiver');
 var StateMachine = require('../../../lib/statemachine');
 var StatsReport = require('../../stats/statsreport');
+var IdentityTrackMatcher = require('../../util/sdp/trackmatcher/identity');
+var OrderedTrackMatcher = require('../../util/sdp/trackmatcher/ordered');
+var MIDTrackMatcher = require('../../util/sdp/trackmatcher/mid');
 
 var defaults = {
   iceServers: [],
@@ -35,6 +38,7 @@ var defaults = {
 
 var isChrome = guessBrowser() === 'chrome';
 var isFirefox = guessBrowser() === 'firefox';
+var isSafari = guessBrowser() === 'safari';
 
 /*
 PeerConnectionV2 States
@@ -135,7 +139,7 @@ function PeerConnectionV2(id, encodingParameters, preferredCodecs, options) {
         ? options.log.createLog('signaling', this)
         : null
     },
-    _mediaSenders: {
+    _rtpSenders: {
       value: new Map()
     },
     _mediaTrackReceivers: {
@@ -194,6 +198,10 @@ function PeerConnectionV2(id, encodingParameters, preferredCodecs, options) {
     },
     _trackIdsToAttributes: {
       value: new Map()
+    },
+    _trackMatcher: {
+      writable: true,
+      value: null
     },
     id: {
       enumerable: true,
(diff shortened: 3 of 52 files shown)


lib/localparticipant.js
```diff
@@ -143,8 +143,8 @@ LocalParticipant.prototype._handleTrackSignalingEvents = function _handleTrackSi
   var signaling = this._signaling;
 
   function localTrackAdded(localTrack) {
-    var mediaStreamTrackOrDataTrackSender = localTrack._dataTrackSender || localTrack.mediaStreamTrack;
-    signaling.addTrack(mediaStreamTrackOrDataTrackSender, localTrack.name);
+    var trackSender = localTrack._dataTrackSender || localTrack._mediaTrackSender;
```
markandrus: Just wondering: would we benefit from letting this property be `_trackSender` in both cases instead of `_dataTrackSender` and `_mediaTrackSender`?

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -22,12 +22,12 @@ function createCancelableRoomSignalingPromise(token, ua, localParticipant, iceSe
 
   var peerConnectionManager = new PeerConnectionManager(iceServerSource, encodingParameters, preferredCodecs, options);
 
-  var mediaStreamTracksAndDataTrackSenders = flatMap(localParticipant.tracks, function(track) {
-    return [track.mediaStreamTrackOrDataTrackTransceiver];
+  var trackSenders = flatMap(localParticipant.tracks, function(trackV2) {
+    return trackV2.trackTransceiver;
```
markandrus: If we `flatMap`, we should return an Array here, even if it's trivial, right? (I believe the implementation is using `concat`, which will accept singular values, but imagine if we were typing this with TypeScript—I believe this would fail to type.)

lib/signaling/v2/peerconnection.js
```diff
@@ -91,6 +92,12 @@ function PeerConnectionV2(id, encodingParameters, preferredCodecs, options) {
   var peerConnection = new RTCPeerConnection(configuration);
 
   Object.defineProperties(this, {
+    _dataChannels: {
```
markandrus: Can we keep this as `_dataTrackSenders`?

lib/signaling/v2/peerconnection.js
```diff
@@ -126,6 +130,12 @@ function PeerConnectionV2(id, encodingParameters, preferredCodecs, options) {
         ? options.log.createLog('signaling', this)
         : null
     },
+    _mediaSenders: {
```
markandrus: Can we let this be `_mediaTrackSenders`?

lib/signaling/v2/peerconnection.js
```diff
@@ -363,7 +370,15 @@ PeerConnectionV2.prototype._handleSignalingStateChange = function _handleSignali
...
 PeerConnectionV2.prototype._handleTrackEvent = function _handleTrackEvent(event) {
   var mediaStreamTrack = event.track;
-  this.emit('trackAdded', mediaStreamTrack);
+  var mediaTrackReceiver = new MediaTrackReceiver(mediaStreamTrack.id, mediaStreamTrack);
+  this._mediaTrackReceivers.add(mediaTrackReceiver);
+
+  var self = this;
+  mediaStreamTrack.addEventListener('ended', function() {
+    self._mediaTrackReceivers.delete(mediaTrackReceiver);
+  });
```
markandrus: This will not fire if `stop` is called. Perhaps this is OK with our usage here?

lib/signaling/v2/peerconnection.js
```diff
@@ -627,23 +642,25 @@ PeerConnectionV2.prototype.addDataTrackSender = function addDataTrackSender(data
...
 };
 
 /**
- * Add a local MediaStream to the {@link PeerConnectionV2}.
+ * Add the {@link MediaTrackSender} to the {@link PeerConnectionV2}.
+ * @param {MediaTrackSender} mediaTrackSender
  * @param {MediaStream} mediaStream
  * @returns {void}
  */
-PeerConnectionV2.prototype.addMediaStream = function addMediaStream(mediaStream) {
```
markandrus: w00t!

lib/signaling/v2/peerconnection.js
```diff
@@ -627,23 +642,25 @@ PeerConnectionV2.prototype.addDataTrackSender = function addDataTrackSender(data
...
-  if (this._peerConnection.signalingState !== 'closed') {
-    mediaStream.getTracks().forEach(function(track) {
-      this._peerConnection.addTrack(track, mediaStream);
-    }, this);
+PeerConnectionV2.prototype.addMediaTrackSender = function addMediaTrackSender(mediaTrackSender, mediaStream) {
+  if (this._peerConnection.signalingState === 'closed' || this._mediaSenders.has(mediaTrackSender)) {
+    return;
   }
+  var sender = this._peerConnection.addTrack(mediaTrackSender.track, mediaStream);
+  mediaTrackSender.addSender(sender);
```
markandrus: Cool 👍

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -135,8 +140,11 @@ PeerConnectionManager.prototype._getOrCreate = function _getOrCreate(id, configu
         self._closedPeerConnectionIds.add(peerConnection.id);
       }
     });
-    peerConnection.addMediaStream(this._localMediaStream);
+
     this._dataTrackSenders.forEach(peerConnection.addDataTrackSender, peerConnection);
+    this._mediaTrackSenders.forEach(function(mediaTrackSender) {
+      peerConnection.addMediaTrackSender(mediaTrackSender, this._localMediaStream);
```
markandrus: IMO, it would be nice if the APIs for `addDataTrackSender` and `addMediaTrackSender` could be made even more consistent by losing the MediaStream parameter. WDYT about passing `_localMediaStream` at construction time to the PeerConnectionV2? PeerConnectionV2 can then be responsible for using it when it calls `addTrack` internally.

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -183,13 +191,13 @@ PeerConnectionManager.prototype.createAndOffer = function createAndOffer() {
...
 
 /**
- * Get the remote MediaStreamTracks and {@link DataTrackReceiver}s of all
+ * Get the {@link DataTrackReceiver}s and {@link MediaTrackReceiver}s of all
  * the {@link PeerConnectionV2}s.
- * @returns {Array<MediaStreamTrack|DataTrackReceiver>} mediaStreamTracksAndDataTrackReceivers
+ * @returns {Array<DataTrackReceiver|MediaTrackReceiver>} trackReceivers
  */
-PeerConnectionManager.prototype.getRemoteMediaStreamTracksAndDataTrackReceivers = function getRemoteMediaStreamTracksAndDataTrackReceivers() {
+PeerConnectionManager.prototype.getTrackReceivers = function getTrackReceivers() {
```
markandrus: 👍

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -313,27 +317,34 @@ function createDummyAudioMediaStreamTrack(audioContext) {
...
     || changes.media.add.size
     || changes.media.remove.size) {
     peerConnectionManager._peerConnections.forEach(function(peerConnection) {
-      peerConnection.removeMediaStream(this._localMediaStream);
       changes.data.remove.forEach(peerConnection.removeDataTrackSender, peerConnection);
+      changes.media.remove.forEach(peerConnection.removeMediaTrackSender, peerConnection);
+    });
+
+    changes.media.remove.forEach(function(mediaTrackSender) {
+      this._localMediaStream.removeTrack(mediaTrackSender.track);
```
markandrus: I would really like us to be able to get rid of this. If the logic for handling the MediaStream could move into PeerConnectionV2, would this simplify things?

I know you had some hesitation on this, due to our bug with too many MediaStreams, but the Mozilla team confirmed it was because we ended up created tens of thousands or more... 1 MediaStream per PeerConnectionV2, esp. if we clean them up, should be safe, IMO.

markandrus:
@manjeshbhargav really awesome refactor 👍  great work

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus We actually store `RTCDataChannel`s in this collection. So isn't this a more appropriate name?

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus We actually store `RTCRtpSender`s in this collection. So isn't this a more appropriate name?

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus I though this should fire at least in Chrome? We don't really remove the tracks in FF/Safari.

lib/signaling/v2/peerconnection.js
markandrus: OK, good point 👍

lib/signaling/v2/peerconnection.js
markandrus: OK, can we call this `_rtpSenders` then?

lib/signaling/v2/peerconnection.js
markandrus: Per [mediacapture-main](https://w3c.github.io/mediacapture-main/),

> When a MediaStreamTrack track ends for any reason other than the stop() method being invoked, the User Agent must queue a task that runs the following steps:
>
> ...
>
> 4. Fire a simple event named ended at the object.

"ended" should not fire if `stop` is called. It's OK to have this if we are clear about the scenario we are using it for; however, it may not be necessary.

lib/util/sdp/trackmatcher/identity.js
```diff
@@ -0,0 +1,38 @@
...
+IdentityTrackMatcher.prototype.match = function match(event) {
+  return event.track.id;
+};
+
+/**
+ * Update the {@link IdentityTrackMatcher} with a new SDP.
+ * @param {string} sdp
+ */
+IdentityTrackMatcher.prototype.update = function update(sdp) {
+  this._sdp = sdp;
```
markandrus: We can ignore the SDP argument and skip the `defineProperties` call, IMO. The idea of IdentityTrackMatcher is that it's stateless and that it's just the "identity function" specialized to MediaStreamTrack ID.

lib/util/sdp/trackmatcher/mid.js
```diff
@@ -0,0 +1,47 @@
...
+      value: new Map(),
+      writable: true
+    }
+  });
+}
+
+/**
+ * Match a given MediaStreamTrack with its ID.
+ * @param {RTCTrackEvent} event
+ * @returns {Track.ID}
```
markandrus: Should be nullable, right?

lib/util/sdp/trackmatcher/mid.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+MIDTrackMatcher.prototype.update = function update(sdp) {
+  var sections = getMediaSections(sdp, '(audio|video)');
+  this._midToTrackIds = sections.reduce(function(midToTrackIds, section) {
+    var midMatches = section.match(/^a=mid:(.+)$/m) || [];
+    var trackIdMatches = section.match(/^a=msid:.+ (.+)$/m) || [];
+    var mid = midMatches[1];
+    var trackId = trackIdMatches[1];
+    return mid && trackId ? midToTrackIds.set(mid, trackId) : midToTrackIds;
+  }, this._midToTrackIds);
```
markandrus: Nitpick, but we are mapping to a singular `trackId`, right? So should the collection and private property here be `_midToTrackId`? IDK if we are very strict on this pattern throughout the codebase. Just a thought...

lib/util/sdp/trackmatcher/ordered.js
```diff
@@ -0,0 +1,122 @@
...
+      enumerable: true,
+      value: create()
+    }
+  });
+}
+
+/**
+ * Attempt to match a new MediaStreamTrack ID.
+ * @param {RTCTrackEvent} event
+ * @returns {?ID} id
```
markandrus: I saw `Track.ID` above--we should be consistent in how we refer to these.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @markandrus So when is "ended" fired on a remote MediaStreamTrack?

lib/util/sdp/trackmatcher/mid.js
manjeshbhargav: @markandrus I think we follow this nomenclature: `midsToTrackIds`. Since we're describing a collection, I think the name should be plural. WDYT?

manjeshbhargav:
@markandrus Closing this PR to reopen against master.

PR #223, opened 2017-11-29
Implement workaround for Issue 8329

markandrus:
_Not ready to merge._

https://bugs.chromium.org/p/webrtc/issues/detail?id=8329

### DIFF
--- a/lib/util/sdp/index.js
+++ b/lib/util/sdp/index.js
@@ -4,6 +4,16 @@ var difference = require('../').difference;
 var flatMap = require('../').flatMap;
 var setSimulcastInMediaSection = require('./simulcast');
 
+/**
+ * A payload type
+ * @typedef {number} PT
+ */
+
+/**
+ * An {@link AudioCodec} or {@link VideoCodec}
+ * @typedef {AudioCodec|VideoCodec} Codec
+ */
+
 // NOTE(mmalavalli): This value is derived from the IETF spec
 // for JSEP, and it is used to convert a 'b=TIAS' value in bps
 // to a 'b=AS' value in kbps.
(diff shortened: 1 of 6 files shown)


lib/util/sdp/index.js
```diff
@@ -31,23 +41,28 @@ function createBLine(modifier, maxBitrate) {
...
+/**
+ * Create a Map from PTs to codec names for the given m= section.
+ * @param {string} mediaSection - The given m= section.
+ * @returns {Map<PT, Codec>} ptToCodecName
+ */
+function createPtToCodecName(mediaSection) {
+  return getPayloadTypesInMediaSection(mediaSection).reduce(function(ptToCodecName, pt) {
+    var rtpmapPattern = new RegExp('a=rtpmap:' + pt + ' ([^/]+)');
+    var matches = mediaSection.match(rtpmapPattern);
+    var codecName = matches ? matches[1].toLowerCase() : '';
```
manjeshbhargav: @markandrus Why are we adding an empty string for the payload type whose codec name doesn't exist?

lib/util/sdp/issue8329.js
```diff
@@ -0,0 +1,205 @@
...
+ * A Set with at least one element
+ * @typedef {Set} NonEmptySet
+ */
+
+/**
+ * Apply the workaround for Issue 8329 to an RTCSessionDescriptionInit.
+ * @param {RTCSessionDescriptionInit} description
+ * @returns {RTCSessionDescription} newDescription
+ */
+function workaround(description) {
```
manjeshbhargav: @markandrus Can we simplify this:
```javascript
function workaroud(description) {
  var descriptionInit = { type: description.type };
  if (description.type !== 'rollback') {
    descriptionInit.sdp = sdpWorkaround(description.sdp);
  }
return new RTCSessionDescription(descriptionInit);
}
```

lib/util/sdp/issue8329.js
```diff
@@ -0,0 +1,205 @@
...
+  var unassociatedRtxPts = Array.from(invalidRtxPts);
+
+  // NOTE(mroberts): We normalize to lowercase.
+  var knownCodecNames = ['h264', 'vp8', 'vp9'];
+  var unassociatedPts = knownCodecNames.reduce(function(unassociatedPts, codecName) {
+    var pts = codecNameToPts.get(codecName) || new Set();
+    return Array.from(pts).reduce(function(unassociatedPts, pt) {
+      return associatedPtToRtxPt.has(pt)
+        ? unassociatedPts
+        : unassociatedPts.add(pt);
```
manjeshbhargav: @markandrus Are you sure `Set#add()` returns the `Set`?

lib/util/sdp/index.js
markandrus: Only dynamic PTs are required to have "a=rtpmap" entries—there are some that do not require these attributes. I guess we could/should look up their names statically, though? We don't really need/use them, though. The Paper doc where I describe this mentions letting the `codecName` be the empty string, though.

lib/util/sdp/issue8329.js
markandrus: Good idea 👍

lib/util/sdp/issue8329.js
markandrus: Yes https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add

manjeshbhargav:
@markandrus Couple of nitpicks, but looks good otherwise.

PR #231, opened 2017-12-07
Fix negotiation loop in Firefox

markandrus:
PeerConnectionV2 had a bug that I noticed when testing Firefox Stable/Beta to Firefox Stable/Beta. The bug is this: in order to decide whether to renegotiate, we always check if the current offer (local _or_ remote) contains at least as many media sections with directions "sendrecv" or "recvonly" as RTCRtpSenders we have (taking into account kind).

However, this doesn't work in the local case with Firefox Stable/Beta. Consider if we have two local audio tracks, we've been offered one remote audio track in a previous round of negotiation, and now we call `createOffer`:

* We have two RTCRtpSenders (one for each of our two local audio tracks)
* We have two m=sections, one with "a=sendrecv" and one with "a=sendonly"

But our check is for "sendrecv" or "recvonly", and so our logic would say to re-negotiate. This can lead to a loop! Here is a fix: instead of checking the current offer (local _or_ remote), we should check the current local description (offer _or_ answer) and "sendrecv" or _"sendonly"_ (instead of "recvonly").

Firefox Nightly (with RTCRtpTransceiver support) isn't triggering this, since the spec says the second media section should have "a=sendrecv", rather than "a=sendonly". This loop may have been slowing down (or failing) our tests—hopefully this fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Bug Fixes
   RTCDataChannel's `binaryType` to "arraybuffer" in order to ensure consistent
   behavior across browsers. (JSDK-1627)
 - We always stringify `name`s passed via LocalTrackOptions now. (JSDK-1565)
+- Fixed some code that could lead to a renegotiation loop in Firefox.
 
 1.6.0 (October 24, 2017)
 ========================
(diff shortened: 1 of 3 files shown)


manjeshbhargav:
@markandrus LGTM. Just one question: Is there any danger of calling `_maybeReoffer()` with a `rollback` `RTCSessionDescription`? In that case, the `.sdp` will be `null`.

markandrus:
`localDescription` can only ever be offer, answer, or null.

PR #239, opened 2017-12-15
JSDK-1599: Track subscription errors

markandrus:
_Not ready to merge._

The point of this PR is to allow raising "trackSubscriptionFailed" events in the event that, for example,

1. A VP8-only Participant connects to a Group Room that supports VP8 and H.264.
2. An H.264-only Participant connects to the same Group Room and publishes a VideoTrack.

Once Step 2 occurs, we want the VP8-only Participant to learn that there was a VideoTrack they ought to have subscribed to, but couldn't, because they didn't support the required codec.

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -384,12 +384,11 @@ function periodicallyPublishStats(roomV2, localParticipant, transport, intervalM
 }
 
 function handleSubscriptionFailures(room) {
-  var remoteTracks = Array.from(room.participants.values()).reduce(function(remoteTracks, participant) {
-    participant.tracks.forEach(function(track) {
-      remoteTracks.set(track.sid, track);
+  var remoteTracks = new Map(util.flatMap(room.participants, function(participant) {
+    return Array.from(participant.tracks.values()).map(function(track) {
+      return [track.sid, track];
     });
-    return remoteTracks;
-  }, new Map());
+  }));
 
   room._subscriptionFailures.forEach(function(error, trackSid) {
     var remoteTrack = remoteTracks.get(trackSid);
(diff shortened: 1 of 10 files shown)


markandrus:
I'd like to add more robust integration tests in a subsequent sprint. It will take a bit of time, though, since we need to be able to

* hit the REST API to create a Room with a particular codec, and then
* munge the SDP locally to omit the specified codec (to simulate lack of support).

In the meantime, I did a manual test with a build of [draw-with-twilio.appspot.com](https://draw-with-twilio.appspot.com) pointing to this branch:

1. Set the draw-with-twilio account's Room Settings to support VP8 only.
2. Let Safari (H.264-only) connect to a Room.
3. Run the following in the console:

  ```js
  room.on('participantConnected', participant =>
    participant.on('trackSubscriptionFailed', (error, trackPublication) =>
      console.error('Failed to subscribe to RemoteTrack %s with name "%s": %s',
        trackPublication.trackSid, trackPublication.trackName, error.message)))
  ```
4. Let Chrome or Firefox connect to the Room.
5. Confirm that you see the following error print in Safari's console:

  > Failed to subscribe to RemoteTrack MT7a5aaddd26aa7148b28ebd7e10d11156 with name "f8a315e7-d2c9-44a2-ae9c-9fcd0e2ea306": No supported codec

lib/signaling/v2/room.js
```diff
@@ -372,4 +383,20 @@ function periodicallyPublishStats(roomV2, localParticipant, transport, intervalM
   });
 }
 
+function handleSubscriptionFailures(room) {
+  var remoteTracks = Array.from(room.participants.values()).reduce(function(remoteTracks, participant) {
```
manjeshbhargav: @markandrus This can be re-written as:
```javascript
var remoteTracks = new Map(flatMap(room.participants, participant => {
  return flatMap(participant.tracks, track => [track.sid, track]);
}));
```
?

manjeshbhargav:
@markandrus On comment, LGTM otherwise.

lib/signaling/v2/room.js
markandrus: good idea 👍

PR #243, opened 2017-12-21
Cross-browser testing framework.

manjeshbhargav:
@markandrus 

This PR contains:

* `TestDriver` interface definition.
* `Transport` interface definition.
* `WSClientTransport` and `WSServerTransport` implementation and tests.
* `DockerBrowserDriver` and `SeleniumBrowserDriver` implementation.
* `DMP` and `DMPRequest` implementation and tests.
* Module (that runs in the browser) which loads the SDK and creates a DMP client.
* Integration test to test the complete setup.


test/lib/sdkdriver/src/transport/index.d.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * A {@link Transport} represents an interface for communicating
+ * between the mocha.js process that drives a cross-browser test
```
markandrus: This is fine for now, but in the future I think we can be more generic. Although we'll initially use Mocha, it's not strictly necessary and this could be used with other things.

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Close the {@link WSClientTransport}.
+   * @returns {void}
+   */
+  close(): void {
+    if (!this._wsClient) {
+      return;
+    }
+    try {
+      this._wsClient.close();
```
markandrus: I think this is only supposed to throw if `code` or `reason` are provided: https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket-close

Do we need the `try`/`catch`?

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Open the {@link WSClientTransport}.
+   * @returns {Promise<void>}
+   */
+  open(): Promise<void> {
+    const { WebSocket } = this._deps;
+    if (!this._wsClient) {
+      this._wsClient = new WebSocket(this._wsUrl);
+    }
+    const wsClient: WebSocket = this._wsClient as WebSocket;
```
markandrus: writing this without the `if` statement may allow omitting the `as WebSocket` cast:

```ts
const wsClient = this._wsClient || new WebSocket(this._wsUrl);
this._wsClient = wsClient;
```

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+        wsClient.removeEventListener('close', onopenfailed);
+        reject({ code, message });
+      };
+      wsClient.addEventListener('close', onopenfailed);
+
+      wsClient.addEventListener('open', () => {
+        this._sendBuffer.splice(0).forEach(item => wsClient.send(item));
+        wsClient.removeEventListener('close', onopenfailed);
+        wsClient.addEventListener('close', () => this.emit('close'));
+        wsClient.addEventListener('message', event => this.emit('message', JSON.parse(event.data)));
```
markandrus: Are we always receiving JSON strings?

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+        resolve();
+      });
+    });
+  }
```
markandrus: I guess if you were to call `close` while this Promise is pending, it would remain pending. Perhaps this is OK—I think WebRTC APIs act this way when you close the RTCPeerConnection.

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Send data if {@link WSClientTransport} is open, enqueue otherwise.
```
markandrus: Technically there is a third-case: if `!this._wsClient`, drop. Is that how we want to handle those messages?

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+   * @param {*} data
+   * @returns {void}
+   */
+  send(data: any): void {
+    if (!this._wsClient) {
+      return;
+    }
+    const { WebSocket: { CONNECTING, OPEN } } = this._deps;
+    const wsClient: WebSocket = this._wsClient as WebSocket;
+    const { readyState } = wsClient;
```
markandrus: Since we only use `readyState` once below, maybe we don't need the destructuring assignment. Typically I use these if I access a property member more than once. WDYT?

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Send data if {@link WSClientTransport} is open, enqueue otherwise.
+   * @param {*} data
+   * @returns {void}
+   */
+  send(data: any): void {
+    if (!this._wsClient) {
+      return;
+    }
+    const { WebSocket: { CONNECTING, OPEN } } = this._deps;
+    const wsClient: WebSocket = this._wsClient as WebSocket;
```
markandrus: If we nest the whole block in

```ts
if (this._wsClient) {
  // Do stuff
}
```

then I think TypeScript should be smart enough to know we did the null-check.

test/lib/sdkdriver/test/unit/spec/transport/websocket/client.ts
```diff
@@ -0,0 +1,157 @@
...
+
+describe('WSClientTransport', () => {
+  describe('#close', () => {
+    let closeEventPromise: Promise<void>;
+    let transport: WSClientTransport;
+    let wsClient: WebSocket;
+
+    before(async () => {
+      transport = new WSClientTransport('foo', { WebSocket });
+      const openPromise: Promise<void> = transport.open();
```
markandrus: Can't TypeScript infer the type here?

test/lib/sdkdriver/test/unit/spec/transport/websocket/client.ts
```diff
@@ -0,0 +1,157 @@
...
+      wsClient = transport._wsClient;
```
markandrus: Would prefer if we accessed the `wsClient` from the WebSocket mock we inject instead of the private property on the WSClientTransport, since the latter could change.

test/lib/sdkdriver/test/unit/spec/transport/websocket/client.ts
```diff
@@ -0,0 +1,157 @@
...
+import * as assert from 'assert';
+import * as sinon from 'sinon';
+
+describe('WSClientTransport', () => {
+  describe('#close', () => {
+    let closeEventPromise: Promise<void>;
+    let transport: WSClientTransport;
+    let wsClient: WebSocket;
+
+    before(async () => {
```
markandrus: Should we `beforeEach` here?

test/lib/sdkdriver/test/unit/spec/transport/websocket/client.ts
```diff
@@ -0,0 +1,157 @@
...
+      {event: 'open', readyState: 'OPEN'},
+      {event: 'close', readyState: 'CLOSED'}
+    ].forEach(({ event, readyState }) => {
+      context(`when the WebSocket ${event === 'open' ? 'opens' : 'fails to open'}`, () => {
+        let promise: Promise<void>;
+        let wsClient: any;
+
+        beforeEach(() => {
+          transport._sendBuffer.push('foo');
+          transport._sendBuffer.push('bar');
```
markandrus: I'm surprised TypeScript allows this. `_sendBuffer` is private. Can we set this test up in a way where we rely only on public API?

.gitignore
```diff
@@ -4,6 +4,7 @@ browsers
 coverage
 dist
 node_modules
+test/lib/sdkdriver/lib
```
markandrus: Additionally, we may want to add this to `.npmignore`, since we probably don't want to publish the TypeScript files. (Or is `test/` already in `.npmignore`?).

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -0,0 +1,100 @@
+import Transport from '../';
+import { EventEmitter } from 'events';
+import * as WebSocket from 'ws';
+
+/**
+ * WebSocket client {@link Transport}.
+ * @class
```
markandrus: `@class` isn't necessary, is it?

test/lib/sdkdriver/src/transport/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+ * and the process in which the test is running.
+ * @interface
```
markandrus: `@interface` isn't necessary, is it?

test/lib/sdkdriver/src/transport/websocket/server.ts
```diff
@@ -0,0 +1,85 @@
+import Transport from '../';
+import { EventEmitter } from 'events';
+import * as WebSocket from 'ws';
+
+/**
+ * WebSocket server {@link Transport}.
+ * @class
```
markandrus: `@class` isn't necessary, is it?

test/lib/sdkdriver/src/transport/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+ * The {@link Transport} was closed.
+ * @event Transport#close
+ * @param {Error} [error] - Optional Error, if closed unintentionally
+ */
+
+/**
+ * The {@link Transport} received a message.
+ * @event Transport#message
+ * @param {object} data
+ */
```
markandrus: It'd be great to include these as part of the interface, but I'm not sure current best practices on typing EventEmitters.

test/lib/sdkdriver/src/transport/websocket/server.ts
```diff
@@ -0,0 +1,85 @@
...
+  private readonly  _deps: any;
+  private readonly _sendBuffer: Array<string>;
+  private readonly _wsServer: WebSocket.Server;
+  private _wsConnection: WebSocket | null;
+
+  /**
+   * Constructor.
+   * @param {number} wsPort - The WebSocket server port.
+   * @param {TestDriver} testDriver - The {@link TestDriver} that connects
+   *   to the {@link WSServerTransport}
```
markandrus: This shouldn't be here, should it?

test/lib/sdkdriver/src/transport/websocket/server.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Close the {@link WSServerTransport}.
+   * @returns {void}
+   */
+  close(): void {
+    if (this._wsConnection) {
+      try {
+        this._wsConnection.close();
```
markandrus: I wonder how this throws—otherwise, I guess we are being safe.

test/lib/sdkdriver/src/transport/websocket/server.ts
```diff
@@ -0,0 +1,85 @@
...
+      } catch (e) {
+        // Do nothing.
+      }
+      this._wsConnection = null;
+    }
+
+    try {
+      this._wsServer.close();
```
markandrus: Should we gate on `if (this._wsServer)` here?

test/lib/sdkdriver/test/unit/spec/transport/websocket/server.ts
```diff
@@ -0,0 +1,152 @@
...
+    let closeEventPromise: Promise<void>;
+    let transport: WSServerTransport;
+    let wsConnection: WebSocket;
+    let wsServer: WebSocket.Server;
+
+    before(async () => {
+      transport = new WSServerTransport(2000, { WebSocket });
+      const openPromise: Promise<void> = transport.open();
+      wsConnection = new WebSocket('foo');
+      wsServer = transport._wsServer;
```
markandrus: I feel similarly about accessing `_wsServer` as I do `_wsClient` in the other tests.

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
+import { EventEmitter } from 'events';
+import Transport from './transport';
+
+/**
+ * DMP Request.
+ * @class
```
markandrus: Necessary?

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
...
+  public readonly data: any;
+
+  /**
+   * Constructor.
+   * @param {number} id - Request ID
+   * @param {*} data - Request data
+   * @param {Transport} transport - The {@link Transport} used for
+   *   sending a response
+   */
+  constructor(id: number, data: any, transport: Transport) {
```
markandrus: I believe you can just do

```
constructor(private readonly _id: number, public readonly data: any, private readonly transport: Transport) {}
```

however, then you don't get to have the nice parameter names—just the underscore prefixed ones. So I guess we ought to leave as-is.

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.data = data;
+  }
+
+  /**
+   * Send a response over the underlying {@link Transport}.
+   * @param {*} data - Response data
+   * @returns {void}
+   */
+  sendResponse(data: any): void {
+    const { _id: id, _transport: transport } = this;
```
markandrus: If we're only accessing `_id` and `_transport` once, maybe the destructuring assignment doesn't buy us much?

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
...
+import Transport from './transport';
+
+/**
+ * DMP Request.
+ * @class
+ * @classdesc A {@link DMPRequest} contains the data associated with
+ *   the incoming request message along with a method for sending a
+ *   response.
+ */
+class DMPRequest {
```
markandrus: we need to export this, right?

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  sendResponse(data: any): void {
+    const { _id: id, _transport: transport } = this;
+    transport.send({ data, id, type: 'response' });
+  }
+}
+
+/**
+ * Driver Messaging Protocol (DMP).
+ * @class
```
markandrus: Necessary?

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,132 @@
...
+      const onresponse: (response: any) => void = response => {
+        const { id, type } = response;
+        if (type === 'response' && id === requestId) {
+          this._transport.removeListener('message', onresponse);
+          resolve(response.data);
+        }
+      };
+
+      this._transport.send({ data, id: requestId, type: 'request' });
+      this._transport.on('message', onresponse);
```
markandrus: There's one potential issue with this: the moment I have 10 outstanding requests or so, by default, Node's EventEmitter is going to warn about a potential leak. We can override this with [`setMaxListeners`](https://nodejs.org/api/events.html#events_emitter_setmaxlisteners_n)

test/lib/sdkdriver/src/browser.ts
```diff
@@ -0,0 +1,39 @@
...
+
+function getUrlParams(): Map<string, string> {
+  const serializedParams: string = window.location.search.substring(1);
+  return serializedParams.split('&').reduce((params, nvpair) => {
+    const [ name, value ] = nvpair.split('=').map(decodeURIComponent);
+    return params.set(name, value);
+  }, new Map());
+}
+
+function loadSdk(url: string): Promise<void> {
```
markandrus: nitpick: although this is fine for now, we should maybe be more generic. we could load an arbitrary module instead of an sdk.

test/lib/sdkdriver/src/browser.ts
```diff
@@ -0,0 +1,39 @@
...
+  const script: any = document.createElement('script');
+  script.src = url;
+  document.body.appendChild(script);
+  return new Promise(resolve => {
+    script.onload = () => resolve();
+  }).then(() => {});
```
markandrus: You could just `await new Promise` instead of chaining the `then`.

test/lib/sdkdriver/src/browser.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+(async () => {
+  const urlParams: Map<string, string> = getUrlParams();
+  const sdkUrl: string = urlParams.get('sdkUrl') as string;
+  const wsUrl: string = urlParams.get('wsUrl') as string;
+  const transport: WSClientTransport = new WSClientTransport(wsUrl);
+
+  await transport.open();
+  await loadSdk(sdkUrl);
```
markandrus: Should we flip the order here? Maybe `loadSdk` loads a dependency required for handling DMP messages? And so it should load before opening the transport?

test/lib/sdkdriver/src/browser.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const dmp: DMP = new DMP(transport);
+  dmp.sendEvent({ sdkVersion: window['Twilio']['Video']['version'] });
```
markandrus: Seems too specific? Also, I imagine that, instead of an "SDK", we actually load a module that will use the DMP instance in a particular way. So, rather than hardcode this here, I think we should pass in a module like

```ts
const { dmp: DMP } = window
dmp.sendEvent({ /* ... */ })
dmp.on('request', (request: DMPRequest) => {
  // ...
})
```

or, rather than depending on `window`, maybe the module should expose a function, e.g.,

```ts
export default function main(dmp: DMP) {
  // ...
}
```

test/lib/sdkdriver/src/testdriver/browser/base.ts
```diff
@@ -0,0 +1,89 @@
...
+    this._webServer = createWebServer();
+  }
+
+  /**
+   * Get the web server.
+   * @returns {HTTPServer}
+   */
+  get webServer(): HTTPServer {
+    return this._webServer;
+  }
```
markandrus: Does this need to be public?

test/lib/sdkdriver/src/testdriver/browser/base.ts
```diff
@@ -0,0 +1,89 @@
...
+    }).join('&');
+
+    return this.startBrowser(`http://${host}:${port}/?${serializedParams}`);
+  }
+
+  /**
+   * Start the web server.
+   * @returns {Promise<void>}
+   */
+  startWebServer(): Promise<void> {
```
markandrus: Should this be private?

test/lib/sdkdriver/src/testdriver/browser/base.ts
```diff
@@ -0,0 +1,89 @@
...
+    return new Promise(resolve => this._webServer.listen(0, resolve));
+  }
+
+  /**
+   * Stop the web server.
+   * @returns {void}
+   */
+  stopWebServer(): void {
```
markandrus: Should this be private?

test/lib/sdkdriver/src/testdriver/browser/base.ts
```diff
@@ -0,0 +1,89 @@
...
+  private readonly _webServer: HTTPServer;
+  protected readonly _options: BrowserDriverOptions;
+  abstract startBrowser(url: string): Promise<void>;
+  abstract stopBrowser(): void;
+
+  /**
+   * Constructor.
+   * @param {BrowserDriverOptions} options
+   */
+  constructor(options: BrowserDriverOptions) {
```
markandrus: Probably `sdkUrl` (or whatever we end up calling it) needs to be passed here?

markandrus: I'd prefer it were more strongly typed than `params: any`, WDYT?

test/lib/sdkdriver/src/testdriver/browser/docker.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+    this._docker = null;
+  }
+
+  /**
+   * Start the browser process.
+   * @param {string} url - URL for running the SDK setup code
+   * @returns {Promise<void>}
+   */
+  startBrowser(url: string): Promise<void> {
```
markandrus: Can this be private?

test/lib/sdkdriver/src/testdriver/browser/docker.ts
```diff
@@ -0,0 +1,56 @@
...
+      });
+    });
+  }
+
+  /**
+   * Stop the browser process.
+   * @returns {void}
+   */
+  stopBrowser(): void {
+    this._docker.kill();
```
markandrus: `if (this._docker)`?

test/lib/sdkdriver/src/testdriver/browser/base.ts
```diff
@@ -0,0 +1,89 @@
...
+  startWebServer(): Promise<void> {
+    return new Promise(resolve => this._webServer.listen(0, resolve));
+  }
+
+  /**
+   * Stop the web server.
+   * @returns {void}
+   */
+  stopWebServer(): void {
+    this._webServer.close();
```
markandrus: Should we `if (this._webServer)` and `null` it out on close?

test/lib/sdkdriver/src/testdriver/browser/selenium.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+    this._webDriver = createWebDriver(browser);
+  }
+
+  /**
+   * Start the browser process.
+   * @param {string} url - URL for running the SDK setup code
+   * @returns {Promise<void>}
+   */
+  async startBrowser(url: string): Promise<void> {
```
markandrus: Actually, I guess this could be made `protected`?

test/lib/sdkdriver/test/integration/spec/setup.ts
```diff
@@ -0,0 +1,65 @@
+import * as assert from 'assert';
+import DMP from '../../../src/dmp';
+import TestDriver from '../../../src/testdriver';
+import BrowserDriver from '../../../src/testdriver/browser';
+import WSServerTransport from '../../../src/transport/websocket/server';
+
+describe('setup', function() {
+  this.timeout(60000);
+
+  ['chrome', 'firefox'].forEach(browser => {
```
markandrus: Nice 👍

test/lib/sdkdriver/src/transport/websocket/client.ts
manjeshbhargav: @markandrus Yes.

test/lib/sdkdriver/src/transport/websocket/client.ts
manjeshbhargav: @markandrus When the WebSocket is closed while the Promise is pending, the `onclose` handler should reject the Promise.

test/lib/sdkdriver/src/transport/websocket/client.ts
manjeshbhargav: @markandrus This is what the method does now.

test/lib/sdkdriver/src/transport/websocket/client.ts
manjeshbhargav: @markandrus I think we should use it wherever it can result in less verbose code.

test/lib/sdkdriver/test/unit/spec/transport/websocket/client.ts
manjeshbhargav: @markandrus The tests specified within the `it()` functions here are independent enough of each other that I don't think we need to re-run this for each of them.

.gitignore
manjeshbhargav: @markandrus `test/` is already in `.npmignore`.

test/lib/sdkdriver/src/transport/websocket/server.ts
manjeshbhargav: @markandrus `._wsServer` is never null.

test/lib/sdkdriver/src/dmp.ts
manjeshbhargav: @markandrus I think we should use destructuring wherever it decreases verbosity.

test/lib/sdkdriver/src/dmp.ts
manjeshbhargav: @markandrus I don't think so, since the constructor is called internally when we are emitting a "request" event on DMP.

test/lib/sdkdriver/src/browser.ts
manjeshbhargav: @markandrus This file will change in the future to be SDK/API agnostic. So I'd suggest we leave this as it is for now in order to test the setup.

test/lib/sdkdriver/src/testdriver/browser/base.ts
manjeshbhargav: @markandrus We need to pass this to the `WSServerTransport` after it is started.

test/lib/sdkdriver/src/testdriver/browser/base.ts
manjeshbhargav: @markandrus TypScript does not allow `abstract` methods to be `private`. So I have to override with public methods.

test/lib/sdkdriver/src/testdriver/browser/base.ts
manjeshbhargav: @markandrus TypScript does not allow `abstract` methods to be `private`. So I have to override with public methods.

test/lib/sdkdriver/src/testdriver/browser/base.ts
manjeshbhargav: @markandrus params can be any sdk specific thing, right?

test/lib/sdkdriver/src/testdriver/browser/docker.ts
manjeshbhargav: @markandrus TypScript does not allow `abstract` methods to be `private`. So I have to override with public methods.

test/lib/sdkdriver/src/testdriver/browser/docker.ts
manjeshbhargav: @markandrus The `._docker` property is never null.

test/lib/sdkdriver/src/testdriver/browser/selenium.ts
manjeshbhargav: @markandrus TypScript does not allow `abstract` methods to be `protected`. So I have to override with public methods.

test/lib/sdkdriver/src/transport/websocket/client.ts
markandrus: I re-read, and indeed onclose fires for these cases.

test/lib/sdkdriver/src/dmp.ts
markandrus: How does a consumer of `DMP` use the `DMPRequest`, though? They want to be able to write something like this, right?

```ts
import DMP, { DMPRequest } from './dmp'

const dmp = new DMP(/* ... */)
dmp.on('request', (request: DMPRequest) => {
  // ...
})
```

test/lib/sdkdriver/src/testdriver/browser/docker.ts
markandrus: It gets nulled out below, though. What happens if we call `stopBrowser` twice?

test/lib/sdkdriver/src/testdriver/browser/base.ts
manjeshbhargav: @markandrus What if we want to restart the `._webServer`?

test/lib/sdkdriver/src/testdriver/browser/base.ts
markandrus: Isn't there some set of parameters that aren't SDK-specific, though? For example, we want the BrowserDriver to load a particular module that will exercise the DMP and SDK modules. How do we provide that? Maybe there are already some subsequent commits that address this. Will keep reading

test/lib/sdkdriver/src/testdriver/browser/base.ts
markandrus: OK

test/lib/sdkdriver/src/transport/websocket/client.ts
```diff
@@ -27,22 +26,18 @@ export default class WSClientTransport extends EventEmitter implements Transport
     this._sendBuffer = [];
     this._wsClient = null;
     this._wsUrl = wsUrl;
+    this.setMaxListeners(Infinity);
```
markandrus: 👍

test/lib/sdkdriver/src/util/index.ts
```diff
@@ -31,10 +31,11 @@ export function createWebDriver(browser: 'chrome' | 'firefox'): WebDriver {
...
 /**
  * Create an express web server that serves the SDK setup code.
+ * @param {string} rootPath - Root folder of the web server
  * @returns {HTTPServer}
  */
-export function createWebServer(): HTTPServer {
+export function createWebServer(rootPath: string): HTTPServer {
   const app: express.Application = express();
-  app.use(express.static(join(__dirname, '..', '..', 'browser')));
+  app.use(express.static(rootPath));
```
markandrus: OK, so the idea is that we pass the `rootPath` and serve things up via Express `static` middleware, not unlike how we do it in our E2E tester? And so, at build time, we'll want to Browserify the necessary modules?

test/lib/sdkdriver/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+   * @param {TestDriver} testDriver
+   * @returns {Promise<SDKDriver>}
+   */
+  static async create(transport: Transport, testDriver: TestDriver): Promise<SDKDriver> {
+    await Promise.all([
+      transport.open(),
+      testDriver.open()
+    ]);
+    return new SDKDriver(transport, testDriver);
+  }
```
markandrus: Any reason not to have an `open` method that does this instead on SDKDriver?

test/lib/sdkdriver/src/dmp.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Driver Messaging Protocol (DMP).
+ * @classdesc A {@link DMP} exchanges DMP messages with its counterpart;
+ *   It can be instantiated in the mocha.js process that drives the test
+ *   process as well as in the test process
+ * @fires DMP#event
+ * @fires DMP#request
+ */
+export default class DMP extends EventEmitter {
```
markandrus: Should the DMP re-emit the underlying Transport's "close" event?

package.json
```diff
@@ -47,7 +52,10 @@
     "selenium-webdriver": "3.3.0",
     "sinon": "^4.0.1",
     "travis-multirunner": "^4.0.0",
+    "ts-node": "^4.0.1",
+    "tslint": "^5.8.0",
     "twilio": "^2.11.1",
+    "typescript": "^2.6.2",
```
markandrus: I think you should pin to a specific version of TypeScript, since, from what I understand, they do not follow SemVer.

markandrus: Also, I'm not sure if this has implications for ts-node and/or tslint.

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+ * @param {Array<*>} args
+ * @param {(room: Room) => void} sendRoomEvents
+ * @returns {Promise<object>}
+ */
+export async function connect(args: any, sendRoomEvents: (room: any) => void): Promise<any> {
+  const [ token, options ] = args;
+  let room: any;
+
+  try {
+    room = await Twilio.Video.connect(token, options);
```
markandrus: maybe better to

```ts
Twilio.Video.connect.apply(null, args)
```

Also, maybe better to let `sendRoomEvents` come first?

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+  } catch (e) {
+    return {
```
markandrus: Can you just do

```ts
return { error }
```

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+        code: e.code,
+        message: e.message
+      }
+    };
+  }
+
+  add(room);
+  add(room.localParticipant);
+  room.localParticipant.tracks.forEach(add);
+  room.localParticipant.trackPublications.forEach(add);
```
markandrus: I'm not sure what `add` does yet, but could we let some LocalParticipant-specific version of `add` do the work of adding `localParticipant.tracks` and `localParticipant.trackPublications`? That way we aren't doing all this work in `connect`?

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  room.participants.forEach((participant: any) => {
+    add(participant);
+    participant.tracks.forEach(add);
```
markandrus: Similar comment as above—can we have some RemoteParticipant-specific `add` that does the work of adding `participant.tracks`?

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+    result: serializeRoom(room)
+  };
+}
+
+/**
+ * Create a {@link LocalTrack}.
+ * @param {Array<*>} args
+ * @returns {Promise<object>}
+ */
+export async function createLocalTrack(args: any): Promise<any> {
```
markandrus: I think we should consider a more-specific type than `any` here if we are going to do more with these `args` than just `apply` them to a twilio-video.js function.

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+  if (!room) {
+    return {
+      error: {
+        message: 'Room not found'
+      }
+    };
+  }
+  room.disconnect();
+  const serializedRoom = serializeRoom(room);
+  remove(room);
```
markandrus: Shouldn't there be a Room-specific `remove` that removes the LocalParticipant, RemoteParticipant, etc.

test/crossbrowser/src/browser/api.ts
```diff
@@ -0,0 +1,362 @@
...
+  const localParticipant: any = lookup(target);
+  if (!localParticipant) {
+    return {
+      error: {
+        message: 'Participant not found'
+      }
+    };
+  }
+
+  const localTrack: any = lookup(args[0]);
```
markandrus: We should type `args`, I think. Otherwise how do we know this contract is fulfilled? (i.e., `args[0]` is what we need it to be?).

test/crossbrowser/src/browser/events.ts
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * Send {@link LocalParticipant} events to the {@link SDKDriver}.
+ * @param {DMP} dmp
+ * @param {LocalParticipant} localParticipant
+ */
+function sendLocalParticipantEvents(dmp: DMP, localParticipant: any): void {
+  localParticipant.on('trackAdded', (track: any) => {
+    dmp.sendEvent({
+      args: [serializeLocalTrack(track)],
+      source: serializeLocalParticipant(localParticipant),
```
markandrus: I guess I though this would be the resource ID? But I guess there is a benefit to sending full state here in case anything changed?

test/crossbrowser/src/browser/resources.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param {*} resource
+ * @returns {void}
+ */
+export function add(resource: any): void {
+  const { lookup, reverseLookup } = resources;
+  if (!reverseLookup.has(resource)) {
+    const resourceId = createResourceId();
+    lookup.set(resourceId, resource);
+    reverseLookup.set(resource, resourceId);
+  }
```
markandrus: Pretty cool 👍

test/crossbrowser/src/browser/serialize.ts
```diff
@@ -0,0 +1,230 @@
...
+export function serializeLocalParticipant(localParticipant: any): any {
+  const {
+    audioTrackPublications,
+    dataTrackPublications,
+    trackPublications,
+    videoTrackPublications
+  } = localParticipant;
+
+  return {
+    audioTrackPublications: Array.from(audioTrackPublications.values()).map(serializeLocalTrackPublication),
```
markandrus: Nitpick: could use spread operator here, e.g.

```ts
[...audioTrackPublications.values()]
```

test/crossbrowser/src/videodriver/localparticipant.ts
```diff
@@ -0,0 +1,219 @@
...
+ * @property {Map<TrackSID, object>} videoTrackPublications
+ * @fires LocalParticipantDriver#trackPublicationFailed
+ * @fires LocalParticipantDriver#trackPublished
+ */
+export default class LocalParticipantDriver extends ParticipantDriver {
+  private readonly _sdkDriver: SDKDriver;
+  audioTrackPublications: Map<string, any>;
+  dataTrackPublications: Map<string, any>;
+  trackPublications: Map<string, any>;
+  videoTrackPublications: Map<string, any>;
```
markandrus: Are these any because we don't have a type of LocalTrackPublication yet?

test/crossbrowser/src/videodriver/room.ts
```diff
@@ -0,0 +1,358 @@
...
+  /**
+   * Disconnect from the {@link Room} in the browser.
+   * @returns {void}
+   */
+  disconnect(): void {
+    this._sdkDriver.sendRequest({
+      api: 'disconnect',
+      target: this._resourceId
+    }).then(() => {
+      // Do nothing.
```
markandrus: We can omit this branch and just `catch`.

package.json
```diff
@@ -65,6 +73,23 @@
     "test:integration:travis": "node ./scripts/integration.js",
     "test:integration": "node ./scripts/karma.js karma/integration.conf.js",
     "test:umd": "mocha ./test/umd/index.js",
+    "test:crossbrowser:build:clean": "rimraf ./test/crossbrowser/lib ./test/crossbrowser/src/browser/index.js",
```
markandrus: Can you change this script to

```
test:crossbrowser:clean
```

?

package.json
```diff
@@ -65,6 +73,23 @@
...
+    "test:crossbrowser:build:lint": "cd ./test/crossbrowser && tslint --project tsconfig.json",
+    "test:crossbrowser:build:tsc": "cd ./test/crossbrowser && tsc",
+    "test:crossbrowser:build:browser": "cd ./test/crossbrowser && browserify lib/crossbrowser/src/browser/index.js > src/browser/index.js",
+    "test:crossbrowser:build": "npm-run-all test:crossbrowser:build:*",
+    "test:crossbrowser:test": "cd ./test/crossbrowser && mocha --compilers ts:ts-node/register test/integration/spec/**/*.ts",
+    "test:crossbrowser": "npm-run-all test:crossbrowser:*",
+    "test:sdkdriver:build:clean": "rimraf ./test/lib/sdkdriver/lib ./test/lib/sdkdriver/test/integration/browser/index.js",
```
markandrus: Can you change this script to

```
test:sdkdriver:clean
```

PR #263, opened 2018-02-07
Convert to TypeScript

markandrus:
_Not ready to merge. Best viewed with `?w=1`, i.e. https://github.com/twilio/twilio-video.js/pull/263/files?w=1._

* [x] Use lebab to convert to ES7
* [x] Review FIXMEs related to Log and converting to `class`
* [x] Use `this` in fat arrow functions where previously we were using `self`
* [x] Review TwilioError class (there is some funny stuff with `stack` we should check on)
  * I noticed we could improve the `name` property for the various TwilioError subclasses (right now, they all read "TwilioError", instead of, e.g., "AccessTokenInvalidError", which is probably more useful). We can do this in a subsequent PR. Otherwise, `stack` and `message` look correct.
* [x] Patch LocalTrack constructors before exporting in order to maintain API compatibility *
* [x] Double-check JSDoc after lebab transforms.
* [ ] Convert .js to .ts
  * [ ] `git mv lib src`
  * [ ] Add `tsconfig.json`
* [ ] Convert JSDoc to typedoc.
* [ ] Convert to async/await

\* In our ES5-classes, we often did an `if (!(this instanceof Class)) { return new Class(); }` check, meaning you could technically call a class constructor without `new`. Just in case this is still being used, we should patch LocalAudioTrack, LocalVideoTrack, and LocalDataTrack to perform such a check before exporting (if possible—otherwise we have to bump major versions).


markandrus:
I haven't started the TypeScript half of this PR yet, but already I think there is enough good stuff—like ES6 classes—to merge some version of this PR ahead of time. There is one problem: twilio-video.js 1.x needs to continue loading/parsing in ES5 environments. This is true for both the

* dist/twilio-video.js and dist/twilio-video.min.js files we ship, and
* the NPM module whose entry point is identified by "main" in package.json.

Here is the solution I propose:

1. Use Babel to compile lib/ to es5/,
2. Browserify es5/ instead of lib/
3. Update "main" in package.json to point to es5/.

I've pushed a commit that implements this. One issue (and it's really the same issue we have with converting to TypeScript): we now require a build step, at least when pinning to the git repo.

package.json
```diff
@@ -107,13 +109,14 @@
     "test:framework:react": "npm-run-all test:framework:react:*",
     "test:framework": "npm-run-all test:framework:meteor test:framework:no-framework test:framework:react",
     "test": "npm-run-all test:unit test:integration",
+    "build:es5": "rimraf ./es5 && babel lib -d es5",
     "build:js": "node ./scripts/build.js ./src/twilio-video.js ./LICENSE.md ./dist/twilio-video.js",
     "build:min.js": "uglifyjs ./dist/twilio-video.js -o ./dist/twilio-video.min.js --comments \"/^! twilio-video.js/\" -b beautify=false,ascii_only=true",
-    "build": "npm-run-all clean lint docs cover test:integration build:js build:min.js test:umd",
+    "build": "npm-run-all clean lint docs cover test:integration build:es5 build:js build:min.js test:umd",
     "build:travis": "npm-run-all clean lint docs cover test:integration:travis build:js build:min.js test:umd test:framework",
```
markandrus: Add `build:es5` here.

PR #264, opened 2018-02-09
Convert to es6

markandrus:
Separating this out from #263 like we mentioned. Mind taking a look? Same comments from that PR apply. Best viewed with `?w=1`, i.e. https://github.com/twilio/twilio-video.js/pull/264/files?w=1.


lib/connect.js
```diff
@@ -224,18 +222,18 @@ function connect(token, options) {
...
     createRoom.bind(null, options));
 
-  cancelableRoomPromise.then(function(room) {
+  cancelableRoomPromise.then(room => {
     log.info('Connected to Room:', room.toString());
     log.info('Room name:', room.name);
     log.debug('Room:', room);
     return room;
-  }, function(error) {
+  }, error => {
```
manjeshbhargav: @markandrus Isn't it safer to enforce braces around the arguments even if there is only one?

lib/data/sender.js
```diff
@@ -1,70 +1,67 @@
...
+class DataTrackSender extends DataTrackTransceiver {
+  /**
+   * Construct a {@link DataTrackSender}.
+   * @param {?number} maxPacketLifeTime
+   * @param {?number} maxRetransmits
+   * @param {boolean} ordered
+   */
+  constructor(maxPacketLifeTime, maxRetransmtis, ordered) {
+    const id = makeUUID();
+    super(id, maxPacketLifeTime, maxRetransmtis, ordered);
```
manjeshbhargav: @markandrus Usually its safe to call `super()` in the first line of the constructor. If we're not using `id` anywhere else, can we do:
```javascript
constructor(maxPacketLifeTime, maxRetransmtis, ordered) {
  super(makeUUID, ...);
  ...
}
```
?

ryan-rowland: Looks like a typo here too, `maxRetransmtis`

lib/data/sender.js
```diff
@@ -1,70 +1,67 @@
...
+    Object.defineProperties(this, {
```
manjeshbhargav: @markandrus Can we declare this as a `private readonly` class member if es6 allows it? Then I don't think we need to do this manually.

lib/media/track/localvideotrack.js
```diff
@@ -1,71 +1,78 @@
...
+  }
 
-LocalVideoTrack.prototype.toString = function toString() {
-  return '[LocalVideoTrack #' + this._instanceId + ': ' + this.id + ']';
-};
+  /**
+   * @private
+   */
+  _end() {
+    return LocalMediaVideoTrack.prototype._end.apply(this, arguments);
```
manjeshbhargav: @markandrus 
```javascript
_end() {
  return super._end.apply(...);
}
```
?

lib/media/track/localvideotrack.js
```diff
@@ -1,71 +1,78 @@
...
- * @fires VideoTrack#disabled
- */
-LocalVideoTrack.prototype.disable = LocalMediaTrack.prototype.disable;
+  /**
+   * Disable the {@link LocalVideoTrack}. This is effectively "pause".
+   * @returns {this}
+   * @fires VideoTrack#disabled
+   */
+  disable() {
+    return LocalMediaVideoTrack.prototype.disable.apply(this, arguments);
```
manjeshbhargav: @markandrus Use `super.disable.apply(...)`?

lib/media/track/localvideotrack.js
```diff
@@ -1,71 +1,78 @@
...
+   * Enable or disable the {@link LocalVideoTrack}. This is effectively "unpause"
+   * or "pause".
+   * @param {boolean} [enabled] - Specify false to pause the
+   *   {@link LocalVideoTrack}
+   * @returns {this}
+   * @fires VideoTrack#disabled
+   * @fires VideoTrack#enabled
+   */
+  enable() {
+    return LocalMediaVideoTrack.prototype.enable.apply(this, arguments);
```
manjeshbhargav: @markandrus `super.enable.apply`?

lib/media/track/localvideotrack.js
```diff
@@ -1,71 +1,78 @@
...
- */
-LocalVideoTrack.prototype.stop = LocalMediaTrack.prototype.stop;
+  /**
+   * Calls stop on the underlying MediaStreamTrack. If you choose to stop a
+   * {@link LocalVideoTrack}, you should unpublish it after stopping.
+   * @returns {this}
+   * @fires LocalVideoTrack#stopped
+   */
+  stop() {
+    return LocalMediaVideoTrack.prototype.stop.apply(this, arguments);
```
manjeshbhargav: @markandrus `super.stop.apply`?

lib/signaling/participant.js
```diff
@@ -37,94 +34,97 @@ var states = {
...
+        writable: true,
+        value: null
+      },
+      identity: {
+        enumerable: true,
+        get() {
+          return this._identity;
+        }
+      },
+      sid: {
```
manjeshbhargav: @markandrus Use a getter method?

lib/connect.js
markandrus: What do you mean by safer? I don't know of any issue omitting braces. I can see that, if a function where to ever gain an additional argument, it's a slightly larger diff (because the parens have to be added), but otherwise I think it is OK.

lib/data/sender.js
markandrus: Nice catch on typo—will fix. If `id` is unused, I'll remove the binding and pass `makeUUID()` directly to `super`.

lib/data/sender.js
markandrus: `private` and `readonly` are TypeScript features. We can introduce these when we move to TypeScript 👍

lib/media/track/localvideotrack.js
markandrus: Wow! I didn't know we could do that :-)

lib/signaling/participant.js
markandrus: That's a good suggestion—maybe we can wait to do that in another PR, though? There are some other changes we could make that I haven't included here (for example, `async`/`await`).

manjeshbhargav:
+1

PR #267, opened 2018-02-15
Move LocalParticipantV2 revision management out of RoomV2

markandrus:
@manjeshbhargav previously, RoomV2 was managing LocalParticipantV2's `revision`, which we identified as being weird. Now, LocalParticipantV2 manages it in response to

* Tracks being added,
* Tracks being removed, and
* Added Tracks changing (except for SID assignment).

Basically, I just moved the logic from lib/signaling/v2/room.js into lib/signaling/v2/localparticipant.js.

I still need to find a way to test the `revision` re-use bug when enabling/disabling a Track early when connecting to a Room.

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -165,7 +165,6 @@ class RoomV2 extends RoomSignaling {
    * @private
    */
   _publishNewLocalParticipantState() {
-    this.localParticipant.incrementRevision();
     this._transport.publish(this._getState());
   }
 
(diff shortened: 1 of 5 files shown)


lib/signaling/v2/room.js
```diff
@@ -274,66 +273,24 @@ class RoomV2 extends RoomSignaling {
...
-    addListener(track);
-    updateLocalParticipantStateAndRenegotiate();
-  }
+    roomV2._peerConnectionManager.setTrackSenders(trackSenders);
+  });
 
-  function trackRemoved(track) {
-    removeListener(track);
-    updateLocalParticipantStateAndRenegotiate();
+  function localParticipantUpdated() {
```
markandrus: Hm, I guess this could be once-per-tick, too?

manjeshbhargav:
+1

PR #273, opened 2018-03-05
Raise SignalingConnectionDisconnectedError when WebSocket is disconnected

markandrus:
@manjeshbhargav

### DIFF
--- a/lib/signaling/v2/transport.js
+++ b/lib/signaling/v2/transport.js
@@ -8,11 +8,14 @@ const SIP = require('../../sip');
 const DefaultSIPJSMediaHandler = require('./sipjsmediahandler');
 const StateMachine = require('../../statemachine');
 const util = require('../../util');
-const TwilioErrors = require('../../util/twilio-video-errors');
-const createTwilioError = TwilioErrors.createTwilioError;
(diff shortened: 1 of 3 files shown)


lib/signaling/v2/transport.js
```diff
@@ -8,11 +8,14 @@ const SIP = require('../../sip');
 const DefaultSIPJSMediaHandler = require('./sipjsmediahandler');
 const StateMachine = require('../../statemachine');
 const util = require('../../util');
-const TwilioErrors = require('../../util/twilio-video-errors');
-const createTwilioError = TwilioErrors.createTwilioError;
-const SignalingConnectionError = TwilioErrors.SignalingConnectionError;
-const SignalingConnectionTimeoutError = TwilioErrors.SignalingConnectionTimeoutError;
-const SignalingIncomingMessageInvalidError = TwilioErrors.SignalingIncomingMessageInvalidError;
+
+const {
```
manjeshbhargav: @markandrus Are we doing object destructuring here because we're babelifying it later?

markandrus: Yes, we can use all the new stuff now (except `async`/`await`—that requires regenerator, which I'm not sure we want to pull in).

PR #274, opened 2018-03-05
Raise SignalingConnectionTimeoutError if keep-alives timeout

markandrus:
@manjeshbhargav


manjeshbhargav:
@markandrus Same question about `es2015` usage as in the other PR. Otherwise +1.

PR #277, opened 2018-03-19
JSDK-1810: Add support for ICE restart

markandrus:
If an ICE failure occurs (i.e., an individual RTCPeerConnection's `iceConnectionState` transitions to "failed"), we now perform an ICE restart. Subsequent offers from the Client will block until the ICE restart has completed (not unlike what we do in the initial round of negotiation with the `_needsInitialAnswer` boolean); however, glare will still force the Client to rollback, apply the Server's answer, and then restart ICE (with yet another unique set of ICE credentials).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ Bug Fixes
   timely manner. Now, WebSocket timeouts can be identified in around 30 seconds.
   If a WebSocket timeout occurs, for example, due to a WebSocket disconnect,
   Room will emit a "disconnected" event with SignalingConnectionTimeoutError.
+- If an ICE failure occurs, for example due to disconnecting a VPN, but the
+  but the signaling connection remains online, twilio-video.js will attempt an
+  ICE restart to repair the media connection. (JSDK-1810)
 
 1.8.0 (February 9, 2018)
 ========================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ Bug Fixes
   If a WebSocket timeout occurs, for example, due to a WebSocket disconnect,
   Room will emit a "disconnected" event with SignalingConnectionTimeoutError.
 - If an ICE failure occurs, for example due to disconnecting a VPN, but the
-  but the signaling connection remains online, twilio-video.js will attempt an
-  ICE restart to repair the media connection. (JSDK-1810)
+  signaling connection remains online, twilio-video.js will attempt an ICE
+  restart to repair the media connection. (JSDK-1810)
 
 1.8.0 (February 9, 2018)
 ========================
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -19,6 +19,9 @@ const DataTrackReceiver = require('../../data/receiver');
 const MediaTrackReceiver = require('../../media/track/receiver');
 const StateMachine = require('../../statemachine');
 const StatsReport = require('../../stats/statsreport');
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ Bug Fixes
   timely manner. Now, WebSocket timeouts can be identified in around 30 seconds.
   If a WebSocket timeout occurs, for example, due to a WebSocket disconnect,
   Room will emit a "disconnected" event with SignalingConnectionTimeoutError.
+- If an ICE failure occurs, for example due to disconnecting a VPN, but the
+  but the signaling connection remains online, twilio-video.js will attempt an
```
manjeshbhargav: @markandrus Extra `but the` at the beginning of this line?

Also this can be rewritten as: `... connection remains online, an ICE restart will be attempted to repair the media connection.`?

lib/signaling/v2/peerconnection.js
```diff
@@ -356,6 +380,31 @@ class PeerConnectionV2 extends StateMachine {
...
+    // Case 1: Transition to "failed".
+    if (this._lastIceConnectionState !== 'failed' && iceConnectionState === 'failed' && !this._shouldRestartIce && !this._isRestartingIce) {
+      this._log.warn('ICE failed; attempting to restart ICE');
+      this._shouldRestartIce = true;
+      this.offer();
+    }
+
+    // Case 2: Transition from "failed".
+    else if (this._lastIceConnectionState === 'failed' && (iceConnectionState === 'connected' || iceConnectionState === 'completed')) {
+      this._log.info('ICE reconnected');
```
manjeshbhargav: @markandrus Should `._shouldRestartIce` again be set to `false` here?

CHANGELOG.md
markandrus: Oh, nice catch. I'll try your re-wording, too.

lib/signaling/v2/peerconnection.js
markandrus: I'm not sure. The way I've implemented here, `iceConnectionState` transitioning to "failed" kicks off an ICE restart, and the ICE restart will be completed* regardless of whether `iceConnectionState` ever transitions out of "failed". I suppose if we were forced to do a rollback and, during that time, `iceConnectionState` transitioned out of "failed", we might be able to skip restarting ICE again... but I think this optimization would complicate the implementation. So I'd prefer not to set it here.

\* An ICE restart is completed once an offer produced by `createOffer({ iceRestart: true })` is answered, taking into account glare may force us to do a rollback before this happens.

CHANGELOG.md
markandrus: I fixed the duplicate "but the", but, after reviewing, I kept the second half as-is, since I think it is more active voice (whereas "will be attempted" is more passive voice).

ramapal:
@markandrus Overall looks good to me. One thing I was wondering is whether you enforce a maximum number of retries before disconnecting the participant. In CPP, we limit it to 5 retries.

markandrus:
Thanks, @ramapal — we will not limit the total number of retries, but we will only retry on a new transition into the "failed" state (i.e., we won't retry an ICE restart unless a previous ICE restart succeeded). I've recommended mobile SDK's change their disconnect policy, and that instead we let the media server disconnect Participants with a failed ICE connection after some timeout.

amitshivnani:
AFAIK, we are not too far away from this disconnect policy in Core, we try 5 times to restart, if in "failed state", and then let media server disconnect Participants with Failed ICE connection.

PR #285, opened 2018-04-04
JSDK-1855: Add "reconnecting" state to Room

markandrus:
~~_Not ready to merge._~~

This PR will eventually expose a "reconnecting" `state` on Room, as well as "reconnecting" and "reconnected" events. While we don't yet support full Network Handoff (we cannot currently recover our Signaling Connection); we are working to support handoff of the Media Connection (useful, for example, if you got routed through a VPN that eventually stops routing your traffic). This feature will eventually compliment full Network Handoff support.


lib/signaling/room.js
```diff
@@ -49,6 +66,10 @@ class RoomSignaling extends StateMachine {
     const RecordingSignaling = options.RecordingSignaling;
 
     Object.defineProperties(this, {
+      _mediaConnectionIsReconnecting: {
```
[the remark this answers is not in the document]
markandrus: The problem is that the `mediaConnectionState` itself won't indicate whether the Media Connection is reconnecting or not. Media Connection states "new" and "checking" may be transitioned through during either the "connecting" or "reconnecting" states. So we need some extra bit of data to indicate if we're reconnecting or not.

lib/room.js
```diff
@@ -360,11 +387,19 @@ function handleSignalingEvents(room, signaling) {
...
     log.info('Transitioned to state:', state);
-    room.emit(state, room, error);
+    if (state === 'disconnected') {
+      room.participants.forEach(participant => {
+        participant._unsubscribeTracks();
+      });
+      room.emit(state, room, error);
+      signaling.removeListener('stateChanged', stateChanged);
+      return;
+    } else if (state === 'reconnecting') {
```
manjeshbhargav: @markandrus Use `if` instead of `else if` since the previous block returns?

markandrus: I prefer this

```js
if (foo) {
  return foo;
} else if (bar) {
  return bar;
}
return baz;
```

to this

```js
if (foo) {
  return foo;
}
if (bar) {
  return bar;
}
return baz;
```

I guess because the `else if` indicates the clauses are related rather than distinct and because it is a little more compact. I guess a third option could be

```js
if (foo) {
  return foo;
} if (bar) {
  return bar;
}
return baz;
```

but that's weird, I think.

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -126,11 +144,15 @@ class PeerConnectionManager extends QueueingEventEmitter {
           peerConnection.removeListener('stateChanged', stateChanged);
           self._peerConnections.delete(peerConnection.id);
           self._closedPeerConnectionIds.add(peerConnection.id);
+          updateIceConnectionState(self);
         }
       });
+      peerConnection.on('iceConnectionStateChanged', () => updateIceConnectionState(this));
```
manjeshbhargav: @markandrus I think we should make `updateIceConnectionState()` a member of `PeerConnectionManager`. WDYT?

markandrus: I don't really like to add members to the prototype if it can be avoided. Perhaps TypeScript will force this, though. But I'd rather defer doing this for now, since this is a pattern we have elsewhere: functions for mutating private members scoped to the module instead of the prototype.

markandrus:
I'm not sure why the Meteor test is failing all of a sudden. I'm going to check it on master. If it fails there, too, I will disable in this PR.

lib/signaling/room.js
```diff
@@ -141,4 +165,42 @@ class RoomSignaling extends StateMachine {
...
+
+/**
+ * @event RoomSignaling#event:signalingConnectionStateChanged
+ */
+
+/**
+ * Maybe update the {@link RoomSignaling} state.
+ * @param {RoomSignaling} roomSignaling
+ */
+function maybeUpdateState(roomSignaling) {
```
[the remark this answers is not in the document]
ryan-rowland: Hmm, I wonder if the special cases could be tracked via flags as well? I imagine the logic would be similar, just checked on `get` rather than computed on state change.

lib/room.js
ryan-rowland: This seems like a classic case for `switch`, no?

lib/signaling/v2/peerconnectionmanager.js
ryan-rowland: Should we be removing this listener at some point? Destruction? If so, we'll need a bound version of this attached to the instance I believe.

lib/room.js
markandrus: We could do a `switch`:

```js
switch (state) {
  case 'disconnected':
    room.participants.forEach(participant => {
      participant._unsubscribeTracks();
    });
    room.emit(state, room, error);
    signaling.removeListener('stateChanged', stateChanged);
    break;
  case 'reconnecting':
    room.emit('reconnecting', new MediaConnectionError());
    break;
  default:
    room.emit('reconnected');
}
```

lib/signaling/room.js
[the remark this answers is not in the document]
markandrus: Sorry, I don't understand or can't visualize how that would work. Perhaps you could do some mutation in the `get`, but I think it's generally best to avoid that pattern? Also, you'd have to kick off whatever computation the `get` encapsulates in response to the underlying state changes. I'm not sure it's a good fit.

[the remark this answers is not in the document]
markandrus: Also, when you say " I wonder if the special cases could be tracked via flags as well?" — how would that be different than the current use of the flag `_mediaConnectionIsReconnecting`?

lib/signaling/v2/peerconnectionmanager.js
markandrus: We don't remove the others, so I don't think we need to remove this one. Or rather, if we decide we do need to remove these, let's tackle in a separate PR.

lib/signaling/room.js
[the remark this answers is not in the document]
ryan-rowland: The `get` would not do any mutation, nor would it need to be called in response to state changes. The concept here is that both our media stream and signaling classes already have concrete state machines, and rather than implementing a third state machine for `Room`, we should treat `Room.state` as a view that translates the underlying state machines' states into a singular state that is relevant to the `Room`.

The getter should not mutate anything; it would check: media state, signaling state, † and †† and output a room state.

PR #287, opened 2018-04-17
JSDK-1437, JSDK-1513 TrackPublication with .isTrackEnabled property.

manjeshbhargav:
@markandrus


lib/media/track/localtrackpublication.js
```diff
@@ -1,42 +1,35 @@
...
     Object.defineProperties(this, {
-      _instanceId: {
-        value: nInstances++
-      },
-      _log: {
-        value: options.log || new Log('default', this, logLevels),
+      _onTrackEvent: {
+        value: () => {
+          this.isTrackEnabled = track.isEnabled;
+        }
```
markandrus: Can you move this out of `defineProperties` and do, e.g.,

```
class Foo {
  bar = () => {
    // ...
  }
}
```

There should be a Babel plugin to allow this.

lib/media/track/localtrackpublication.js
```diff
@@ -48,16 +41,12 @@ class LocalTrackPublication {
...
-        enumerable: true,
-        value: track.name
-      },
-      trackSid: {
-        enumerable: true,
-        value: sid
       }
     });
+
+    this.trackSid = trackSid;
```
markandrus: I think you want to write to `_trackSid` instead of `trackSid`?

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,83 @@
...
+      _log: {
+        value: options.log || new Log('default', this, logLevels)
+      },
+      _trackSid: {
+        value: null,
+        writable: true
+      },
+      trackName: {
+        value: trackName
+      }
```
markandrus: This can be written as a getter.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,83 @@
...
+  get isTrackEnabled() {
+    return this._isTrackEnabled;
+  }
+
+  set isTrackEnabled(isTrackEnabled) {
+    if (this._isTrackEnabled !== isTrackEnabled) {
+      this._isTrackEnabled = isTrackEnabled;
+      this.emit(this._isTrackEnabled ? 'trackEnabled' : 'trackDisabled');
+    }
+  }
```
markandrus: Can we avoid the side-effecting setter? Or at least move the side-effects to a setter for `_isTrackEnabled` instead of `isTrackEnabled`? IMO, `isTrackEnabled` should not have a setter.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,83 @@
...
+const { DEFAULT_LOG_LEVEL } = require('../../util/constants');
+const Log = require('../../util/log');
+let nInstances = 0;
+
+/**
+ * A {@link TrackPublication} represents a {@link Track} that
+ * has been published to a {@link Room}.
+ * @property {boolean} isTrackEnabled - whether the published {@link Track} is enabled
+ * @property {string} trackName - the {@link Track}'s name
+ * @property {Track.SID} trackSid - SID assigned to the published {@link Track}
```
markandrus: I think you can document the properties on their getters.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,83 @@
...
+ * @emits TrackPublication#trackDisabled
+ * @emits TrackPublication#trackEnabled
+ */
+class TrackPublication extends EventEmitter {
+  constructor(isTrackEnabled, trackName, options) {
```
markandrus: Why isn't Track SID a parameter to the constructor? It's required IIRC.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  get trackSid() {
+    return this._trackSid;
+  }
+
+  set trackSid(trackSid) {
+    this._trackSid = trackSid;
+  }
```
markandrus: Please remove. `_trackSid` should be set instead. But actually, isn't Track SID constant anyway? Why not accept it in the constructor?

lib/media/track/trackpublication.js
manjeshbhargav: @markandrus What is the use of doing that. The name doesn't change, so initial assignment to the property should be fine.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,91 @@
...
+      _isTrackEnabled: {
+        get() {
+          return _isTrackEnabled;
+        },
+        set(isTrackEnabled) {
+          if (_isTrackEnabled !== isTrackEnabled) {
+            _isTrackEnabled = isTrackEnabled;
+            this.emit(_isTrackEnabled ? 'trackEnabled' : 'trackDisabled');
+          }
+        }
```
markandrus: I just had a realization. Don't we want the property that

```js
publication.track.on('enabled', () => {
  assert.equal(publication.isTrackEnabled, track.isEnabled);
});

publication.track.on('disabled', () => {
  assert.equal(publication.isTrackEnabled, track.isEnabled);
})
```

Will this design help or hinder fulfilling this property?

manjeshbhargav: @markandrus If we want to do this, then the `Track` has to have a set of `.publications` which are publishing the it, and then it has to update each publication's `.isTrackEnabled` whenever its .isEnabled changes.

lib/media/track/localtrackpublication.js
```diff
@@ -48,38 +39,47 @@ class LocalTrackPublication {
...
- * @typedef {object} LocalTrackPublicationOptions
- * @property {LogLevel|LogLevels} logLevel - Log level for 'media' modules
+ * The published {@link Track} was disabled.
+ * @event TrackPublication#trackDisabled
+ */
+
+/**
+ * The published {@link Track} was enabled.
+ * @event TrackPublication#trackEnabled
  */
```
markandrus: These declarations should eventually move to trackpublication.js. We can do that now or later.

lib/media/track/trackpublication.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+/**
+ * {@link TrackPublication} options
+ * @typedef {object} TrackPublicationOptions
+ * @property {LogLevel|LogLevels} logLevel - Log level for 'media' modules
+ */
+
+module.exports = TrackPublication;
```
markandrus: Can you add this to the JSDoc generation script?

markandrus:
just a couple of comments. otherwise +1

manjeshbhargav:
Re-opening the PR with a `twilio` remote branch to enable Travis to run correctly: https://github.com/twilio/twilio-video.js/pull/288

PR #299, opened 2018-05-11
JSDK-1912: Implement MediaStreamTrack cloning

markandrus:
_This is a breaking API change, so we will not merge it until we have a plan for releasing 2.0.0._

__Why make this change?__

Whenever a LocalAudioTrack or LocalVideoTrack is published to a Room, we're going to clone the underlying MediaStreamTrack. This will ensure we're always signaling a unique MSID in SDPs and a unique Track ID in our signaling messages. Our signaling protocol doesn't work well if the same LocalAudioTrack or LocalVideoTrack is 1. published, unpublished, and then published, and 2. these messages cross on the wire. This should also let us work around issues like WebKit [Bug 174327](https://bugs.webkit.org/show_bug.cgi?id=174327).

__Why is it a breaking API change?__

Although we consider Track IDs "deprecated" in favor of Track Names or Track SIDs ([since 1.4.0](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md#140-october-2-2017)), Track IDs are still a part of our RemoteTrack and RemoteTrackPublication APIs. So it'd be natural to assume if you publish LocalTrack with ID "1", SID "MTxyz", and name "webcam" that you get a RemoteTrack on the other side with ID "1", SID "MTxyz" and name "webcam". However, since we'll clone the underlying MediaStreamTracks before publishing them, the ID will not equal "1"—it'll be something different. Therefore, if you relied on this property, the behavior has changed.

__What will happen in 2.0.0?__

In 2.0.0, we'll remove RemoteTrack and RemoteTrackPublication IDs entirely (LocalTrack IDs will remain). Then, the cloning behavior introduced by this PR will be unobservable. We'll also have freed ourselves up for the media server to rewrite MediaStreamTrack IDs as necessary, too. We'll add deprecation warnings for anything we plan on removing before we do this.

__Todo__

- [x] Update `getStats` — it will need to be aware of the new IDs and we'll need to surface the correct LocalTrack IDs.

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -131,13 +131,13 @@ class LocalParticipant extends Participant {
     }
 
     function localTrackDisabled(localTrack) {
-      const trackSignaling = signaling.tracks.get(localTrack.id);
+      const trackSignaling = signaling.getPublication(localTrack._trackSender);
       trackSignaling.disable();
       log.debug(`Disabled the ${util.trackClass(localTrack, true)}:`, localTrack.id);
     }
 
     function localTrackEnabled(localTrack) {
-      const trackSignaling = signaling.tracks.get(localTrack.id);
+      const trackSignaling = signaling.getPublication(localTrack._trackSender);
       trackSignaling.enable();
       log.debug(`Enabled the ${util.trackClass(localTrack, true)}:`, localTrack.id);
     }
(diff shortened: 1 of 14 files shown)


ryan-rowland:
Looks good to me 👍

markandrus:
Thanks for the reviews. I'm now closing, and I will re-open against the 2.x branch.

PR #306, opened 2018-05-16
JSDK-1930: Start publishing SDK version in "connected" message

markandrus:


lib/signaling/v2/transport.js
```diff
@@ -82,7 +82,10 @@ class Transport extends StateMachine {
     options = Object.assign({
       InsightsPublisher,
       NullInsightsPublisher,
-      SIPJSMediaHandler: DefaultSIPJSMediaHandler
+      SIPJSMediaHandler: DefaultSIPJSMediaHandler,
+      userAgent: typeof navigator !== 'undefined' && navigator.userAgent
```
ryan-rowland: Should this call `util.getUserAgent()`?

markandrus: Oops, I thought I fixed that--good catch

PR #308, opened 2018-05-22
JSDK-1997 Set RemoteTrack's .sid before calling Participant#addTrack().

manjeshbhargav:
@markandrus


markandrus:
@manjeshbhargav change looks simple enough 👍 Can you land a unit test demonstrating JSDK-1997, though?

PR #311, opened 2018-05-30
JSDK-2004 Implementing TwilioConnection class + unit tests.

manjeshbhargav:
@markandrus This PR contains the implementation of the `TwilioConnection` class along with
unit tests.

TODO:
=====

[✓] Create echo server using `twilio-connections` java library.
[✓] Run integration tests against the server and update the PR if necessary.


lib/twilioconnection.js
```diff
@@ -0,0 +1,191 @@
...
+
+  _handleHeartbeatTimeout() {
+    if (this.state !== 'open') {
+      return;
+    }
+    this._consecutiveHeartbeatsMissed++;
+    const { maxConsecutiveMissedHeartbeats } = this._options;
+
+    if (this._consecutiveHeartbeatsMissed < maxConsecutiveMissedHeartbeats) {
+      setTimeout(() => this._handleHeartbeatTimeout(), this._heartbeatTimeout);
```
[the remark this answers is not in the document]
manjeshbhargav: @markandrus Since we're using `setTimeout` instead of `setInerval`, we just stop calling `setTimeout` once we reach `maxConsecutiveMissedHeartbeats`.

test/unit/spec/twilioconnection.js
```diff
@@ -0,0 +1,288 @@
...
+      twilioConnection = new TwilioConnection('foo', {
+        WebSocket: FakeWebSocket
+      });
+    });
+
+    it('should return an instance of TwilioConnection', () => {
+      assert(twilioConnection instanceof TwilioConnection);
+    });
+
+    it('should set the .state to "connecting"', () => {
```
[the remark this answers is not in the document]
manjeshbhargav: @markandrus I was thinking about that too. I'll remove the "new" state.

lib/twilioconnection.js
```diff
@@ -0,0 +1,341 @@
...
+        default:
+          this.emit('error', new Error(`Unknown message type: ${message.type}`));
+          break;
+      }
+    });
+
+    ws.addEventListener('open', () => {
+      this._sendHello();
+      const { welcomeTimeout } = this._options;
+      setTimeout(() => this._handleWelcomeTimeout(), welcomeTimeout);
```
markandrus: I think we should clean up this timeout if the TwilioConnection is closed or a welcome is received, as open handles like the timeouts created by `setTimeout` will keep the Node process open

markandrus:
@manjeshbhargav nice work 👍 just one little comment i left. do you think we should create a `remove-sip` or similarly-named branch against `2.x` and then change this PR's base branch to `remove-sip`? i ask, because we have some more code to land (and delete) before this can be useful. actually, i guess so long as nothing `require`s the new code it will be excluded from bundles. so maybe it is OK to merge. i leave the strategy up to you

manjeshbhargav:
@markandrus What we can do is, keep this PR open, and then start using the current `twilio-connections` branch as the base to open further PRs against. At the end, we can merge this branch with `2.x`. WDYT?

lib/util/timeout.js
```diff
@@ -0,0 +1,58 @@
...
+   */
+  clear() {
+    clearTimeout(this._timeout);
+    this._timeout = null;
+  }
+
+  /**
+   * Reset the {@link Timeout}.
+   */
+  reset() {
```
markandrus: Can you add `@returns` to these?

lib/util/timeout.js
```diff
@@ -0,0 +1,58 @@
...
+ * A {@link Timeout} represents a resettable and clearable timeout.
+ */
+class Timeout {
+  /**
+   * Construct a {@link Timeout}.
+   * @param {function} fn - Function to call
+   * @param {number} delay - Delay in milliseconds
+   * @param {Array<*>} args - Optional arguments to be passed to the function
+   */
+  constructor(fn, delay, ...args) {
```
markandrus: I recommend nixing the `args` functionality—It can be handled by passing an appropriate `fn`, which then simplifies Timeout.

lib/util/timeout.js
```diff
@@ -0,0 +1,58 @@
...
+        value: fn
+      },
+      _timeout: {
+        value: null,
+        writable: true
+      }
+    });
+
+    this._start();
+  }
```
markandrus: Just thinking: it might be nice to have a getter that says whether or not the Timeout is active/live/stopped/etc. something like

```
get isFoo() {
  return this._timeout !== null;
}
```

manjeshbhargav:
Re-opening against `master`.

PR #313, opened 2018-05-30
Network Quality API for Group Rooms

markandrus:
This PR contains changes to support the upcoming Network Quality API for Group Rooms. The necessary Media Server changes are currently in our dev environment, so if you test this PR out now, you'll see that `networkQualityLevel` is always `null`. Once the server-side changes are deployed to prod, the way this will work is as follows:

1. twilio-video.js requests to negotiate an RTCDataChannel with the Media Server to exchange Network Quality-related messages.
2. If accepted, the Media Server opens an RTCDataChannel to twilio-video.js.
3. twilio-video.js gathers statistics necessary to compute Network Quality and sends them to the Media Server over the RTCDataChannel.
4. The Media Server responds with a Network Quality Level from 1–5 for the LocalParticipant.
5. Eventually, the Media Server will also respond with Network Quality Levels for RemoteParticipants.

Example

```js
// networkQualityLevel is initially null.
assert.strictEqual(room.localParticipant.networkQualityLevel, null);

room.localParticipant.on('networkQualityLevelChanged', networkQualityLevel => {
  // Once the first network quality level is computed, it's always 0–5.
  assert(networkQualityLevel >= 0 && networkQualityLevel <= 5);
});
```

TODO

- [x] Update CHANGELOG.md
- [x] The LocalParticipant's `networkQualityLevel` should be set to zero during media reconnections


ryan-rowland:
Very cool feature 👍

lib/signaling/v2/networkqualitymonitor.js
```diff
@@ -0,0 +1,119 @@
...
+    clearTimeout(this._timeout);
+    this._timeout = null;
+  }
+}
+
+/**
+ * @param {NetworkQualityMonitor}
+ * @returns {Promise<NetworkQualityInputs>}
+ */
+function  next(monitor) {
```
manjeshbhargav: @markandrus Nitpick: extra space between `function` and `next`.

lib/stats/icereport.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+/**
+ * @property {number} [availableSend] - bps (undefined in Firefox)
+ * @property {number} recv - bps
+ * @property {number} [rtt] - s (undefined in Firefox)
+ * @property {number} send - bps
+ */
+class IceReport {
```
manjeshbhargav: @markandrus Can't we re-use our Active Ice Candidate Pair stats that we already collect? Or are these stats different?

lib/signaling/v2/networkqualitymonitor.js
markandrus: Oops, thanks

lib/stats/icereport.js
markandrus: IceReport is computing an additional piece of information beyond the Active ICE Candidate Pair, `bps`; however, we _should_ identify a way to share the stats we're collecting between Insights and Network Quality API. Ideally we'd just share an RTCStatsReport between both, but Firefox complicates this, since there's no way to tie RTCStats instances to particular RTCRtpSenders/RTCRtpReceivers/MediaStreamTracks unless you call `getStats` on those objects directly (relates in part to Firefox [Bug 1463430](https://bugzilla.mozilla.org/show_bug.cgi?id=1463430)).

manjeshbhargav:
@markandrus Overall approach looks good. I'll leave it to your judgement to merge. 👍

lib/signaling/v2/networkqualitymonitor.js
```diff
@@ -0,0 +1,119 @@
...
+        if (this._timeout !== timeout) {
+          return;
+        }
+        if (reports.length) {
+          const [report] = reports;
+          this._signaling.put(report);
+        }
+        this.start();
+      });
+    }, 200);
```
markandrus: This may be too often. Revisit.

PR #315, opened 2018-06-04
Add Dominant Speaker detection for Group Rooms

markandrus:
This PR contains changes to support Dominant Speaker detection in Group Rooms. Like #313, the necessary Media Server changes are current in our dev environment, so if you test this PR out now, you'll see that `dominantSpeaker` is always `null`. Once the server-side changes are deployed to prod (and once we enable this feature by default), `dominantSpeaker` will be set to the RemoteParticipant that represents the Dominant Speaker in the Room, if any. The "dominantSpeakerChanged" event will signal changes to this property.

**Example**

```js
// dominantSpeaker is initially null
assert.strictEqual(room.dominantSpeaker, null);

// dominantSpeaker will never be set to the LocalParticipant
assert.notEqual(room.dominantSpeaker, room.localParticipant);

// If set, dominantSpeaker is always a RemoteParticipant
if (room.dominantSpeaker) {
  assert.equal(room.dominantSpeaker, room.participants.get(room.dominantSpeaker.sid));
}

room.on('dominantSpeakerChanged', dominantSpeaker => {
  if (dominantSpeaker) {
    // There is a new Dominant Speaker; update the UI to display their video prominently
  } else {
    // There is no longer a Dominant Speaker
  }
});
```

**TODO**

- [x] Need to ensure `activeSpeaker` is null-ed out if the Active Speaker disconnects
- [x] Update CHANGELOG.md
- [x] Consider renaming this API to "Dominant Speaker", since there may be many Active Speakers but only one "Dominant Speaker" (`dominantSpeaker` & "dominantSpeakerChanged")

### DIFF
--- a/lib/room.js
+++ b/lib/room.js
@@ -14,6 +14,8 @@ let nInstances = 0;
  * <br><br>
  * You can connect to a {@link Room} by calling {@link connect}.
  * @extends EventEmitter
+ * @property {?RemoteParticipant} activeSpeaker - The Active Speaker in the
+ *   {@link Room}, if any
  * @property {boolean} isRecording - Whether or not the {@link Room} is being
  *   recorded
  * @property {LocalParticipant} localParticipant - Your {@link LocalParticipant}
@@ -71,6 +73,12 @@ class Room extends EventEmitter {
       },
       _signaling: {
         value: signaling
+      },
+      activeSpeaker: {
+        enumerable: true,
+        get() {
+          return this.participants.get(signaling.activeSpeakerSid) || null;
+        }
       },
       isRecording: {
         enumerable: true,
@@ -141,6 +149,15 @@ class Room extends EventEmitter {
  * that uniquely identifies a {@link Room}.
  * @type string
  * @typedef Room.SID
+ */
+
+/**
+ * The Active Speaker in the {@link Room} changed. Either the Active Speaker is
+ * a new {@link RemoteParticipant} or the Active Speaker has been reset and is
+ * now null.
+ * @param {?RemoteParticipant} activeSpeaker - The Active Speaker in the
+ *   {@link Room}, if any
+ * @event Room#activeSpeakerChanged
  */
 
 /**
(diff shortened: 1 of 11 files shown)


lib/signaling/v2/activespeakersignaling.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+const { EventEmitter } = require('events');
+
+/**
+ * @property {?Track.SID} loudestParticipantSid
+ * @emits ActiveSpeakerSignaling#updated
+ */
+class ActiveSpeakerSignaling extends EventEmitter {
```
markandrus: The more I write classes like these—EventEmitters with a single event, e.g. "updated"—the more I'm feeling like we should go ahead and adopt Observable. The [zen-observable](https://github.com/zenparsing/zen-observable) is lightweight and pretty solid.

lib/signaling/room.js
```diff
@@ -73,6 +74,11 @@ class RoomSignaling extends StateMachine {
       _options: {
         value: options
       },
+      activeSpeaker: {
```
manjeshbhargav: @markandrus Nitpick: `activeSpeakerParticipantSid` or `activeSpeakerSid`?

markandrus: Can do 👍

lib/room.js
```diff
@@ -427,6 +448,8 @@ function handleSignalingEvents(room, signaling) {
     + 'ParticipantSignalings that connect to the RoomSignaling');
   signaling.on('participantConnected', connectParticipant.bind(null, room));
 
+  signaling.on('dominantSpeakerChanged', () => room.emit('dominantSpeakerChanged', room.dominantSpeaker));
```
markandrus: Write a test case for the case where:

1. `dominantSpeaker` is `null`,
2. An invalid Dominant Speaker SID arrives, and
3. Therefore we emit "dominantSpeakerChanged", although `dominantSpeaker` is still `null`.

PR #318, opened 2018-06-05
JSDK-1931

manjeshbhargav:
@markandrus 

In Firefox and Safari (SFU Room) for some reason, the `RoomSignaling#mediaConnectionState` transitions to "closed" before TransportV2 transitions to "disconnected". This will result in `maybeUpdateState()` in `lib/signaling/room.js` to transition the RoomSignaling's `.state` to "disconnected" without passing the `TwilioError`, which it doesn't have access to anyway.

So, the solution for this is to let the transition to "disconnected" be managed by `handleTransportEvents()` in `lib/signaling/v2/room.js` and ignore it in `maybeUpdateState()`.

Please let me know if this solution might affect other transitions which I might have missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,13 @@ printNetworkQualityLevel(participant.networkQualityLevel);
 participant.on('networkQualityLevelChanged', printNetworkQualityLevel);
 ```
 
+Bug Fixes
+---------
+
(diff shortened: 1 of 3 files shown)


markandrus:
Nice work! Let's update the CHANGELOG.md, too

PR #320, opened 2018-06-12
JSDK-2038 Clarifying "trackPublished" behavior.

manjeshbhargav:
@markandrus 

I was not sure if I should update `RemoteTrackPublication` guide in `CHANGELOG.md`, since it is present under an earlier version.

Please let me know if this clarification is sufficient, and whether we should include something in `CHANGELOG.md`.

### DIFF
--- a/lib/remoteparticipant.js
+++ b/lib/remoteparticipant.js
@@ -171,10 +171,25 @@ class RemoteParticipant extends Participant {
  */
 
 /**
- * A {@link RemoteTrack} was published by the {@link RemoteParticipant}.
+ * A {@link RemoteTrack} was published by the {@link RemoteParticipant} after
+ * connecting to the {@link Room}. This event is not emitted for
+ * {@link RemoteTrack}s that were published while the {@link RemoteParticipant}
+ * was connecting to the {@link Room}.
  * @event RemoteParticipant#trackPublished
  * @param {RemoteTrackPublication} publication - The {@link RemoteTrackPublication}
  *   which represents the published {@link RemoteTrack}
+ * @example
+ * function trackPublished(publication) {
+ *   console.log(`Track ${publication.trackSid} was published`);
+ * }
+ *
+ * room.on('participantConnected', participant => {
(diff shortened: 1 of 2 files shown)


lib/remoteparticipant.js
```diff
@@ -171,7 +171,10 @@ class RemoteParticipant extends Participant {
  */
 
 /**
- * A {@link RemoteTrack} was published by the {@link RemoteParticipant}.
+ * A {@link RemoteTrack} was published by the {@link RemoteParticipant} after
+ * connecting to the {@link Room}. This event is not emitted for
+ * {@link RemoteTrack}s that were published while the {@link RemoteParticipant}
+ * was connecting to the {@link Room}.
```
markandrus: I think it's worth including an `@example`. For example,

```jsdoc
@example
// Handle already published Tracks
participant.trackPublications.forEach(handleTrackPublication);

// Handle asynchronously published Tracks
participant.on('trackPublished', handleTrackPublication);

function handleTrackPublication(publication) {
  console.log(`Track ${publication.trackSid} was published`);
}
```

lib/room.js
```diff
@@ -271,7 +271,10 @@ class Room extends EventEmitter {
  */
 
 /**
- * A {@link RemoteTrack} was published by a {@link RemoteParticipant} to the {@link Room}.
+ * A {@link RemoteTrack} was published by a {@link RemoteParticipant} after
+ * connecting to the {@link Room}. This event is not emitted for
+ * {@link RemoteTrack}s that were published while the {@link RemoteParticipant}
+ * was connecting to the {@link Room}.
```
markandrus: If we make the change I suggest above, I recommend we include a similar example here.

manjeshbhargav:
@markandrus I think LocalParticipant's "trackPublished" is pretty straight-forward. We can tackle the differences between "trackAdded/trackRemoved" and "trackPublished/trackUnpublished" in the 2.x migration guide.

markandrus:
+1

PR #324, opened 2018-06-14
JSDK-2006 Network Handoff

manjeshbhargav:
@markandrus 

This PR contains changes for network handoff. As and when a commit is added for each of the tickets below, the corresponding checkbox will be checked.

- [x] [JSDK-2034](https://issues.corp.twilio.com/browse/JSDK-2034)
- [x] [JSDK-2035](https://issues.corp.twilio.com/browse/JSDK-2035)
- [x] [JSDK-2036](https://issues.corp.twilio.com/browse/JSDK-2036)
- [ ] [JSDK-2037](https://issues.corp.twilio.com/browse/JSDK-2037)

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -138,7 +138,8 @@ function connect(token, options) {
     preferredAudioCodecs: [],
     preferredVideoCodecs: [],
     realm: constants.DEFAULT_REALM,
-    signaling: SignalingV2
+    signaling: SignalingV2,
+    useTwilioConnection: false
   }, util.filterObject(options));
 
   /* eslint new-cap:0 */
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -139,7 +139,7 @@ function connect(token, options) {
     preferredVideoCodecs: [],
     realm: constants.DEFAULT_REALM,
     signaling: SignalingV2,
-    useTwilioConnection: false
+    _useTwilioConnection: false
   }, util.filterObject(options));
 
   /* eslint new-cap:0 */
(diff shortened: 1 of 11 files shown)


lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -0,0 +1,418 @@
...
+ * @event TwilioConnectionTransport#connected
+ * @param {object} initialState
+ */
+
+/**
+ * @event TwilioConnectionTransport#message
+ * @param {object} state
+ */
+
+function reducePeerConnections(peerConnections) {
```
markandrus: I think we should factor these and related methods for reducing RSP message out.

manjeshbhargav: @markandrus Since we are not using these functions anywhere else (SIP-based Transport) will go away eventually), I think having them here makes more sense.

lib/connect.js
```diff
@@ -138,7 +138,8 @@ function connect(token, options) {
     preferredAudioCodecs: [],
     preferredVideoCodecs: [],
     realm: constants.DEFAULT_REALM,
-    signaling: SignalingV2
+    signaling: SignalingV2,
+    _useTwilioConnection: false
```
markandrus: What do you think about naming this property be `transport` and letting it be one of "sip" or "tcmp", defaulting to "sip"?

markandrus:
One comment about property name, otherwise +1

lib/connect.js
manjeshbhargav: @markandrus Generally I would agree, if this were a public ConnectOptions flag. Since its private and temporary, I don't know if its worth the effort to change it.

manjeshbhargav:
As per discussion with @markandrus , merging this PR for now so that existing behavior can be tested. Another PR will be opened for implementing reconnection/network handoff.

PR #329, opened 2018-06-21
Add workaroundWebKitBug180748 option

markandrus:
_Not ready to merge. Still needs testing._

---

You can specify this in any of the following ways:

```js
connect(token, { audio: { workaroundWebKitBug180748: true });
createLocalAudioTrack({ workaroundWebKitBug180748: true });
createLocalTracks({ audio: { workaroundWebKitBug180748: true } });
```

In Safari, this *might* workaround [WebKit Bug 180748](https://bugs.webkit.org/show_bug.cgi?id=180748). In the worst case, it will slow down LocalAudioTrack creation.

### DIFF
--- a/lib/createlocaltrack.js
+++ b/lib/createlocaltrack.js
@@ -91,6 +91,10 @@ function createLocalVideoTrack(options) {
  * @property {LogLevel|LogLevels} logLevel
  * @property {string} [name] - The {@link LocalTrack}'s name; by default,
  *   it is set to the {@link LocalTrack}'s ID.
+ * @property {boolean} [workaroundWebKitBug180748=false] - Only valid for
+ *   {@link LocalAudioTrack}s; setting this attempts to workaround WebKit Bug
+ *   180748, where, in Safari, getUserMedia may return a silent audio
+ *   MediaStreamTrack.
  */
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


lib/webaudio/detectsilence.js
```diff
@@ -0,0 +1,46 @@
...
+   * @returns {Promise<boolean>}
+   */
+  function doDetectSilence() {
+    if (timeoutDidFire) {
+      return Promise.resolve(true);
+    }
+    analyser.getByteTimeDomainData(samples);
+    return samples.every(sample => sample)
+      ? Promise.resolve(false)
+      : doDetectSilence();
```
markandrus: This call needs to be delayed at least 1 tick.

ryan-rowland:
Nice fix. Implementation looks good. May be useful on Voice as well.

lib/createlocaltrack.js
```diff
@@ -91,6 +91,10 @@ function createLocalVideoTrack(options) {
  * @property {LogLevel|LogLevels} logLevel
  * @property {string} [name] - The {@link LocalTrack}'s name; by default,
  *   it is set to the {@link LocalTrack}'s ID.
+ * @property {boolean} [workaroundWebKitBug180748=false] - Only valid for
+ *   {@link LocalAudioTracks}; setting this attempts to workaround WebKit Bug
```
manjeshbhargav: @markandrus Nitpick: `{@link LocalAudioTrack}s`?

markandrus: Good catch—thanks.

PR #330, opened 2018-06-22
Track Subscription REST API support.

manjeshbhargav:
@markandrus 

TODO:

Unit + integration tests.

### DIFF
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -107,6 +107,9 @@ class Participant extends EventEmitter {
       _trackPublicationEventReemitters: {
         value: new Map()
       },
+      _trackSignalingUpdatedEventCallbacks: {
+        value: new Map()
+      },
       audioTracks: {
         enumerable: true,
         value: audioTracks
@@ -301,7 +304,7 @@ class Participant extends EventEmitter {
         trackSignalingSubscribed(signaling);
       }
 
-      signaling.on('updated', () => {
+      self._trackSignalingUpdatedEventCallbacks.set(signaling.sid, () => {
         if (isSubscribed !== signaling.isSubscribed) {
           isSubscribed = signaling.isSubscribed;
           if (isSubscribed) {
@@ -311,13 +314,18 @@ class Participant extends EventEmitter {
           trackSignalingUnsubscribed(signaling);
         }
       });
+      signaling.on('updated', self._trackSignalingUpdatedEventCallbacks.get(signaling.sid));
     }
 
     function trackSignalingRemoved(signaling) {
       if (signaling.isSubscribed) {
         signaling.setTrackTransceiver(null);
       }
-      signaling.removeAllListeners('updated');
+      const updated = self._trackSignalingUpdatedEventCallbacks.get(signaling.sid);
+      if (updated) {
+        signaling.removeListener('updated', updated);
+        self._trackSignalingUpdatedEventCallbacks.delete(signaling.id);
+      }
       const publication = self.trackPublications.get(signaling.sid);
       if (publication) {
         self._removeTrackPublication(publication);
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -296,39 +296,64 @@ class Participant extends EventEmitter {
       const publication = new RemoteTrackPublication(signaling, { log });
       self._addTrackPublication(publication);
 
-      signaling.getTrackTransceiver().then(trackReceiver => {
-        const RemoteTrack = {
-          audio: RemoteAudioTrack,
-          video: RemoteVideoTrack,
-          data: RemoteDataTrack
-        }[signaling.kind];
+      let isSubscribed = signaling.isSubscribed;
+      if (isSubscribed) {
+        trackSignalingSubscribed(signaling);
+      }
 
-        // NOTE(mroberts): It should never be the case that the TrackSignaling and
-        // MediaStreamTrack or DataTrackReceiver kinds disagree; however, just in
-        // case, we handle it here.
-        if (!RemoteTrack || signaling.kind !== trackReceiver.kind) {
-          return;
+      signaling.on('updated', () => {
+        if (isSubscribed !== signaling.isSubscribed) {
+          isSubscribed = signaling.isSubscribed;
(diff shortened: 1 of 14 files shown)


lib/participant.js
```diff
@@ -296,39 +296,64 @@ class Participant extends EventEmitter {
...
-        // MediaStreamTrack or DataTrackReceiver kinds disagree; however, just in
-        // case, we handle it here.
-        if (!RemoteTrack || signaling.kind !== trackReceiver.kind) {
-          return;
-        }
-
-        const track = signaling.kind === 'data'
-          ? new RemoteTrack(trackReceiver, { log, name: signaling.name })
-          : new RemoteTrack(trackReceiver, signaling.isEnabled, { log, name: signaling.name });
+      let isSubscribed = signaling.isSubscribed;
```
markandrus: Nitpick: I think we can prefer `const` here. Also, we could destructure to avoid repeating "isSubscribed".

lib/participant.js
```diff
@@ -296,39 +296,64 @@ class Participant extends EventEmitter {
...
-          self._removeTrack(track, publication);
-        }
-        if (publication) {
-          self._removeTrackPublication(publication);
-        }
-      });
+      if (signaling.isSubscribed) {
+        signaling.setTrackTransceiver(null);
+      }
+      signaling.removeAllListeners('updated');
```
markandrus: `removeAllListeners` is a risky API. If a class other than Participant set a listener on TrackSignaling, it will be removed, too. I recommend we do not use this API, and we do this differently.

lib/signaling/track.js
```diff
@@ -70,7 +58,7 @@ class TrackSignaling extends EventEmitter {
       trackTransceiver: {
         enumerable: true,
         get() {
-          return trackTransceiver;
+          return this._trackTransceiver;
         }
```
markandrus: Nitpick: We could convert this to a `get trackTransceiver() { return this._trackTransciever; }` on the `class`.

lib/signaling/track.js
```diff
@@ -109,21 +97,18 @@ class TrackSignaling extends EventEmitter {
...
    * @returns {this}
    */
+
   setTrackTransceiver(trackTransceiver) {
-    this._trackTransceiver = trackTransceiver;
+    trackTransceiver = trackTransceiver || null;
+    if (this.trackTransceiver !== trackTransceiver) {
+      this._trackTransceiver = trackTransceiver;
+      this.emit('updated');
+    }
```
markandrus: Nice 👍

lib/signaling/v2/room.js
```diff
@@ -232,6 +232,16 @@ class RoomV2 extends RoomSignaling {
           this._subscriptionFailures.set(trackState.sid, trackState.error);
         }
       }, this);
+
+      const subscribedTrackIds = new Set(roomState.subscribed.tracks
+        .filter(trackState => !!trackState.id)
```
markandrus: I guess we filter to filter out track subscription errors?

lib/signaling/v2/room.js
```diff
@@ -232,6 +232,16 @@ class RoomV2 extends RoomSignaling {
...
+        .map(trackState => trackState.id));
+
+      this._subscribed.forEach((trackSid, trackId) => {
+        if (!subscribedTrackIds.has(trackId)) {
+          this._subscribed.delete(trackId);
+        }
+      });
```
markandrus: Nice, one of the missing pieces :-)

lib/signaling/v2/room.js
```diff
@@ -516,6 +526,17 @@ function handleSubscriptionFailures(room) {
...
     }
   });
+
+  const subscribedTrackIds = new Set(Array.from(room._subscribed.keys()));
+  remoteTracks.forEach(remoteTrack => {
+    if (!subscribedTrackIds.has(remoteTrack.id)) {
+      remoteTrack.setTrackTransceiver(null);
+      return;
+    }
+    if (!remoteTrack.isSubscribed) {
```
markandrus: Should we have a clause that checks if the `remoteTrack`'s SID is in `room._subscribed`? i.e., we do the `_getTrackReceiver` call _unless_ we know we should be subscribed?

lib/participant.js
manjeshbhargav: @markandrus `isSubscribed` is updated every time "updated" is fired. So we should use `let` here.

markandrus: Ah, right 👍

lib/signaling/v2/room.js
manjeshbhargav: @markandrus I think this is implied if the first `if` statement (above this one) is false.

test/integration/spec/rest.js
```diff
@@ -0,0 +1,110 @@
...
+
+(enableRestApiTests ? describe : describe.skip)('REST APIs', function() {
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(60000);
+
+  describe('Track Subscription', () => {
+    let rooms;
+
+    before(async () => {
+      const options = Object.assign({ name: randomName() }, defaults);
```
markandrus: Recommend we use

```js
const { smallVideoConstraints } = require('../../lib/util');
```

test/integration/spec/rest.js
```diff
@@ -0,0 +1,110 @@
...
+            await subscribeTrack(publication, room);
+          }
+          subscribedOrUnsubscribedTrack = await trackSubscribedOrUnsubscribed;
+        });
+
+        it(`should emit "${event}" on the RemoteParticipant`, () => {
+          if (trackAction === 'unsubscribe') {
+            assert.equal(publication.track, null);
+          }
+          assert.equal(subscribedOrUnsubscribedTrack, trackAction === 'subscribe' ? publication.track : track);
```
markandrus: Could we either leave a comment explaining that `publication.track` should not equal `track` in the "subscribe" case or could we name the first track `originalTrack` and bind a new variable, `subsequentTrack`?

markandrus:
Overall looks great 👍

PR #335, opened 2018-06-27
JSDK 2063: Fix removeTracks

syerrapragada:
@markandrus @manjeshbhargav

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -410,7 +410,7 @@ class LocalParticipant extends Participant {
     const publication = this.unpublishTrack(track);
     track = publication && publication.track;
     stop = typeof stop === 'boolean' ? stop : true;
-    if (track && stop) {
+    if (track && stop && track.kind !== 'data' ) {
       track.stop();
       this._log.info('Stopped LocalTrack:', track);
     }
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -438,7 +438,7 @@ class LocalParticipant extends Participant {
     stop = typeof stop === 'boolean' ? stop : true;
     return this.unpublishTracks(tracks).map(publication => {
       const track = publication.track;
-      if (stop) {
+      if (stop && track.kind !== 'data') {
         track.stop();
       }
       return track;
--- a/test/unit/spec/localparticipant.js
+++ b/test/unit/spec/localparticipant.js
@@ -104,7 +104,7 @@ describe('LocalParticipant', () => {
         context(`when called with a Local${kind}Track`, () => {
           it('should not throw', () => {
             if (method === 'addTrack') {
-              mockPublishTrack(test.participant)
+              mockPublishTrack(test.participant);
             } else {
               mockUnpublishTrack(test.participant);
             }
@@ -134,7 +134,7 @@ describe('LocalParticipant', () => {
       beforeEach(() => {
         test = makeTest();
         if (method === 'addTracks' || method === 'publishTracks') {
-          mockPublishTrack(test.participant)
+          mockPublishTrack(test.participant);
         } else {
           mockUnpublishTrack(test.participant);
         }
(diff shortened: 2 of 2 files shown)


test/unit/spec/localparticipant.js
```diff
@@ -122,19 +122,16 @@ describe('LocalParticipant', () => {
     'unpublishTracks',
   ].forEach(method => {
     describe(`#${method}`, () => {
-      const trackMethod = method.slice(0, -1);
+      const trackMethod = method.match('(add|publish).*') ? 'publishTrack' : 'unpublishTrack';
```
markandrus: I think the former (`slice(0, -1)`) is actually a little easier to read than the `match`; however, I also see you've changed the result of `trackMethod` here: "addTracks" and "removeTracks" now map to "publishTrack' and "unpublishTrack" instead of "addTrack" and "removeTrack". Is this intentional? Also, is this change necessary?

markandrus: Is this change necessary? Also, I see that "addTracks" and "removeTracks" now map to "publishTrack" and "unpublishTrack", rather than "addTrack" and "removeTrack".

test/unit/spec/localparticipant.js
```diff
@@ -122,19 +122,16 @@ describe('LocalParticipant', () => {
...
       beforeEach(() => {
         test = makeTest();
         test.participant[trackMethod] = sinon.spy(track => {
-          if (trackMethod.startsWith('publish')) {
             // TODO(mroberts): Here is a very bare-bones "mock"
             // LocalTrackPublication. We should improve this later.
             return { track, sid: track.id };
-          }
-          return track.kind === 'audio' ? null : { foo: 'bar', stop: sinon.spy() };
         });
```
markandrus: OK, so with the change above, now `test.participant[trackMethod]` is always supposed to emulate `publishTrack` or `unpublishTrack`, therefore it's OK that we return `{ track, sid: track.id }` (a little mock TrackPublication)? Also, a small nitpick: can you de-indent the body of the `if`-clause you removed?

lib/localparticipant.js
```diff
@@ -438,7 +438,7 @@ class LocalParticipant extends Participant {
     stop = typeof stop === 'boolean' ? stop : true;
     return this.unpublishTracks(tracks).map(publication => {
       const track = publication.track;
-      if (stop) {
+      if (stop && track.kind !== 'data') {
```
markandrus: Nice, this is what I expected we needed to change 👍 Can you also include a test validating that we do not regress in this particular area, or does your test change below cover that?

test/unit/spec/localparticipant.js
syerrapragada: Yes. All the methods under test call either publishTrack or unpublishTrack. Hence the regex match above @markandrus

test/unit/spec/localparticipant.js
syerrapragada: It is necessary so that we mock the publishTrack and unpublishTrack.

test/unit/spec/localparticipant.js
syerrapragada: Lint didnt catch indentation block. does it?

lib/localparticipant.js
syerrapragada: We need to also apply this fix in removeTrack as well. Making changes

test/unit/spec/localparticipant.js
```diff
@@ -1018,3 +1020,15 @@ function makeSignaling(options) {
...
 }
+
+function mockPublishTrack(participant) {
+  participant['publishTrack'] = sinon.spy(track => {
+    return Promise.resolve({ track, sid: track.id });
+  });
+}
+
+function mockUnpublishTrack(participant) {
+  participant['unpublishTrack'] = sinon.spy(track => {
```
manjeshbhargav: @syerrapragada See my comment above.

markandrus:
@syerrapragada these changes look good. I see in Travis logs this failure:

```
/home/travis/build/twilio/twilio-video.js/test/unit/spec/localparticipant.js
   107:49  error    Missing semicolon                                     semi
   137:45  error    Missing semicolon                                     semi
   221:10  warning  Unexpected use of undefined                           no-undefined
  1025:15  error    ["publishTrack"] is better written in dot notation    dot-notation
  1031:15  error    ["unpublishTrack"] is better written in dot notation  dot-notation
```

If you fix these up, we can merge and cut a 1.11.0-rc1 today.

PR #345, opened 2018-07-06
Add `ice_servers` property to "connect" message

markandrus:
This property will indicate if

* `iceServers` was overridden in ConnectOptions, and therefore ECS was bypassed;
* Acquiring ICE servers from ECS succeeded; or
* Acquiring ICE servers from ECS failed.

**EDIT**: I've now manually tested the "success" case, the "failure" case, and the "overrode" case.

### DIFF
--- a/lib/iceserversource/nts.js
+++ b/lib/iceserversource/nts.js
@@ -93,7 +93,7 @@ class NTSIceServerSource extends EventEmitter {
         value: log
       },
       _status: {
-        value: 'failure',
+        value: null,
         writable: true
       },
       // This Deferred remains unresolved until `stop` is called. We use it to
@@ -110,23 +110,19 @@ class NTSIceServerSource extends EventEmitter {
       // This is the Access Token NTSIceServerSource makes requests to ECS with.
       _token: {
         value: token
-      },
-      isStarted: {
-        enumerable: true,
-        get() {
-          return !!this._currentPoll;
-        }
-      },
-      status: {
-        enumerable: true,
-        get() {
-          return this._status;
-        }
-      },
+      }
     });
 
     this._log.info('Created a new NTSIceServerSource');
     this._log.debug('ECS server:', this._ecsServer);
+  }
+
+  get isStarted() {
+    return !!this._currentPoll;
+  }
+
+  get status() {
+    return this._status;
   }
 
   start() {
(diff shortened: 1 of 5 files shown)


lib/iceserversource/nts.js
```diff
@@ -112,7 +116,13 @@ class NTSIceServerSource extends EventEmitter {
         get() {
           return !!this._currentPoll;
         }
-      }
+      },
+      status: {
```
manjeshbhargav: @markandrus Can we use a getter method here?

markandrus: Yes, good idea 👍

lib/iceserversource/nts.js
```diff
@@ -92,6 +92,10 @@ class NTSIceServerSource extends EventEmitter {
       _log: {
         value: log
       },
+      _status: {
```
manjeshbhargav: @markandrus Any danger of setting the default value to "failure" here?

lib/iceserversource/nts.js
```diff
@@ -206,8 +216,11 @@ function poll(client) {
     if (!config) {
       throw alreadyStopped;
     }
-    return parseECSConfig(client, config);
+    const iceServersAndTTL = parseECSConfig(client, config);
+    client._status = 'success';
+    return iceServersAndTTL;
   }).catch(error => {
+    client._status = 'failure';
```
manjeshbhargav: @markandrus If the default value is "failure" is this line really necessary?

manjeshbhargav:
@markandrus A couple of questions, but +1 overall.

lib/iceserversource/nts.js
markandrus: Hmm... the calling code in `connect` is supposed to ensure we get a first ICE server emitted from the IceServerSource before proceeding. So long as we retain that contract, there should be no danger here. Alternatively, we could allow for `null`. WDYT?

manjeshbhargav: @markandrus I think it would be nice to have a separate default value. That way, if we report it in the "connect" message, we know that it's actually not a failure, but some other problem. Ideally we should never report the default value.

markandrus: OK, let's allow for `null` now.

lib/iceserversource/nts.js
markandrus: This will poll multiple times, possibly succeeding or failing. Granted we only advertise/use the result once at connect-time, but still, I think we should set it here. I'm also making the default value `null`, so we'll need to keep this.

PR #348, opened 2018-07-11
fixed race condition on _userAgent property in Transport constructor …

mhuynh5757:
…| JSDK-2005
@manjeshbhargav @markandrus


markandrus:
Code change looks good and Appveyor passed (although that only runs a subset of tests). How about Travis CI? You may have to `git rebase master JSDK-2005` first to resolve the conflict.

lib/signaling/v2/transport.js
```diff
@@ -118,7 +114,21 @@ class Transport extends StateMachine {
...
+    Object.defineProperties(this, {
+      _session: {
+        value: createSession(this,
+          name,
+          accessToken,
+          localParticipant,
+          peerConnectionManager,
+          ua,
+          options.SIPJSMediaHandler,
+          options.negotiateNetworkQuality,
```
markandrus: Missing `options.iceServerSourceStats` after `options.SIPJSMediaHandler`.

PR #357, opened 2018-07-16
Updating CHANGELOG.md to add deprecation notes.

manjeshbhargav:
@markandrus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,60 @@
 1.12.0 (in progress)
 ====================
 
+Deprecations
+------------
+
+The following `1.x` APIs/events are now deprecated and scheduled for removal
+in `twilio-video.js@2.0.0`:
+
+- Participant's "trackAdded" and "trackRemoved" events
+- RemoteTrack's `id` property
+- RemoteTrack's `isSubscribed` property
+- RemoteTrack's "unsubscribed" event
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,12 +1,61 @@
...
+  });
+
+  participant.on('trackUnsubscribed', (track, publication) => {
+    console.log(`Unsubscribed from a RemoteTrack: ${track} ${publication}`);
+  });
+  ```
+
+### Migrating from RemoteTrack's deprecated properties and events
+
+- Instead of the RemoteTrack's `id` property, use its `sid` property.
```
markandrus: Recommendation: "use its `sid` or `name` property."

CHANGELOG.md
```diff
@@ -1,12 +1,61 @@
 1.12.0 (in progress)
 ====================
 
+Deprecations
+------------
+
+The following `1.x` APIs/events are now deprecated and scheduled for removal
+in `twilio-video.js@2.0.0`:
+
+- Participant's "trackAdded" and "trackRemoved" events
```
markandrus: Nice, btw, I think we need to update the usage example in README.md?

manjeshbhargav: @markandrus WDYT about updating README.md in the 2.x branch? I think we can do it once and for all there.

markandrus:
+1

PR #362, opened 2018-07-19
JSDK-2058: Clone DataTracks

markandrus:


markandrus:
@manjeshbhargav yeah, i believe we removed or disabled that test. let me revisit and add back before merging.

PR #364, opened 2018-07-23
JSDK-2087 Adding small group Room test.

manjeshbhargav:
@markandrus

### DIFF
--- a/test/integration/spec/rest.js
+++ b/test/integration/spec/rest.js
@@ -20,25 +20,47 @@ const {
 } = require('../../lib/util');
 
 const connect = require('../../../lib/connect');
+const { RoomMaxParticipantsExceededError } = require('../../../lib/util/twilio-video-errors');
 
 // TODO(mmalavalli): Use twilio-node to call Track Subscription REST APIs.
 
-function subscribedTracks(publication, room, trackAction) {
-  const { localParticipant, sid } = room;
+function post(resource, data) {
   return new Promise((resolve, reject) => {
     const request = https.request({
       auth: `${apiKeySid}:${apiKeySecret}`,
       headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
       hostname: 'video.twilio.com',
       method: 'POST',
-      path: `/v1/Rooms/${sid}/Participants/${localParticipant.sid}/SubscribedTracks`
+      path: resource
     }, response => {
       response.on('data', () => {});
       response.on('end', resolve);
     });
     request.once('error', reject);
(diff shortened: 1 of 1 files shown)


test/integration/spec/rest.js
```diff
@@ -20,26 +20,48 @@ const {
...
     request.end();
   });
+
+}
+
+function completeRoom(name) {
+  return post(`/v1/Rooms/${name}`, {
+    Status: 'completed'
+  });
+}
```
markandrus: IMO, we should start using twilio-node. See how we can complete a Room [here](https://github.com/twilio/twilio-node/blob/master/lib/rest/video/v1/room.js#L690).

test/integration/spec/rest.js
```diff
@@ -20,26 +20,48 @@ const {
...
+
+function createRoom(name, type) {
+  return post('/v1/Rooms', {
+    Type: type,
+    UniqueName: name
+  });
+}
```
markandrus: IMO, we should start using twilio-node. See how we can create a Room [here](https://github.com/twilio/twilio-node/blob/master/lib/rest/video/v1/room.js#L86).

test/integration/spec/rest.js
```diff
@@ -20,26 +20,48 @@ const {
...
+
+function subscribedTracks(publication, room, trackAction) {
+  const { localParticipant, sid } = room;
+  return post(`/v1/Rooms/${sid}/Participants/${localParticipant.sid}/SubscribedTracks`, {
+    Status: trackAction,
+    Track: publication.trackSid
+  });
```
markandrus: IMO, we should start using twilio-node. See how we can subscribe to/unsubscribe from Tracks [here](https://github.com/twilio/twilio-node/blob/master/lib/rest/video/v1/room/roomParticipant/roomParticipantSubscribedTrack.js#L336).

test/integration/spec/rest.js
manjeshbhargav: @markandrus `twilio-node` does not seem to work in the browser due to node-specific dependencies.

test/integration/spec/rest.js
manjeshbhargav: @markandrus `twilio-node` does not seem to work in the browser due to node-specific dependencies.

test/integration/spec/rest.js
manjeshbhargav: @markandrus `twilio-node` does not seem to work in the browser due to node-specific dependencies.

manjeshbhargav:
Re-opening against `master` here: https://github.com/twilio/twilio-video.js/pull/366

PR #367, opened 2018-07-23
JSDK-2025: Improve Object.keys & JSON.stringify behavior for public classes

markandrus:
_Work-in-progress. Not ready to merge._

When we moved to ES6 classes, we didn't really think about the effect of adopting ES6 class getters. These differ from the instance getters we previously defined, because they don't show up when calling `Object.getOwnProperties(instance)`. This is because they are defined on the `prototype`. As a consequence of this, `Object.assign` and object spread (`{ ...foo }`) might not copy all the properties we're interested in. Of course, you can always pick out the properties by hand, but it'd be nice if we could support this syntax. To do so, we'd have to abandon ES6 getters and move these back into the constructor.

So, open question, would we consider moving away from ES6 getters? I haven't begun implementing in this PR yet. **EDIT:** Based on discussion today, we're OK with it.

Changes included in this PR:

* [x] Adding `toJSON` (which is used by `JSON.stringify`) to public classes
* [x] Cleaning up enumerable, "own" properties (as returned by `Object.getOwnProperties(instances)`), and filtering out "extra" properties (including those set by EventEmitter's constructor).
* [x] Moving class-level getters back into constructors.

### DIFF
--- a/lib/media/track/index.js
+++ b/lib/media/track/index.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const EventEmitter = require('../../eventemitter');
-const buildLogLevels = require('../../util').buildLogLevels;
+const { buildLogLevels, valueToJSON } = require('../../util');
 const DEFAULT_LOG_LEVEL = require('../../util/constants').DEFAULT_LOG_LEVEL;
 const Log = require('../../util/log');
 
@@ -66,11 +66,7 @@ class Track extends EventEmitter {
   }
 
   toJSON() {
-    return {
-      id: this.id,
-      kind: this.kind,
-      name: this.name
-    };
+    return valueToJSON(this);
   }
 }
 
--- a/lib/media/track/index.js
+++ b/lib/media/track/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const EventEmitter = require('events').EventEmitter;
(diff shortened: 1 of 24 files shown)


lib/util/index.js
```diff
@@ -475,6 +475,38 @@ function checkIfSdpSemanticsIsSupported() {
...
+}
+
+/**
+ * Creates a new subclass which, in the constructor, sets all
+ * underscore-prefixed properties non-enumerable. This is useful for patching up
+ * classes like EventEmitter which may set properties like `_events`.
+ * @param {Function} klass
+ * @returns {Function} subclass
+ */
+function makeUnderscorePrefixedPropertiesSetByConstructorNonEnumerable(klass) {
```
manjeshbhargav: @markandrus As discussed, this can be `hidePrivateProperties(klass)`.

lib/media/track/index.js
```diff
@@ -64,6 +64,14 @@ class Track extends EventEmitter {
   get id() {
     return this._id;
   }
+
+  toJSON() {
```
manjeshbhargav: @markandrus As discussed, `toJSON` can be generic so that we only have to define it in the base classes.

syerrapragada:
+1

manjeshbhargav:
+1

PR #371, opened 2018-07-26
2.x

markandrus:
Let's take another look at the PR, see if anything stands out, then merge to master? We've already cut support-1.x for any backports/support we need to do on 1.x.


markandrus:
Chatted w/ @manjeshbhargav and we are +1

PR #373, opened 2018-07-26
Enable Dominant Speaker by default (2.x)

markandrus:


markandrus:
Chatted w/ @manjeshbhargav and we are +1

PR #376, opened 2018-07-27
Jsdk 1884 1.x

markandrus:


CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+1.13.0 (in progress)
+====================
+
+New Features
+------------
+
+- Previously, if a Room was completed via the REST API while your
```
manjeshbhargav: @markandrus 
```
- When a Room is completed using the REST API, the Room emits a "disconnected" event with a new parameter - a TwilioError with code 53118 and message "Room completed". In some applications, this will be expected, so you should set an event listener on the Room as follows:
```
?

CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
...
+  twilio-video.js-based application was connected to it, the Room would emit a
+  "disconnected" event, but it wouldn't explain _why_ the disconnect occurred.
+  Now, the "disconnected" event will include a TwilioError 53118, "Room
+  completed" (JSDK-1884). In some applications, this will be expected, so you
+  should set an event listener on the Room as follows:
+
+  ```js
+  room.once('disconnected', (room, error) => {
+    switch (error.code) {
```
manjeshbhargav: @markandrus 

What about the case where there is no `error`?

markandrus: Good point

PR #383, opened 2018-08-01
JSDK-2097: Always add dummyAudioTrack first

markandrus:
Notes:

* I had to update the call to `makeMediaStream`; otherwise, the test trivially passed because no other MediaStreamTracks were added, meaning the dummy MediaStreamTrack was always first.
* In the future, we should eliminate the dummy MediaStreamTrack in favor of a single call to `addTransceiver('audio')`.

### DIFF
--- a/lib/signaling/v2/peerconnectionmanager.js
+++ b/lib/signaling/v2/peerconnectionmanager.js
@@ -247,11 +247,10 @@ class PeerConnectionManager extends QueueingEventEmitter {
    */
   setTrackSenders(trackSenders) {
     const dataTrackSenders = new Set(trackSenders.filter(trackSender => trackSender.kind === 'data'));
-    const mediaTrackSenders = new Set(trackSenders.filter(trackSender => trackSender.kind === 'audio' || trackSender.kind === 'video'));
 
-    if (this._dummyAudioTrackSender) {
-      mediaTrackSenders.add(this._dummyAudioTrackSender);
-    }
+    const mediaTrackSenders = new Set([this._dummyAudioTrackSender]
+      .concat(trackSenders)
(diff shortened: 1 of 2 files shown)


lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -247,11 +247,12 @@ class PeerConnectionManager extends QueueingEventEmitter {
    */
   setTrackSenders(trackSenders) {
     const dataTrackSenders = new Set(trackSenders.filter(trackSender => trackSender.kind === 'data'));
-    const mediaTrackSenders = new Set(trackSenders.filter(trackSender => trackSender.kind === 'audio' || trackSender.kind === 'video'));
 
-    if (this._dummyAudioTrackSender) {
-      mediaTrackSenders.add(this._dummyAudioTrackSender);
-    }
+    const mediaTrackSenders = trackSenders.reduce((mediaTrackSenders, trackSender) =>
```
manjeshbhargav: @markandrus I think this can be simplified to:
```js
trackSenders = (this._dummyAudioTrackSender
  ? [this._dummyAudioTrackSender]
  : []).concat(trackSenders);

const mediaTrackSenders  = new Set(trackSenders.filter(trackSender => {
  return trackSender.kind === 'audio' || trackSender.kind === 'video';
}));
```
WDYT?

markandrus: I'd rather we not reassign `trackSenders`, so we could do this instead

```js
const mediaTrackSenders = new Set([this._dummyAudioTrackSender]
  .concat(mediaTrackSenders)
  .filter(trackSender => trackSender && (trackSender.kind === 'audio' || trackSender.kind === 'video')));
```

PR #384, opened 2018-08-01
Prep for twilio-video.js@2.0.0-rc1.

manjeshbhargav:
@markandrus 

* Updating `@twilio/webrtc` dependency to `3.0.0-rc1`.
* Hiding `EventEmitter`'s prototype non-method properties.
* Setting `version` to 2.0.0-dev.

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -481,12 +481,10 @@ function checkIfSdpSemanticsIsSupported() {
  * @returns {void}
  */
 function hidePrivateProperties(object) {
-  Object.getOwnPropertyNames(object).forEach(name => {
-    if (name.startsWith('_')) {
-      const descriptor = Object.getOwnPropertyDescriptor(object, name);
-      descriptor.enumerable = false;
-      Object.defineProperty(object, name, descriptor);
-    }
+  Object.getOwnPropertyNames(object).filter(name => {
+    return name.startsWith('_');
+  }).forEach(name => {
+    hideProperty(object, name);
   });
 }
 
@@ -503,8 +501,34 @@ function hidePrivatePropertiesInClass(klass) {
     constructor(...args) {
       super(...args);
       hidePrivateProperties(this);
+      hidePrototypeProperties(this, klass);
     }
   };
+}
+
+/**
+ * Hide a property of an object.
+ * @param {object} object
+ * @param {string} name
+ */
+function hideProperty(object, name) {
+  const descriptor = Object.getOwnPropertyDescriptor(object, name);
+  descriptor.enumerable = false;
+  Object.defineProperty(object, name, descriptor);
+}
+
+/**
+ * Hide the prototype properties of an object.
+ * @param {object} object
+ * @param {Function} klass
+ */
+function hidePrototypeProperties(object, klass) {
+  const { prototype } = klass;
+  Object.getOwnPropertyNames(prototype).filter(name => {
+    return typeof prototype[name] !== 'function';
+  }).forEach(name => {
+    hideProperty(object, name);
+  });
 }
 
 /**
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -481,27 +481,29 @@ function checkIfSdpSemanticsIsSupported() {
  * @returns {void}
  */
 function hidePrivateProperties(object) {
-  Object.getOwnPropertyNames(object).filter(name => {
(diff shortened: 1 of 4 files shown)


lib/util/index.js
```diff
@@ -503,10 +501,36 @@ function hidePrivatePropertiesInClass(klass) {
...
+
+/**
+ * Hide the prototype properties of an object.
+ * @param {object} object
+ * @param {Function} klass
+ */
+function hidePrototypeProperties(object, klass) {
+  const { prototype } = klass;
+  Object.getOwnPropertyNames(prototype).filter(name => {
+    return typeof prototype[name] !== 'function';
```
markandrus: This seems a little too general. What if we want to allow a public class with an ES6 getter on the prototype? Or what about classes like WebSocket that define `WebSocket.prototype.CLOSING` and other properties? I know we don't have these today, but I was thinking it would be better to apply only a fix specifically for `domain` in `lib/eventemitter.js`. WDYT?

lib/util/index.js
```diff
@@ -503,10 +501,36 @@ function hidePrivatePropertiesInClass(klass) {
...
+ * @param {string} name
+ */
+function hideProperty(object, name) {
+  const descriptor = Object.getOwnPropertyDescriptor(object, name);
+  descriptor.enumerable = false;
+  Object.defineProperty(object, name, descriptor);
+}
+
+/**
+ * Hide the prototype properties of an object.
```
markandrus: This should be "non-function prototype properties" I think.

lib/util/index.js
```diff
@@ -481,12 +481,10 @@ function checkIfSdpSemanticsIsSupported() {
...
-    if (name.startsWith('_')) {
-      const descriptor = Object.getOwnPropertyDescriptor(object, name);
-      descriptor.enumerable = false;
-      Object.defineProperty(object, name, descriptor);
-    }
+  Object.getOwnPropertyNames(object).filter(name => {
+    return name.startsWith('_');
+  }).forEach(name => {
+    hideProperty(object, name);
   });
```
markandrus: The number of properties here is small, so it doesn't really matter, but `filter` followed by `forEach` is going to traverse twice, whereas keeping the `if`-statement inside traverses once. If you do keep the `filter`, I'd at least recommend making it as terse as possible, i.e.

```js
Object.getOwnPropertyNames(object)
  .filter(name => name.startsWith('_'))
  .forEach(name => hideProperty(object, name));
```

lib/util/index.js
```diff
@@ -483,30 +483,55 @@ function checkIfSdpSemanticsIsSupported() {
 function hidePrivateProperties(object) {
   Object.getOwnPropertyNames(object).forEach(name => {
     if (name.startsWith('_')) {
-      const descriptor = Object.getOwnPropertyDescriptor(object, name);
-      descriptor.enumerable = false;
-      Object.defineProperty(object, name, descriptor);
+      hideProperty(object, name);
```
markandrus: 👍

lib/util/index.js
```diff
@@ -483,30 +483,55 @@ function checkIfSdpSemanticsIsSupported() {
...
+ * Hide the given public properties of an object.
+ * @param {object} object
+ * @param {Array<string>} props
+ */
+function hidePublicProperties(object, props) {
+  props.forEach(name => {
+    if (object.hasOwnProperty(name)) {
+      hideProperty(object, name);
+    }
+  });
```
markandrus: I thought the issue was that the Browserify EventEmitter sets a property on `prototype`. This no longer checks prototype, though.

lib/util/index.js
```diff
@@ -483,30 +483,55 @@ function checkIfSdpSemanticsIsSupported() {
...
+  descriptor.enumerable = false;
+  Object.defineProperty(object, name, descriptor);
+}
+
+/**
+ * Hide the given public properties of an object.
+ * @param {object} object
+ * @param {Array<string>} props
+ */
+function hidePublicProperties(object, props) {
```
markandrus: Worth letting this default to `[]` I think.

lib/util/index.js
manjeshbhargav: @markandrus I looked at the code, and EventEmitter's constructor explicitly sets an instance variable `domain` to `null`. So we have to hide it in the constructor.

markandrus:
+1

PR #386, opened 2018-08-03
JSDK-2095: DataTransport should send "ready" message

markandrus:
Since we move sending the "ready" message into DataTransport, it'll fire for Dominant Speaker and Network Quality APIs.


syerrapragada:
LGTM

PR #390, opened 2018-08-06
JSDK-2118: Reduce MediaStream usage in Firefox

markandrus:
I wonder if integration tests will be improved or not with these changes. I still want to do some manual testing before merging. **EDIT:** Oops, I didn't realize [Bug 1231414](https://bugzilla.mozilla.org/show_bug.cgi?id=1231414) is still a thing, so we can't just `addTrack` without a MediaStream argument.

### DIFF
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -71,7 +71,6 @@ class MediaTrack extends Track {
     this._log.debug('Started');
     this._isStarted = true;
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
       this._dummyEl.oncanplay = null;
     }
     // eslint-disable-next-line no-use-before-define
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -622,6 +622,13 @@ class PeerConnectionV2 extends StateMachine {
         description.sdp,
         this._preferredAudioCodecs,
         this._preferredVideoCodecs);
+      // NOTE(mroberts): Do this to reduce our MediaStream count in Firefox. By
+      // mapping MediaStream IDs in the SDP to "-", we ensure the "track" event
+      // doesn't include any new MediaStreams in Firefox. Its `streams` member
+      // will always be the empty Array.
+      if (isFirefox) {
+        description.sdp = filterOutMediaStreamIds(description.sdp);
+      }
     }
     description = new this._RTCSessionDescription(description);
     return this._peerConnection.setRemoteDescription(description).then(() => {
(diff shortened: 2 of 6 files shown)


lib/media/track/mediatrack.js
```diff
@@ -71,7 +71,6 @@ class MediaTrack extends Track {
     this._log.debug('Started');
     this._isStarted = true;
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
```
manjeshbhargav: @markandrus Since you've removed the `_detachElement()` call here, I don't see it being called anywhere in `VideoTrack` (it is being called in `AudioTrack`). Any reason why this is being removed from the common base class?

markandrus:
@manjeshbhargav thanks for the review. I'll have to either remove the change for JSDK-2119 or see if we can adopt `addTransceiver` more broadly before we merge. (I'd also like to get the 1.13.1 release out first.)

markandrus:
Just to recap MediaStream- and Firefox-related changes here:

* We no longer construct a MediaStream per-PeerConnectionV2.
  * In Group Rooms, we save 1 MediaStream.
  * In Peer-to-Peer Rooms, we save N MediaStreams.
* We no longer construct two distinct MediaStreams for "started" and "dimensionsChanged" VideoTrack events.
  * In both Group and Peer-to-Peer Rooms, we save 1 MediaStream per-VideoTrack.
* We map all remote MediaStream IDs to "-" in the SDP.
  * In both Group and Peer-to-Peer Rooms, we save 1 MediaStream per-RemoteParticipant.

For example, a Group Room with three Participants all sharing audio and video can expect the following improvements in Firefox.

| | **MediaStreams Before** | **MediaStreams After** |
|:--- |:--- |:--- |
| 1 PeerConnectionV2 | 1 | 0 |
| 3 AudioTracks | 3 | 3 |
| 3 VideoTracks | 6 | 3 |
| 2 RemoteParticipants | 2 | 0 |
| **Total** | 12 | 6 |

Also,

* We recycle m=sections by calling RTCRtpTransceiver's `stop` method whenever the RTCRtpTransceiver is inactive.
* We workaround Firefox [Bug 1481335](https://bugzilla.mozilla.org/show_bug.cgi?id=1481335) in Group Rooms.

lib/signaling/v2/peerconnection.js
```diff
@@ -139,7 +152,9 @@ class PeerConnectionV2 extends StateMachine {
         value: null
       },
       _localMediaStream: {
-        value: new options.MediaStream()
+        value: isFirefox
```
markandrus: Actually, I think this needs to be gated on `isFirefox && RTCPeerConnection.prototype.addTransceiver`.

PR #406, opened 2018-08-20
Firefox/RTCRtpTransceiver fixes for 1.x

markandrus:
We did an experiment with RTCRtpTransceivers in 1.4.0-rc1. It revealed some issues. This PR tries to fix them up.

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -82,6 +82,7 @@ class PeerConnectionV2 extends StateMachine {
     super('open', states);
 
     options = Object.assign({
+      dummyAudioMediaStreamTrack: null,
       iceServers: [],
       logLevel: DEFAULT_LOG_LEVEL,
       offerOptions: {},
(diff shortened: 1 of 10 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -1000,4 +1002,68 @@ function filterOutMediaStreamIds(sdp) {
   return sdp.replace(/a=msid:[^ ]+ /g, 'a=msid:- ');
 }
 
+/**
+ * @param {string} sdp
+ * @param {string} mid
+ * @returns {?string} direction
+ */
+function getTransceiverDirection(sdp, mid) {
+  const section = sdp.split('\r\nm=').find(section => section.match(`a=mid:${mid}`));
```
manjeshbhargav: @markandrus I think we can re-use `getMediaSections()` from `utils` here?

lib/signaling/v2/peerconnection.js
```diff
@@ -1000,4 +1002,68 @@ function filterOutMediaStreamIds(sdp) {
...
+  if (!section) {
+    return null;
+  } else if (section.match(/a=sendrecv/)) {
```
manjeshbhargav: @markandrus This can be:
```js
const matches = section.match(/a=(sendrecv|sendonly|recvonly|inactive)/);
if (!matches) {
  return null;
}
return matches[1];
```

PR #408, opened 2018-08-30
Merge support-1.x back into master

markandrus:
There were some improvements we should add from 1.x to 2.x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,43 @@ Breaking Changes
   longer maintains the deprecated Track-based collections.
 - We removed support for Bower.
 
+1.14.0 (August 28, 2018)
+========================
+
+New Features
+------------
+
+- Added a new property to ConnectOptions, `dominantSpeaker`, for enabling the
+  Dominant Speaker API. Once the Dominant Speaker API is generally available,
+  you will need to set the `dominantSpeaker` property to `true`. This will only
+  take effect in Group Rooms.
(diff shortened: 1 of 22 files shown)


lib/room.js
```diff
@@ -396,15 +396,21 @@ function connectParticipant(room, participantSignaling) {
     'trackSubscriptionFailed',
     'trackUnpublished',
     'trackUnsubscribed'
-  ].map(event => {
+  ].map(eventOrEventPair => {
```
manjeshbhargav: @markandrus This can be reverted back to the simpler re-emit logic, since "trackAdded" and "trackRemoved" are no longer emitted.

markandrus:
> Wow, so many commits on 1.x 😱 . I just want to make sure these commits only cover whatever changes were made for 1.x since we cut the support branch.

Understood, and yes—that's why I created the branch this way:

```
git checkout master
git checkout -b merge-support-1.x-back-into-master
git merge support-1.x
```

and then patched up the merge conflicts in the merge commit. Per Slack, although the commits are numerous, the changes here are mainly

- addition of `dominantSpeaker` and `networkQuality` ConnectOptions
- more conservative use/re-use of HTMLMediaElements for reducing MediaStream usage
- RTCRtpTransceiver support for Firefox (and related MediaStream usage reduction)
- Bug fix for our workaround for Issue 8329

The diff should match this.

> Can you update the `TwilioConnectionTransport` class with the changes in `Transport`?

Good catch 👍

PR #418, opened 2018-09-14
Signal end of candidates

syerrapragada:
@manjeshbhargav Implemented signal end-of-candidates on Edge browser.

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -29,6 +29,7 @@ const workaroundIssue8329 = require('../../util/sdp/issue8329');
 const isChrome = guessBrowser() === 'chrome';
 const isFirefox = guessBrowser() === 'firefox';
 const isSafari = guessBrowser() === 'safari';
+const isEdge = guessBrowser() === 'edge';
 
 /*
 PeerConnectionV2 States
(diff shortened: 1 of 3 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -279,8 +280,11 @@ class PeerConnectionV2 extends StateMachine {
    * @returns {Promise<void>}
    */
   _addIceCandidate(candidate) {
+    if (!(candidate || isEdge)) {
```
manjeshbhargav: @syerrapragada This can be simplified to:
```js
return Promise.resolve().then(() => {
  if (candidate || isEdge) {
    candidate = candidate && new this._RTCIceCandidate(candidate);
    return this._peerConnection.addIceCandidate(candidate);
  }
});
```

lib/signaling/v2/peerconnection.js
```diff
@@ -279,8 +280,11 @@ class PeerConnectionV2 extends StateMachine {
...
+      return Promise.resolve();
+    }
     return Promise.resolve().then(() => {
-      candidate = new this._RTCIceCandidate(candidate);
+      candidate = candidate && new this._RTCIceCandidate(candidate);
```
manjeshbhargav: @syerrapragada Add a NOTE about calling `addIceCandidate(null)` for Edge here?

PR #422, opened 2018-09-19
JSDK-2160 Stop transceivers only after remote "answer" is applied

manjeshbhargav:
@markandrus


markandrus:
LGTM. If this passes QE we need to backport to 1.x as well.

PR #423, opened 2018-09-19
Use unified sdpFormat for Edge

syerrapragada:
@manjeshbhargav @markandrus 
Should we care about sdpSemantics in the connect options?

I still need to test by passing sdpSemantics='plan-b' and check how Edge behaves.


markandrus:
> Should we care about sdpSemantics in the connect options?

For Edge, no. `sdpSemantics` is a Chrome-specific extension. What you've done here looks good.

> I still need to test by passing sdpSemantics='plan-b' and check how Edge behaves.

It should be ignored, AFAICT.

PR #429, opened 2018-09-26
TCMP Re-connect

manjeshbhargav:
@markandrus @syerrapragada 

This PR contains the implementation of re-connect to VSG with exponential backoff.

TODO:

Test multiple network handoff scenarios before handing off to QA.

### DIFF
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -172,13 +172,6 @@ class TwilioConnectionTransport extends StateMachine {
     setupEventListeners(this);
   }
 
-  get _isReconnecting() {
-    const { maxReconnectAttempts } = this._options;
-    return this.state === 'connected'
-      && this._reconnectAttemptsLeft >= 0
-      && this._reconnectAttemptsLeft < maxReconnectAttempts;
-  }
-
   /**
    * Send a Connect, Sync or Disconnect RSP message.
    * @private
@@ -368,10 +361,6 @@ function reduceUpdates(updates) {
 
 function setupEventListeners(transport) {
   function connect() {
-    if (transport._isReconnecting) {
-      transport.sync();
-      return;
-    }
     transport._sendConnectOrSyncOrDisconnectMessage();
   }
 
@@ -401,6 +390,9 @@ function setupEventListeners(transport) {
   }
 
   function reconnect() {
+    if (transport.state === 'connected') {
+      transport.preempt('syncing');
+    }
     transport._reconnectAttemptsLeft--;
     const { maxReconnectAttempts } = transport._options;
     const reconnectAttempts = maxReconnectAttempts - transport._reconnectAttemptsLeft;
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -386,6 +386,9 @@ function setupEventListeners(transport) {
   }
 
   function disconnect(error) {
+    if (transport.state === 'disconnected') {
+      return;
+    }
     if (!error) {
       transport.disconnect();
       return;
@@ -411,6 +414,9 @@ function setupEventListeners(transport) {
   }
 
   function startConnect() {
+    if (transport.state === 'disconnected') {
+      return;
+    }
     const twilioConnection = createOrResetTwilioConnection();
     twilioConnection.once('close', disconnect);
     twilioConnection.on('message', handleMessage);
@@ -463,6 +469,9 @@ function setupEventListeners(transport) {
         return;
       case 'syncing':
         switch (message.type) {
+          case 'error':
+            transport.disconnect(createTwilioError(message.code, message.message));
+            return;
           case 'connected':
           case 'update':
             transport._updatesReceived.push(message);
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -13,6 +13,9 @@ const {
   SignalingConnectionError
 } = require('../../util/twilio-video-errors');
 
+const MAX_RECONNECT_ATTEMPTS = 5;
+const RECONNECT_BACKOFF_JITTER = 10;
+const RECONNECT_BACKOFF_MS = 50;
 const SDK_NAME = `${packageInfo.name}.js`;
 const SDK_VERSION = packageInfo.version;
 const VERSION = 1;
@@ -79,6 +82,9 @@ class TwilioConnectionTransport extends StateMachine {
       InsightsPublisher,
       NullInsightsPublisher,
       TwilioConnection,
+      maxReconnectAttempts: MAX_RECONNECT_ATTEMPTS,
+      reconnectBackOffJitter: RECONNECT_BACKOFF_JITTER,
+      reconnectBackOffMs: RECONNECT_BACKOFF_MS,
       userAgent: util.getUserAgent()
     }, options);
     super('connecting', states);
@@ -117,8 +123,21 @@ class TwilioConnectionTransport extends StateMachine {
       _networkQuality: {
         value: options.networkQuality
       },
+      _options: {
+        value: options
+      },
       _peerConnectionManager: {
         value: peerConnectionManager
+      },
+      _reconnectAttemptsLeft: {
+        value: options.maxReconnectAttempts,
+        writable: true
+      },
+      _reconnectBackOffJitter: {
+        value: options.reconnectBackOffJitter
+      },
+      _reconnectBackOffMs: {
+        value: options.reconnectBackOffMs
       },
       _sdpSemantics: {
         value: options.sdpSemantics
@@ -128,7 +147,8 @@ class TwilioConnectionTransport extends StateMachine {
         writable: true
       },
       _twilioConnection: {
-        value: new options.TwilioConnection(wsServer, options)
+        value: null,
+        writable: true
       },
       _updatesReceived: {
         value: []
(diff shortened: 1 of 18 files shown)


test/integration/spec/localparticipant.js
```diff
@@ -802,11 +827,11 @@ describe('LocalParticipant', function() {
       };
     });
 
-    after(() => {
+    after(async () => {
       thisAudioTrack.stop();
       thisVideoTrack.stop();
-      thisRoom.disconnect();
-      thatRoom.disconnect();
+      [thisRoom, thatRoom].filter(room => room).forEach(room => room.disconnect());
```
syerrapragada: Do we need to filter?

test/integration/spec/localparticipant.js
```diff
@@ -956,8 +984,9 @@ describe('LocalParticipant', function() {
         });
       });
 
-      after(() => {
-        [thisRoom, ...thoseRooms].forEach(room => room.disconnect());
+      after(async () => {
+        [thisRoom, ...thoseRooms].filter(room => room).forEach(room => room.disconnect());
```
syerrapragada: Same as above. Is filter needed?

test/integration/spec/room.js
```diff
@@ -270,31 +279,29 @@ describe('Room', function() {
...
 
-    after(() => {
-      if (thisRoom) {
-        thisRoom.disconnect();
-      }
-      if (thatRoom) {
-        thatRoom.disconnect();
-      }
+    after(async () => {
+      [thisRoom, thatRoom].filter(room => room).forEach(room => room.disconnect());
```
syerrapragada: Same as above

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -340,7 +385,27 @@ function setupEventListeners(transport) {
...
+      transport.disconnect(new SignalingConnectionError());
+    }
+    reconnect();
+  }
+
+  function reconnect() {
+    transport._reconnectAttemptsLeft--;
+    const { maxReconnectAttempts } = transport._options;
+    const reconnectAttempts = maxReconnectAttempts - transport._reconnectAttemptsLeft;
+    const backOffMs = (1 << reconnectAttempts) * transport._reconnectBackOffMs;
```
markandrus: Nitpick: I think `2 ** x` is more readable than `1 << x`, e.g.

```js
(2 ** reconnectAttempts) * transport._reconnectionBackOffMs;
```

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -340,7 +385,27 @@ function setupEventListeners(transport) {
       transport.disconnect();
       return;
     }
-    transport.disconnect(new SignalingConnectionError());
+    if (transport._reconnectAttemptsLeft <= 0) {
+      transport.disconnect(new SignalingConnectionError());
+    }
+    reconnect();
```
markandrus: Should `connect` and `disconnect` be responsible for removing the "close" and "open" events on the current transport, respectively?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -332,6 +373,10 @@ function reduceUpdates(updates) {
 
 function setupEventListeners(transport) {
   function connect() {
+    if (transport._isReconnecting) {
+      transport.sync();
```
markandrus: Should `sync` should be private? It must be called in a particular way...

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -340,7 +385,27 @@ function setupEventListeners(transport) {
       transport.disconnect();
       return;
     }
-    transport.disconnect(new SignalingConnectionError());
+    if (transport._reconnectAttemptsLeft <= 0) {
+      transport.disconnect(new SignalingConnectionError());
```
markandrus: `return` after this? Otherwise, we'll hit the `reconnect` call.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -138,11 +158,32 @@ class TwilioConnectionTransport extends StateMachine {
...
         value: options.userAgent
+      },
+      _wsServer: {
+        value: wsServer
       }
     });
     setupEventListeners(this);
   }
 
+  get _isReconnecting() {
```
markandrus: I think this is misnamed. AFAICT, the underlying TwilioConnection can have been successfully replaced, all Sync/Synced traffic can have completed, and yet `_isReconnecting` will return `true`. To me, this feels more like a mixture of `_hasReconnected` and `_canReconnect`.

markandrus: Looking at how it's used below, actually, could this be as simple as the following?

```js
get _hasReconnected() {
  return this._reconnectAttemptsLeft < this._options.maxReconnectAttempts;
}
```

markandrus: Will `_resetReconnectAttemptsLeft` create issues here, though?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -428,7 +489,7 @@ function setupEventListeners(transport) {
         return;
       }
       case 'disconnected':
-        twilioConnection.removeListener('message', handleMessage);
+        transport._twilioConnection.removeListener('message', handleMessage);
```
markandrus: What about the TwilioConnection instances that were replaced due to reconnection? Where do we remove those "message" handlers?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -439,6 +500,20 @@ function setupEventListeners(transport) {
...
+/**
+ * Add random jitter to a given value in the range [-jitter, jitter].
+ * @private
+ * @param {number} value
+ * @param {number} jitter
+ * @returns {number} value + random(-jitter, +jitter)
+ */
+function withJitter(value, jitter) {
+  const rand = Math.random();
+  return value - jitter + Math.floor(2 * jitter * rand + 0.5);
```
markandrus: Yes, let's factor this out now rather than check in more duplicated code.

lib/signaling/v2/twilioconnectiontransport.js
markandrus: Ah, I see you fixed in a subsequent commit.

markandrus:
Also, I think a good sanity check for the test changes would be to temporarily disable ad-hoc Room creation on the video-javascript@twilio.com account that we use for testing. We cannot leave it this way, since we must allow older twilio-video.js integration test suites to continue working; however, it'd validate whether our new tests scheme is working.

test/integration/spec/localparticipant.js
manjeshbhargav: @syerrapragada If a test fails, the Room objects will be `undefined`. So it is just safe to check for non-undefined Room objects.

test/integration/spec/localparticipant.js
manjeshbhargav: @syerrapragada See my response above.

test/integration/spec/room.js
manjeshbhargav: @syerrapragada See my response above.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -138,11 +158,32 @@ class TwilioConnectionTransport extends StateMachine {
...
       }
     });
     setupEventListeners(this);
   }
 
+  get _isReconnecting() {
+    const { maxReconnectAttempts } = this._options;
+    return this.state === 'connected'
+      && this._reconnectAttemptsLeft >= 0
+      && this._reconnectAttemptsLeft < maxReconnectAttempts;
```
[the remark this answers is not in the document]
manjeshbhargav: @markandrus If we remove that clause, the `isReconnecting` will be true even `this._reconnectAttemptsLeft` is equal to `maxReconnectAttempts`.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @markandrus I think (1 << x) is pretty well established as power-of-2. So I would like to keep this as is.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @markandrus We are using `.once` and not `.on` for "close" and "open".

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @markandrus `_isReconnecting` will only be true if `transport.state` is "connected" and `0 <= transport._reconnectAttemptsLeft < maxReconnectAttempts`. After "synced" message is received, we reset `transport._reconnectAttempts` to `maxReconnectAttempts`. So now, `isReconnecting` will be false.

lib/signaling/v2/twilioconnectiontransport.js
markandrus: Right, but if "open" fires, a listener remains for "close" and vice-versa. I don't think we'll have a bug in the underlying TwilioConnection instance, but we'll be accruing listeners nevertheless. I'm not sure if this could lead to a memory leak or not.

lib/signaling/v2/twilioconnectiontransport.js
markandrus: Should we factor this out and let both this line and `createOrResetTwilioConnection` call this?

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @markandrus Since we're using an entirely new TwilioConnection upon re-connect, I don't think we'll be accruing listeners. and if the previous TwilioConnection gets garbage-collected, there should be no memory leaks. Let me make sure there are no closure references to it.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @markandrus Hmm... if we transition to "syncing" right after we get disconnected, then we might not need the `._isReconnecting` property, and the Room#reconnect event will be fired sooner than it is now. Good point. 👍

PR #430, opened 2018-09-26
Print the Edge experimental support warning

syerrapragada:
@markandrus @manjeshbhargav 
Should I include that P2P, Group rooms work? I have manually tested that they work

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -23,7 +23,7 @@ const util = require('./util');
 // belongs to. Each call to connect() increments this counter.
 let connectCalls = 0;
 
-let didPrintSafariWarning = false;
+let didPrintSafariOrEdgeWarning = false;
 
 /**
  * Connect to a {@link Room}.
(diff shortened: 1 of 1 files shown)


lib/connect.js
```diff
@@ -180,6 +180,17 @@ function connect(token, options) {
...
   }
 
+  if (guessBrowser() === 'edge'
+    && !didPrintSafariOrEdgeWarning
+    && (log.logLevel !== 'error' && log.logLevel !== 'off')) {
+    didPrintSafariOrEdgeWarning = true;
+    log.warn([
+      'This release of twilio-video.js includes experimental support for',
+      'Edge build 10547 and newer. Please test this release',
+      'and report any issues to https://github.com/twilio/twilio-video.js'
```
markandrus: How did you determine the build number? Also, I think it would be good if @ktoraskartwilio weighed in here. IMO, in terms of support, Chrome and Firefox are "tier 1", Safari is "tier 2", and Edge may be "tier 3" or below. I'd like this warning to reflect that, at least at the time of writing, Edge is perhaps "more experimental" than Safari ever was.

syerrapragada: Our implementation is based on Edge's ORTC API's and ORTC API's are supported on Edge since 10547.

PR #439, opened 2018-10-04
JSDK-2146 Using addTransceiver for unified plan Chrome.

manjeshbhargav:
@markandrus 

This PR enables the use of RTCPeerConnection's `addTransceiver` method for unified plan Chrome.
This depends on a version of twilio-webrtc.js which does not suppress the native RTCTrackEvent.

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -497,7 +497,7 @@ class PeerConnectionV2 extends StateMachine {
   _maybeReoffer(localDescription) {
     let shouldReoffer = this._shouldOffer;
 
-    // NOTE(mmalavalli): In Firefox, if the remote RTCPeerConnection sends
+    // NOTE(mmalavalli): For "unified-plan" sdps, if the remote RTCPeerConnection sends
     // an offer with fewer audio m= lines than the number of audio RTCRTPSenders
     // in the local RTCPeerConnection, then the local RTCPeerConnection creates
     // an answer with the same number of audio m= lines as in the offer. This
@@ -510,7 +510,7 @@ class PeerConnectionV2 extends StateMachine {
     // We can reduce the number of cases where renegotiation is needed by
     // re-introducing 'offerToReceiveAudio' to the default RTCOfferOptions with a
     // value > 1.
-    if (isFirefox) {
+    if (this._sdpFormat === 'unified') {
       const senders = this._peerConnection.getSenders().filter(sender => sender.track);
       shouldReoffer = ['audio', 'video'].reduce((shouldOffer, kind) => {
         const mediaSections = getMediaSections(localDescription.sdp, kind, '(sendrecv|sendonly)');
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -101,12 +101,16 @@ class PeerConnectionV2 extends StateMachine {
     const peerConnection = new RTCPeerConnection(configuration, options.chromeSpecificConstraints);
     const sdpFormat = getSdpFormat(options.sdpSemantics);
 
-    const localMediaStream = isFirefox && RTCPeerConnection.prototype.addTransceiver
+    const localMediaStream = sdpFormat === 'unified' && RTCPeerConnection.prototype.addTransceiver
       ? null
       : new options.MediaStream();
 
     if (options.dummyAudioMediaStreamTrack) {
-      peerConnection.addTrack(options.dummyAudioMediaStreamTrack, localMediaStream || new MediaStream());
+      if (localMediaStream) {
+        peerConnection.addTrack(options.dummyAudioMediaStreamTrack, localMediaStream);
+      } else {
+        peerConnection.addTransceiver(options.dummyAudioMediaStreamTrack);
+      }
     }
 
     // NOTE(mroberts): We do this to workaround the following bug:
--- a/lib/util/sdp/simulcast.js
+++ b/lib/util/sdp/simulcast.js
@@ -24,7 +24,8 @@ class TrackAttributes {
   /**
    * Construct a {@link MediaStreamTrack} attribute store.
    * @param {Track.ID} trackId - The MediaStreamTrack ID
-   * @param {string} streamId - The MediaStream ID
+   * @param {?string} streamId - The MediaStream ID for "plan-b",
+   *   null for "unified-plan"
    * @param {string} cName - The MediaStream cname
    */
   constructor(trackId, streamId, cName) {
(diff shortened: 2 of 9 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -101,12 +101,16 @@ class PeerConnectionV2 extends StateMachine {
     const peerConnection = new RTCPeerConnection(configuration, options.chromeSpecificConstraints);
     const sdpFormat = getSdpFormat(options.sdpSemantics);
 
-    const localMediaStream = isFirefox && RTCPeerConnection.prototype.addTransceiver
+    const localMediaStream = sdpFormat === 'unified' && RTCPeerConnection.prototype.addTransceiver
```
markandrus: We do this check a couple times—can we have a flag, `isUnified`, instead? IMO, it would be more readable and less error-prone in the event the string changes.

lib/signaling/v2/peerconnection.js
```diff
@@ -101,12 +101,16 @@ class PeerConnectionV2 extends StateMachine {
...
       ? null
       : new options.MediaStream();
 
     if (options.dummyAudioMediaStreamTrack) {
-      peerConnection.addTrack(options.dummyAudioMediaStreamTrack, localMediaStream || new MediaStream());
+      if (localMediaStream) {
+        peerConnection.addTrack(options.dummyAudioMediaStreamTrack, localMediaStream);
+      } else {
+        peerConnection.addTransceiver(options.dummyAudioMediaStreamTrack);
```
markandrus: I am not sure this is a safe refactor. If `offerToReceiveAudio` is true, `addTrack` and `addTransceiver` will yield different results. `addTrack` will reuse the media section introduced by `offerToReceiveAudio`, whereas `addTransceiver` will not. Since this code block deals with the `dummyAudioMediaStreamTrack`, which we intentionally want to be assigned to the first media section, I think we should stick with `addTrack`.

lib/util/sdp/simulcast.js
```diff
@@ -109,10 +110,15 @@ class TrackAttributes {
...
     const ssrcs = rtxPairs.length ? flatMap(rtxPairs) : simSSRCs;
 
+    const msidAttrPrefix = this.streamId
+      ? `${this.streamId} `
+      : '';
+
     const attrLines = flatMap(ssrcs, ssrc => [
       `a=ssrc:${ssrc} cname:${this.cName}`,
-      `a=ssrc:${ssrc} msid:${this.streamId} ${this.trackId}`
+      `a=ssrc:${ssrc} msid:${msidAttrPrefix}${this.trackId}`
```
markandrus: I do not think this is valid. If there is no MediaStream associated with the MediaStreamTrack, then we should stamp "-". Per [JSEP](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-24),

```
      *  If no MediaStream is associated with the transceiver, a single
         "a=msid" line with the special value "-" in place of the
         MediaStream ID, as specified in [I-D.ietf-mmusic-msid],
         Section 3.  The track ID MUST be selected as described above.
```

Also, rather than `msidAttrPrefix`, we could call this something like `msidId` to better match the definition in [draft-ietf-mmusic-msid-16](https://tools.ietf.org/html/draft-ietf-mmusic-msid-16#section-2).

lib/util/sdp/simulcast.js
```diff
@@ -179,6 +185,18 @@ function createTrackIdsToAttributes(section) {
...
+  // probably dealing with a "unified-plan" m= section.
+  if (ssrcAttrTuples.length === 0) {
+    const trackId = flatMap(getMatches(section, '^a=msid:.+ (.+)$'))[0];
+    const ssrcs = flatMap(getMatches(section, '^a=ssrc:(.+) cname:.+$'));
+    if (trackId) {
+      ssrcs.map(ssrc => [ssrc, null, trackId]).forEach(tuple => {
+        ssrcAttrTuples.push(tuple);
+      });
+    }
+  }
```
markandrus: Can we structure this code differently? When we call this code, we know what SDP format we are using. So can we not parameterize this function by the SDP format rather than use `ssrcAttrTuples.length === 0`?`ssrcAttrTuples.length === 0` seems indirect and might not actually indicate Unified Plan.

test/lib/mocksdp.js
```diff
@@ -73,20 +73,26 @@ a=mid:mid_${id}\r
 a=msid:stream ${id}\r
 `) + `\
 a=ssrc:${ssrc} cname:0\r
-a=ssrc:${ssrc} msid:stream ${id}\r
+a=ssrc:${ssrc} msid:${type === 'planb' ? 'stream ' : ''}${id}\r
```
markandrus: "-" rather than ""

test/lib/mocksdp.js
```diff
@@ -73,20 +73,26 @@ a=mid:mid_${id}\r
...
+ */
+function makeSdpForSimulcast(type, ssrcs) {
+  const sdp = makeSdpWithTracks(type, {
     audio: ['audio-1'],
     video: [{ id: 'video-1', ssrc: ssrcs[0] }]
   });
   const ssrcSdpLines = ssrcs.length === 2 ? [
     `a=ssrc:${ssrcs[1]} cname:0`,
-    `a=ssrc:${ssrcs[1]} msid:stream video-1`
+    `a=ssrc:${ssrcs[1]} msid:${type === 'planb' ? 'stream ' : ''}video-1`
```
markandrus: "-" rather than ""

lib/util/sdp/simulcast.js
```diff
@@ -169,15 +171,52 @@ function getSSRCRtxPairs(section) {
   return new Map(getMatches(section, rtxPairPattern).map(pair => pair.reverse()));
 }
 
+/**
+ * Create SSRC attribute tuples.
+ * @param {string} section
+ * @param {'planb' | 'unified'} sdpFormat
+ * @returns {Array<[SSRC, ?string, Track.ID]>}
```
markandrus: Can you clarify what the middle type is? I guess MediaStream ID? And you've said it's always null for the Unified Plan case? But that's not always true—there may be `a=msid` attributes that use a value other than `-`. So why make it null?

lib/util/sdp/simulcast.js
```diff
@@ -169,15 +171,52 @@ function getSSRCRtxPairs(section) {
...
+  return getMatches(section, '^a=ssrc:([0-9]+) msid:([^\\s]+) ([^\\s]+)$');
+}
+
+/**
+ * Create "unified-plan" SSRC attribute tuples.
+ * @param {string} section
+ * @returns {Array<[SSRC, null, Track.ID]>}
+ */
+function createUnifiedPlanSSRCAttributeTuples(section) {
+  const trackId = flatMap(getMatches(section, '^a=msid:.+ (.+)$'))[0];
```
markandrus: Nitpick: destructure?

PR #441, opened 2018-10-05
Update CHANGELOG.md

markandrus:


markandrus:
@manjeshbhargav that is already include below 👍

PR #444, opened 2018-10-11
JSDK-2146 Enabling addTransceiver for unified plan Chrome.

manjeshbhargav:
@markandrus 

As we discussed, I've added a NOTE about the [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=894231) and [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1498036) resulting in "trackSubscribed" events not being fired for RemoteTracks that are published a second time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
   },
   "dependencies": {
     "@twilio/sip.js": "^0.7.7",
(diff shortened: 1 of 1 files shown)


markandrus:
+1

PR #459, opened 2018-11-14
Disallow DataTrack creation in Edge browser

syerrapragada:


lib/media/track/localdatatrack.js
```diff
@@ -57,6 +58,11 @@ class LocalDataTrack extends Track {
    * @param {LocalDataTrackOptions} [options] - {@link LocalDataTrack} options
    */
   constructor(options) {
+    // NOTE(syerrapragada): DataChannels are not implemented yet in
```
manjeshbhargav: @syerrapragada Can you re-word this to:
```
otalk/peerconnectionshim does not support RTCDataChannels. So
DataTracks are not supported in Edge for now.
```

lib/media/track/localdatatrack.js
```diff
@@ -57,6 +58,11 @@ class LocalDataTrack extends Track {
...
+    // Edge and hence we cannot support Data tracks.
+    if (guessBrowser() === 'edge') {
+      throw new Error('Data tracks are not supported in Edge');
```
manjeshbhargav: @syerrapragada Can you re-word this to:
```
DataTracks are not supported in Edge.
```

lib/signaling/v2/remoteparticipant.js
```diff
@@ -95,7 +96,11 @@ class RemoteParticipantV2 extends RemoteParticipantSignaling {
    */
   addTrack(track) {
     super.addTrack(track);
-    this._getTrackReceiver(track.id).then(track.setTrackTransceiver.bind(track));
+    // NOTE(syerrapragada): DataChannels are rejected in Edge so there
```
manjeshbhargav: @syerrapragada Nitpick: "**RTCDataChannels** are not supported in Edge, so there is no **TrackReceiver** for the corresponding **RemoteDataTrackPublication**.

lib/signaling/v2/remoteparticipant.js
```diff
@@ -95,7 +96,11 @@ class RemoteParticipantV2 extends RemoteParticipantSignaling {
...
+    // is no trackReceiver for corresponding DataTrackPublication
+    if (guessBrowser() !== 'edge' || track.kind !== 'data') {
```
manjeshbhargav: @syerrapragada Can you print a warning to the console once to mention that you won't be able to subscribe to RemoteDataTracks on Edge?

```js
if (guessBrowser() === 'edge' && track.kind === 'data') {
  warnOnce('DataTracks are not supported in Edge. So you won't be able to '
    + 'subscribe to RemoteDataTracks.');
  return;
}
this._getTrackReceiver(track.id).then(track.setTrackTransceiver.bind(track));
```

manjeshbhargav:
@syerrapragada 

Also,

* Can you add a unit test for LocalDataTrack to make sure it throws on Edge? For that you would have to inject `guessBrowser()` as a dependency in LocalDataTrackOptions instead of using the module directly.
```js
const { guessBrowser } = require('../../util');

class LocalDataTrack {
  constructor(options) {
    options = Object.assign({
        guessBrowser,
    }, options);

    if (options.guessBrowser() === 'edge') {
      throw(...);
    }
  }
}
```

syerrapragada:
@manjeshbhargav Most of the integration tests create DataTrack. So making them run well on Edge is large effort than I initially estimated. So in this PR, I have added unit test and I will create another task for the integration test.

PR #464, opened 2018-11-22
Setting longer backoff and jitter values.

manjeshbhargav:
@innerverse


manjeshbhargav:
@innerverse @syerrapragada I'm merging this PR since I have to cut an RC very shortly.

PR #480, opened 2018-12-11
JSDK-2226

manjeshbhargav:
* JSDK-2226 - Transition to "disconnected" state for "error" RSP messages.
* ~~JSDK-2245 - Set version to 2 in RSP messages.~~

@syerrapragada

### DIFF
--- a/lib/signaling/v2/transport.js
+++ b/lib/signaling/v2/transport.js
@@ -25,9 +25,9 @@ const {
   createTwilioError,
 } = require('../../util/twilio-video-errors');
 
-const RSP_VERSION = 2;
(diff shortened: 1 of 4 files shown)


manjeshbhargav:
@syerrapragada While checking out integration test failures with group Rooms, I realized that bumping the RSP version to 2 involves more than just sending the version in the RSP payload. Here is the [implementation plan](https://paper.dropbox.com/doc/RSP-v2-Implementation-Plan--ATivXF~NSkx~SF9rL1g1vwiHAg-1O1vltIn44d70iDvVLaPA) which describes what needs to be done to use RSP v2.

PR #481, opened 2018-12-11
JSDK-2245 Bump RSP version to 2.

manjeshbhargav:
@syerrapragada 

Since we consume the `published` and `subscribed` payloads in RSP, we now bump the RSP version to 2.

### DIFF
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -357,7 +357,7 @@ class Participant extends EventEmitter {
     }
 
     function trackSignalingUnsubscribed(signaling) {
-      const track = self.tracks.get(signaling.id);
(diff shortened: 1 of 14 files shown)


manjeshbhargav:
@syerrapragada While checking out integration test failures with group Rooms, I realized that bumping the RSP version to 2 involves more than just sending the version in the RSP payload. Here is the [implementation plan](https://paper.dropbox.com/doc/RSP-v2-Implementation-Plan--ATivXF~NSkx~SF9rL1g1vwiHAg-1O1vltIn44d70iDvVLaPA) which describes what needs to be done to use RSP v2. So can you take a look at the changes that have been added since you last reviewed?

PR #482, opened 2018-12-12
Document that getStats is not supported on Safari yet

syerrapragada:


lib/room.js
```diff
@@ -132,6 +132,9 @@ class Room extends EventEmitter {
 
   /**
    * Get the {@link Room}'s media statistics.
+   *
```
manjeshbhargav: @syerrapragada Nitpick:
```
Get the {@link Room}'s media statistics. This is not supported in Safari 12.0 or below due to this bug: <link>
```

PR #521, opened 2019-01-30
Workaround a case where Chrome failed to set description due to dupli…

syerrapragada:
@manjeshbhargav 

This PR is a workaround for a case where Chrome produces an Offer/Answer with duplicate `rtx` payload mappings.

Excerpts from an Offer SDP with duplicate rtx mappings:
Case 1:
.....
**a=rtpmap:99 rtx/90000
a=fmtp:99 apt=126**
a=rtpmap:101 rtx/90000
a=fmtp:101 apt=100
a=rtpmap:97 rtx/90000
a=fmtp:97 apt=96
**a=rtpmap:99 rtx/90000
a=fmtp:99 apt=98**
a=rtpmap:121 rtx/90000
a=fmtp:121 apt=102
....

Case 2: 
....
a=rtpmap:97 H264/90000
a=rtcp-fb:97 goog-remb
a=rtcp-fb:97 ccm fir
a=rtcp-fb:97 nack
a=rtcp-fb:97 nack pli
a=fmtp:97 level-asymmetry-allowed=1;packetization-mode=0;profile-level-id=42e01f
**a=rtpmap:97 rtx/90000
a=fmtp:97 apt=96**
....

This PR detects duplicate mappings and deletes them:

1. If the original PT is mapped to a non-rtx stream, we will delete all duplicate rtx mapping from SDP
2. If the original PT is mapped to a rtx stream, we will preserve the first mapping and delete rest from SDP

### DIFF
--- a/lib/util/sdp/issue8329.js
+++ b/lib/util/sdp/issue8329.js
@@ -191,7 +191,7 @@ function createAssociatedPtToRtxPt(rtxPtToAssociatedPt, invalidRtxPts) {
  * @returns {string} newMediaSection
  */
 function deleteFmtpAttributesForRtxPt(mediaSection, rtxPt) {
-  const pattern = new RegExp(`a=fmtp:${rtxPt} apt=.*\r\n`, 'gm');
+  const pattern = new RegExp(`a=fmtp:${rtxPt}.*\r\n`, 'gm');
   return mediaSection.replace(pattern, '');
 }
 
@@ -201,7 +201,7 @@ function deleteFmtpAttributesForRtxPt(mediaSection, rtxPt) {
  * @returns {string} newMediaSection
  */
 function deleteRtpmapAttributesForRtxPt(mediaSection, rtxPt) {
-  const pattern = new RegExp(`a=rtpmap:${rtxPt} rtx.*\r\n`, 'gm');
+  const pattern = new RegExp(`a=rtpmap:${rtxPt}.*\r\n`, 'gm');
   return mediaSection.replace(pattern, '');
 }
 
--- a/lib/util/sdp/issue8329.js
+++ b/lib/util/sdp/issue8329.js
@@ -96,33 +96,16 @@ function mediaSectionWorkaround(mediaSection) {
 function deleteDuplicateRtxPts(mediaSection, ptToCodecName) {
   // NOTE(syerrapragada): In some cases Chrome produces an offer/answer
   // with duplicate "rtx" payload mapping in media section. When applied,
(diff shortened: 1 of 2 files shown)


lib/util/sdp/issue8329.js
```diff
@@ -87,6 +88,43 @@ function mediaSectionWorkaround(mediaSection) {
   return mediaSection;
 }
 
+/**
+ * @param {string} mediaSection
+ * @param {Map<PT, Codec>} ptToCodecName
+ * @returns @param {string} newMediaSection
+ */
+function deleteDuplicateRtxPts(mediaSection, ptToCodecName) {
```
manjeshbhargav: @syerrapragada Can we generalize this to remove duplicate rtp payload type mappings irrespective of whether they map to `rtx` or not?

lib/util/sdp/issue8329.js
```diff
@@ -87,6 +88,43 @@ function mediaSectionWorkaround(mediaSection) {
...
+  // NOTE(syerrapragada): In some cases Chrome produces an offer/answer
+  // with duplicate "rtx" payload mapping in media section. When applied,
+  // Chrome rejects the SDP. We work around this by deleting duplicate
+  // "rtx" mappings found in SDP.
+  Array.from(ptToCodecName.keys()).forEach(pt => {
+    // 1. Find if there are duplicate PTs
+    const rtpmapPattern = new RegExp(`a=rtpmap:${pt}.*`, 'g');
+    const matches = mediaSection.match(rtpmapPattern);
+    if (matches && matches.length > 1) {
+      // 2. If orig is not 'rtx' type, delete all duplicate 'rtx' pts
```
manjeshbhargav: @syerrapragada 

If we generalize to delete any duplicate mapping, then we just need `3.` and we can get rid of `.2`. WDYT?

manjeshbhargav:
@syerrapragada 

Just to recap, what if we simplify the logic like this:
```js
function deleteDuplicatePayloadTypeMappings(section, ptToCodecName) {
  return Array.from(ptToCodecName.keys()).reduce((section, pt) => {
    const rtpmapRegex = new RegExp(`^a=rtpmap:${pt} .+$`, 'gm');
    return (section.match(rtpmapRegex) || []).slice(1).reduce((section, rtpmap) => {
      const rtpmapRegex = new RegExp(`\r\n${rtpmap}`);
      const fmtpmapRegex = new RegExp(`\r\na=fmtp:${pt} apt=[0-9]+`);
      return section.replace(rtpmapRegex, '').replace(fmtpMapRegex, '');
    }, section);
  });
}
```
?

PR #527, opened 2019-02-05
JSDK-2219 Queueing calls to createOffer() until the current negotiation is complete.

manjeshbhargav:
@syerrapragada 

~~WIP.~~

While #518 reduced the instances of SDPs with changing mids for the same m= sections during negotiation, it did not fully eliminate them because we still can create an offer in the middle of a negotiation (that is while an offer has already been sent and awaiting an answer).

This PR makes sure that calls to createOffer() are queued until the existing offer <--> answer negotiation is complete.

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -177,7 +177,7 @@ class PeerConnectionV2 extends StateMachine {
       _mediaTrackReceivers: {
         value: new Set()
       },
-      _needsInitialAnswer: {
+      _needsAnswer: {
         writable: true,
         value: false
       },
@@ -553,6 +553,7 @@ class PeerConnectionV2 extends StateMachine {
    */
   _offer() {
     const offerOptions = Object.assign({}, this._offerOptions);
+    this._needsAnswer = true;
     if (this._shouldRestartIce) {
       this._shouldRestartIce = false;
       this._isRestartingIce = true;
@@ -573,7 +574,6 @@ class PeerConnectionV2 extends StateMachine {
       this._shouldOffer = false;
       if (!this._negotiationRole) {
         this._negotiationRole = 'offerer';
-        this._needsInitialAnswer = true;
       }
       return this._setLocalDescription({
         type: 'offer',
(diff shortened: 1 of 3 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -328,14 +328,23 @@ class PeerConnectionV2 extends StateMachine {
...
+    }).then(answer => {
       if (!isFirefox) {
         answer = workaroundIssue8329(answer);
       }
       return this._setLocalDescription(answer);
-    }).then(() => this._checkIceBox(offer)).then(() => this._peerConnection.localDescription
-      ? this._maybeReoffer(this._peerConnection.localDescription)
-      : Promise.resolve(false)).catch(error => {
+    }).then(() => {
+      this._needsAnswer = false;
```
syerrapragada: Do we need to set this to false here?

lib/signaling/v2/peerconnection.js
```diff
@@ -387,9 +396,13 @@ class PeerConnectionV2 extends StateMachine {
       this._isRestartingIce = false;
       this._shouldRestartIce = true;
     }
-    return Promise.resolve().then(() => this._setLocalDescription({ type: 'rollback' }))
-      .then(() => this._answer(offer))
-      .then(didReoffer => didReoffer ? Promise.resolve() : this._offer());
+    return Promise.resolve().then(() => {
+      return this._setLocalDescription({ type: 'rollback' });
+    }).then(() => {
+      return this._answer(offer);
```
syerrapragada: This line answers above question. But could we set `needsAnswer = false` here right after rollback? and remove it from `_answer(...)` function?

manjeshbhargav: @syerrapragada `_answer()` may re-offer, so I think the existing location is the ideal place for setting `_needAnswer` to false.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @syerrapragada see comment below.

lib/signaling/v2/peerconnection.js
syerrapragada: We have already rolled back the previous offer successfully so do we still need to wait for an answer?

PR #537, opened 2019-02-12
ICE Improvements

manjeshbhargav:
@syerrapragada @ceaglest 

This PR contains the following changes:

- [x] Don't send RTCIceCandidates to ICE-lite agents. (JSDK-2188)
- [x] Avoid unnecessary re-negotiations for LocalDataTrack (un)publishing (JSDK-2256 and JSDK-2274).

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -105,14 +105,11 @@ class PeerConnectionV2 extends StateMachine {
       RTCSessionDescription: DefaultRTCSessionDescription
     }, options);
 
-    const configuration = getConfiguration(options);
-    const logLevels = buildLogLevels(options.logLevel);
     const RTCPeerConnection = options.RTCPeerConnection;
+    const configuration = getConfiguration(options);
+    const localMediaStream = isUnifiedPlan ? null : new options.MediaStream();
+    const logLevels = buildLogLevels(options.logLevel);
     const peerConnection = new RTCPeerConnection(configuration, options.chromeSpecificConstraints);
-
-    const localMediaStream = isUnifiedPlan && RTCPeerConnection.prototype.addTransceiver
-      ? null
-      : new options.MediaStream();
 
     if (options.dummyAudioMediaStreamTrack) {
       peerConnection.addTrack(options.dummyAudioMediaStreamTrack, localMediaStream || new options.MediaStream());
@@ -298,6 +295,21 @@ class PeerConnectionV2 extends StateMachine {
   }
 
   /**
+   * Whether the {@link PeerConnectionV2} has negotiated or is in the process
+   * of negotiating the application m= section.
+   * @returns {boolean}
+   */
+  get isApplicationSectionNegotiated() {
+    // NOTE(mmalavalli): Because of VIDEO-1981, we only avoid re-negotiations
+    // for LocalDataTrack changes when the remote peer is an SFU.
+    // TODO(mmalavalli): Remove the ICE-lite agent check once VIDEO-1981
+    // is fixed.
+    return this._isIceLite && this._peerConnection.localDescription
+      ? getMediaSections(this._peerConnection.localDescription.sdp, 'application').length > 0
+      : false;
+  }
+
+  /**
    * Add an ICE candidate to the {@link PeerConnectionV2}.
    * @private
    * @param {object} candidate
@@ -459,7 +471,6 @@ class PeerConnectionV2 extends StateMachine {
     const peerConnectionState = {
       ice: {
         candidates: this._isIceLite ? [] : this._localCandidates.slice(),
-        revision: this._localCandidatesRevision++,
         ufrag: this._localUfrag
       },
       id: this.id
@@ -468,6 +479,7 @@ class PeerConnectionV2 extends StateMachine {
       peerConnectionState.ice.complete = true;
     }
     if (!(this._isIceLite && event.candidate)) {
+      peerConnectionState.ice.revision = this._localCandidatesRevision++;
       this.emit('candidates', peerConnectionState);
     }
   }
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -300,11 +300,7 @@ class PeerConnectionV2 extends StateMachine {
    * @returns {boolean}
    */
   get isApplicationSectionNegotiated() {
-    // NOTE(mmalavalli): Because of VIDEO-1981, we only avoid re-negotiations
-    // for LocalDataTrack changes when the remote peer is an SFU.
-    // TODO(mmalavalli): Remove the ICE-lite agent check once VIDEO-1981
-    // is fixed.
-    return this._isIceLite && this._peerConnection.localDescription
+    return this._peerConnection.localDescription
       ? getMediaSections(this._peerConnection.localDescription.sdp, 'application').length > 0
       : false;
   }
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -140,6 +140,10 @@ class PeerConnectionV2 extends StateMachine {
       _encodingParameters: {
         value: encodingParameters
       },
+      _isIceLite: {
+        writable: true,
+        value: false
+      },
       _isRestartingIce: {
         writable: true,
         value: false
@@ -454,7 +458,7 @@ class PeerConnectionV2 extends StateMachine {
     }
     const peerConnectionState = {
       ice: {
-        candidates: this._localCandidates.slice(),
+        candidates: this._isIceLite ? [] : this._localCandidates.slice(),
         revision: this._localCandidatesRevision++,
         ufrag: this._localUfrag
       },
@@ -463,7 +467,9 @@ class PeerConnectionV2 extends StateMachine {
     if (!event.candidate) {
       peerConnectionState.ice.complete = true;
     }
-    this.emit('candidates', peerConnectionState);
+    if (!(this._isIceLite && event.candidate)) {
+      this.emit('candidates', peerConnectionState);
+    }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -1975,6 +2007,9 @@ function makeDescription(type, options) {
       type === 'pranswer') {
     const session = 'session' in options ? options.session : Number.parseInt(Math.random() * 1000);
     description.sdp = 'o=- ' + session + '\r\n';
+    if (options.iceLite) {
+      description.sdp += 'a=ice-lite\r\n';
+    }
```
ceaglest: What if the agent is not ice-lite, should the mocked sdp indicate trickle-ice support?

test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -1326,84 +1346,96 @@ describe('PeerConnectionV2', () => {
...
-        });
+        if (which === 'final') {
+          it('with completed set to true', () => {
+            assert(iceState.ice.complete);
+          });
+        } else {
+          it('with completed unset', () => {
+            assert(!iceState.ice.complete);
+          });
+        }
```
ceaglest: I believe this test confirms that PeerConnection will emit a single event with no candidates and ice.complete set to true.

lib/signaling/v2/peerconnection.js
```diff
@@ -426,7 +430,9 @@ class PeerConnectionV2 extends StateMachine {
     if (!event.candidate) {
       peerConnectionState.ice.complete = true;
     }
-    this.emit('candidates', peerConnectionState);
+    if (!(this._isIceLite && event.candidate)) {
+      this.emit('candidates', peerConnectionState);
+    }
```
ceaglest: This is the simplest way to emit the end-of-candidates event (and no ice candidates) after encountering an ice-lite agent. The approach I took was to emit the end of candidates event as soon the Client:

1. Applies a remote ice-lite answer.
2. Applies a local answer to an ice-lite offer.

The benefit is the signal comes sooner and more reliably (since gathering takes up to 40 seconds in Chrome). The drawback is more complexity in event handling. I'll let you decide what makes sense, because what you have here still meets the spirit of the RSP's [PeerConnection Handling](
https://code.hq.twilio.com/client/room-signaling-protocol/blob/master/docs/peerconnection_handling.md#ice-lite).

manjeshbhargav: @ceaglest ,

I think sending end of candidates when the PeerConnection actually finishes gathering candidates will be more accurate just from a logging perspective, unless we are gaining anything performance-wise by sending it sooner.

test/unit/spec/signaling/v2/peerconnection.js
manjeshbhargav: @ceaglest 

`a=ice-lite` is a session attribute, where as `a=ice-options:trickle` is an m= section specific attribute. So, I think we can safely assume that if `a=ice-lite` is not present, then the peer either supports trickle-ice or it gathers all candidates beforehand.

manjeshbhargav:
@ceaglest 

Thanks for your comments. We don't have integration tests for this. Travis is pretty flaky right now, and I don't want to add tests which might fail some of the time. We already have some tests like this, which pass on the local machine but fail on Travis. Maybe Circle CI is more stable, so we will add these kinds of tests once we move to Circle CI.

lib/signaling/v2/peerconnection.js
```diff
@@ -417,7 +421,7 @@ class PeerConnectionV2 extends StateMachine {
     }
```
syerrapragada: Do we need to cache candidates in case of Ice-lite?

lib/signaling/v2/peerconnection.js
```diff
@@ -417,7 +421,7 @@ class PeerConnectionV2 extends StateMachine {
...
     const peerConnectionState = {
       ice: {
-        candidates: this._localCandidates.slice(),
+        candidates: this._isIceLite ? [] : this._localCandidates.slice(),
         revision: this._localCandidatesRevision++,
```
syerrapragada: We are incrementing revision even when we are not emitting candidates event. Is that ok?

lib/signaling/v2/peerconnection.js
```diff
@@ -276,6 +277,21 @@ class PeerConnectionV2 extends StateMachine {
...
+   */
+  get isApplicationSectionNegotiated() {
+    // NOTE(mmalavalli): Because of VIDEO-1981, we only avoid re-negotiations
+    // for LocalDataTrack changes when the remote peer is an SFU.
+    // TODO(mmalavalli): Remove the ICE-lite agent check once VIDEO-1981
+    // is fixed.
+    return this._isIceLite && this._peerConnection.localDescription
+      ? getMediaSections(this._peerConnection.localDescription.sdp, 'application').length > 0
+      : false;
+  }
```
ceaglest: We have a similar check for the presence of an `m=application` section in C++, so that we can re-negotiate to add the section if needed. 

I am slightly confused about the VIDEO-1981 workaround, because the ticket describes Peer-to-Peer and your comment mentions SFU.

lib/signaling/v2/peerconnection.js
ceaglest: I am okay either way, but I would also like to report ice gathering to insights in the future 👍.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @ceaglest Because of VIDEO-1981, we always need to re-negotiate when we are in a Peer-to-Peer Room. So, we only avoid re-negotiation when we are in a group Room (the assumption here is that an ICE-lite agent is the SFU). Does that clarify things for you?

test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -32,6 +32,34 @@ describe('PeerConnectionV2', () => {
...
+    context('when the underlying RTCPeerConnection has a local RTCSessionDescription', () => {
+      [true, false].forEach(hasApplicationSection => {
+        context(`when the RTCSessionDescription ${hasApplicationSection ? 'has' : 'does not have'} an application m= section`, () => {
+          it(`should be set to ${hasApplicationSection}`, async () => {
+            const test = makeTest({
+              offers: [
+                makeOffer({ application: hasApplicationSection })
+              ]
+            });
+            // TODO(mmalavalli): Remove this line once VIDEO-1981 is fixed.
```
makarandp0: https://issues.corp.twilio.com/browse/VIDEO-1981 is closed now. can this TODO be removed?

test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -32,6 +32,34 @@ describe('PeerConnectionV2', () => {
...
+            await test.pcv2.offer();
+            assert.equal(test.pcv2.isApplicationSectionNegotiated, hasApplicationSection);
+          });
+        });
+      });
+    });
+    context('when the underlying RTCPeerConnection does not have a local RTCSessionDescription', () => {
+      it('should be set to false', async () => {
+        const test = makeTest();
+        // TODO(mmalavalli): Remove this line once VIDEO-1981 is fixed.
```
makarandp0: https://issues.corp.twilio.com/browse/VIDEO-1981 is closed now. can this TODO be removed? looks like this comment is repeated multiple times. Can we get rid of all instances.

lib/signaling/v2/peerconnection.js
```diff
@@ -454,16 +466,18 @@ class PeerConnectionV2 extends StateMachine {
     }
     const peerConnectionState = {
       ice: {
-        candidates: this._localCandidates.slice(),
-        revision: this._localCandidatesRevision++,
+        candidates: this._isIceLite ? [] : this._localCandidates.slice(),
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I think our current babel config does not parse `...`.

lib/signaling/v2/peerconnection.js
```diff
@@ -718,6 +732,9 @@ class PeerConnectionV2 extends StateMachine {
       if (isFirefox) {
         description.sdp = filterOutMediaStreamIds(description.sdp);
       }
+      if (!this._peerConnection.remoteDescription) {
+        this._isIceLite = /a=ice-lite/.test(description.sdp);
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 

> 1. you are setting this._isIceLite only if remoteDescription is not already set - is it because 2nd setRemoteDescription would get rejected anyways?

Determining whether the remote peer is an ICE-lite agent is a one time operation per PeerConnection. So we only calculate it when we receive the first remote RTCSessionDescription.

> 2. You are setting this._isIceLite before this._peerConnection.setRemoteDescription resolves. if it rejects then we would we end-up with this._isIceLite set to invalid value?

`_iceLite` should be calculated only once, irrespective of whether `setRemoteDescription` resolves or rejects because its calculation is not dependent on whether the SDP is valid or not. It is only dependent on whether `a=ice-lite` is present in the SDP.

lib/signaling/v2/peerconnectionmanager.js
```diff
@@ -367,7 +367,11 @@ function applyTrackSenderChanges(peerConnectionManager, changes) {
       changes.media.remove.forEach(peerConnection.removeMediaTrackSender, peerConnection);
       changes.data.add.forEach(peerConnection.addDataTrackSender, peerConnection);
       changes.media.add.forEach(peerConnection.addMediaTrackSender, peerConnection);
-      peerConnection.offer();
+      if (changes.media.add.size
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I agree. Can you find out if there is an eslint rule that we can add that throws an error for falsy checks? We can add that rule and then make the necessary changes in a separate pass.

lib/signaling/v2/peerconnection.js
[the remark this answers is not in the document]
makarandp0: makes sense. Thanks

PR #579, opened 2019-03-19
JSDK-2314 Enabling VP8 simulcast support for Safari.

manjeshbhargav:
@syerrapragada


lib/signaling/v2/peerconnection.js
```diff
@@ -612,7 +611,7 @@ class PeerConnectionV2 extends StateMachine {
         // properties are immutable.
         description = {
           type: description.type,
-          sdp: isChrome && vp8SimulcastRequested
+          sdp: !isFirefox && vp8SimulcastRequested
```
syerrapragada: Should we explicitly enable vp8Simulcast for Chrome and Safari? `!isFirefox` may also enable it for other browsers like Edge in the future.

PR #587, opened 2019-03-23
JSDK-2294 COMMON_ISSUES.md entry for autoplay policy

manjeshbhargav:
@syerrapragada @ryan-rowland

### DIFF
--- a/COMMON_ISSUES.md
+++ b/COMMON_ISSUES.md
@@ -8,6 +8,42 @@ known or a workaround is available. Please also take a look at the
 release. If your issue hasn't been reported, consider submitting
 [a new issue](https://github.com/twilio/twilio-video.js/issues/new).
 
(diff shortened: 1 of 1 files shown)


COMMON_ISSUES.md
```diff
@@ -8,6 +8,42 @@ known or a workaround is available. Please also take a look at the
 release. If your issue hasn't been reported, consider submitting
 [a new issue](https://github.com/twilio/twilio-video.js/issues/new).
 
+Working around the browsers' autoplay policy
```
syerrapragada: browsers' -> browser's

manjeshbhargav: @syerrapragada `browsers'` signifies more than one browser, so this is the correct way to use an apostrophe.

PR #591, opened 2019-03-28
JSDK-2312, JSDK-2303

manjeshbhargav:
@syerrapragada 

* JSDK-2312 - Support unified plan on Chrome 72+
* JSDK-2303 - Convert audioLevel range to 0-32767

PR #592, opened 2019-03-29
JSDK-2312, JSDK-2303

manjeshbhargav:
@syerrapragada 

* JSDK-2312 - Support unified plan on Chrome 72+
* JSDK-2303 - Convert audioLevel range to 0-32767

I was not fully happy with the RTCRtpTransceiver recycling logic in #591 because I was searching for a recycled RTCRtpTransceiver when adding a MediaStreamTrack to the RTCPeerConnection, which can potentially happen during the middle of a negotiation. In this case, there is a chance that the selected RTCRtpTransceiver might become inactive after the negotiation is complete.

In this PR, I maintain a list of recycled RTCRtpTransceivers, which is updated only after a negotiation is complete. I've had lesser flakiness in integration tests with this approach. Please take a look when you can.


lib/signaling/v2/peerconnection.js
```diff
@@ -299,6 +309,26 @@ class PeerConnectionV2 extends StateMachine {
...
+  /**
+   * Add a new RTCRtpTransceiver or update an existing RTCRtpTransceiver for the
+   * given MediaStreamTrack.
+   * @private
+   * @param {MediaStreamTrack} track
+   * @returns {RTCRtpTransceiver}
+   */
+  _addOrUpdateTransceiver(track) {
+    const transceiver = this._recycledTransceivers[track.kind].shift();
+    if (transceiver && transceiver.sender) {
```
syerrapragada: Should we check for another recycled transceiver in case there is no sender?

manjeshbhargav:
@syerrapragada Since VP8 simulcast for Safari is now released and `master` is available, I will re-open this PR against `master`.

PR #586, opened 2019-03-22
JSDK-2255: Implement different levels of network quality score reports for local and remote participants. 

syerrapragada:
This PR attempts to implement different levels of network quality score reports for local and remote participants. 

@manjeshbhargav Added integration test.

PR #603, opened 2019-04-10
JSDK-2255: Implement different levels of network quality score reports for local and remote participants.

manjeshbhargav:
@syerrapragada 

Reopening #586 against `master-chrome-planb` since `master` has Chrome unified plan changes.


CHANGELOG.md
```diff
@@ -1,5 +1,53 @@
...
+=========================
+
+New Features
+------------
+
+- You can now use the [Network Quality](https://www.twilio.com/docs/video/using-network-quality-api)
+  API to control the verbosity of the network quality statistics that are reported
+  for the LocalParticipant and RemoteParticipants in a Group Room. A Participant will
+  now have an additional property `networkQualityStats` which contains the network
+  quality statistics used to calculate the `networkQualityLevel`. (JSDK-2255)
```
aymenn: - Maybe [Network Quality API](https://www.twilio.com/docs/video/using-network-quality-api)
- It almost seems there are two unrelated things here: NQ API to control verbosity and an addition property on Participant. Are they related?
- Also, for new features, I think it's worth to introduce a section stating what this could be used for. e.g. With this API, you can create an network status overlay or alert the user of long lasting network issues causing a break in communication"

CHANGELOG.md
```diff
@@ -1,5 +1,53 @@
...
+ 
+  ```js
+  const { connect } = require('twilio-video');
+  const room = await connect(token, {
+    networkQuality: {
+      local: 1, // Verbosity level for LocalParticipant [1 - 3]
```
aymenn: Should we have these as "localVerbosityLevel" and "remoteVerbosityLevel"?

lib/connect.js
```diff
@@ -375,8 +393,38 @@ const LogLevel = {
...
+   */
+  minimal: 1,
+  /**
+   * Reports {@link NetworkQualityLevel} and {@link NetworkQualityStats} for the {@link Participant}.
+   * {@link NetworkQualityStats} is populated with audio and video {@link NetworkQualityLevel}s
+   * based on which the {@link Participant}'s {@link NetworkQualityLevel} is calculated.
+   */
+  moderate: 2,
+  /**
+   * Reports {@link NetworkQualityLevel} and {@link NetworkQualityStats} for the participant.
```
aymenn: Maybe replace participant with "{@link Participant}"

Could also say something like, "in addition to the reported statistics reported for minimal, the corresponding {@link NetworkQualityMediaStats} is also reported.

lib/connect.js
```diff
@@ -375,8 +393,38 @@ const LogLevel = {
...
+   */
+  none: 0,
+  /**
+   * Reports {@link NetworkQualityLevel} for the {@link Participant}.
+   */
+  minimal: 1,
+  /**
+   * Reports {@link NetworkQualityLevel} and {@link NetworkQualityStats} for the {@link Participant}.
+   * {@link NetworkQualityStats} is populated with audio and video {@link NetworkQualityLevel}s
+   * based on which the {@link Participant}'s {@link NetworkQualityLevel} is calculated.
```
aymenn: Do we need to mention "based on which the {@link Participant}'s {@link NetworkQualityLevel} is calculated"?
Ive seen that in a few places. What's the purpose?

lib/stats/networkqualitylatencystats.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+/**
+ * Latency network quality statistics.
+ * @property {?number} jitter - media jitter in seconds
```
aymenn: Just double checking, is it in seconds?

lib/stats/networkqualitylatencystats.js
```diff
@@ -0,0 +1,32 @@
...
+ * @property {?number} rtt - round trip time in seconds
```
aymenn: And this one also, is it in seconds?

CHANGELOG.md
manjeshbhargav: @aymenn ,

> It almost seems there are two unrelated things here: NQ API to control verbosity and an addition property on Participant. Are they related?

Yes. By introducing verbosity levels, we are also introducing `networkQualityStats` which houses the statistics based on which the NQ level is calculated.

> Also, for new features, I think it's worth to introduce a section stating what this could be used for. e.g. With this API, you can create an network status overlay or alert the user of long lasting network issues causing a break in communication"

I agree, although I think the wagtail tutorial page for Network Quality (which already exists) is the place to do it.

lib/connect.js
manjeshbhargav: @aymenn ,

Yes. We need to clarify that the stats are the basis for our calculation of the NQ level.

lib/stats/networkqualitylatencystats.js
manjeshbhargav: @aymenn Yes.

lib/stats/networkqualitylatencystats.js
manjeshbhargav: @aymenn Yes.

CHANGELOG.md
manjeshbhargav: @aymenn ,

The names are formalized in the [IDL schema](https://code.hq.twilio.com/client/video-sdk-api/blob/master/idl/networkqualityconfiguration.md).

lib/connect.js
manjeshbhargav: @aymenn ,

We already say this with the following phrase:

> Reports **{@link NetworkQualityLevel}** and {@link NetworkQualityStats}

CHANGELOG.md
aymenn: Should we reference that tutorial here?

lib/connect.js
aymenn: Thanks @manjeshbhargav. I really meant more, why do we need to clarify it? and why do we need to clarify it so may times?

manjeshbhargav: @aymenn ,

The idea is that there are different paths that customers can use to navigate the docs, and having these kinds of clarifications at multiple places makes sure that nobody misses them.
Does that make sense?

aymenn: Ok I think in the future we dont need to. This for example should just be mentioned where we document networkQualityLevel. Mentioning at every occurrence adds little value. i.e. less is more.

PR #629, opened 2019-05-06
JSDK-2338: add region in connectOptions 

makarandp0:
@manjeshbhargav 

This change adds a `region` to connectOptions. The  `region` specified would be used for connecting to signaling server.  

TODO:
- [x] make and verify region change
- [x] add unit tests
- [x] add integration tests
- [x] fix the link for available regions  
- [x] add region option in the signaling-sample app [https://code.hq.twilio.com/client/simpler-signaling.appspot.com/pull/33]


lib/connect.js
```diff
@@ -303,6 +304,9 @@ function connect(token, options) {
  *   for the {@link LocalParticipant} to {@link NetworkQualityVerbosity}<code style="padding:0 0">#minimal</code>
  *   and the {@link NetworkQualityVerbosity} for {@link RemoteParticipant}s to
  *   {@link NetworkQualityVerbosity}<code style="padding:0 0">#none</code>.
+ * @property {?string} [region='gll'] - Preferred region. This region will be used
```
manjeshbhargav: @makarandp0 Two things here:
1. Can you convert `{?string} to `{string}`? A "?" implies that the argument is nullable.

2. Can we make the description closer to the spec:
```
- Preferred region; By default, the client will connect to the nearest signaling server
determined by latency based routing. Setting a value other than <code style="padding:0 0">gll</code>
bypasses routing and guarantees that signaling traffic will be terminated in the region that you prefer;
If you are connecting to a Group Room created with the <code style="padding:0 0">gll</code>
Media Region (either Ad-Hoc or via the REST API), then the Room's Media Region will be selected
based upon your Client's region
```

test/unit/spec/util/constants.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable new-cap */
+'use strict';
+
+const assert = require('assert');
+
+const Constants = require('../../../../lib/util/constants');
+
+describe('Constants', () => {
+  describe('WS_SERVER.', () => {
```
manjeshbhargav: @makarandp0 

IMHO, these tests can be moved to the `connect` unit tests:

*test/unit/spec/connect.js*
```js
describe('connect', () => {
  ...
  describe('called with ConnectOptions#region', () => {
    [/* Values for region */].forEach(region => {
      context(`When region is "${region}"`, () => {
        ...
      });
    });
  });

  describe('called with ConnectOptions#environment', () => {
    [/* Values for environment */].forEach(environment => {
      context(`When environment is "${environment}"`, () => {
        ...
      });
    });
  });
  ...
});
```

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
 
 New Features
 ------------
+- [When connecting to a room](https://www.twilio.com/docs/video/javascript-v2-getting-started#connect-to-a-room),
```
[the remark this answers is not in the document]
makarandp0: thanks, that reads much better :) I will fix.

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
...
+  you can now specify a region that is closest to the participant in ConnectOptions. This region will be used for choosing signaling server for lower latencies. (JSDK-2338).
+
+  ```js
+  const { connect } = require('twilio-video');
+  const room = await connect(token, {
+    region: 'de1'
+  });
+  ```
+  Above option will use signaling servers in Germany. For other possible values for region
```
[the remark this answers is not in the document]
makarandp0: Thanks @manjeshbhargav . I will fix this the text. 
About the link - it mentions more regions than what we have currently ( of those regions only us1, in1 and de1 work right now) - @innerverse - where should we point to for definitive list of regions that developers can use?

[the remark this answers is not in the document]
innerverse: @makarandp0 - We will deploy to every region listed once any SDK goes GA.

lib/connect.js
makarandp0: Thanks. makes sense for both.

test/unit/spec/util/constants.js
makarandp0: I did add a test to connect to make sure that integration between connect and util works. But these unit tests, I think are more useful to be closer to implementation.

makarandp0: now I have added similar tests to connct.js as well.

makarandp0:
Thanks you for your feedback @manjeshbhargav. I have incorporated it and also added integration tests. Can you take another look when you get a chance. Thanks!

test/unit/spec/util/constants.js
manjeshbhargav: @makarandp0 

Since `utils` is mostly used by other modules, we should test them by actually testing the context in which they will be used. That way, when anyone reads the tests, they know that ConnectOptions#region is expected to behave a certain way.

test/integration/spec/connect.js
```diff
@@ -103,6 +104,52 @@ describe('connect', function() {
     });
   });
 
+  describe('region selection ', async () => {
```
manjeshbhargav: @makarandp0 Nitpick: Remove extra space after `selection`?

test/unit/spec/util/constants.js
makarandp0: @manjeshbhargav,  I did take your suggestion and added tests in connect.js My reasoning for keeping the tests in constants.WS_SERVER was the ability to go into details of how the format of the url  looks like for example:

```
    it('maps gll to global for region', () => {
      const wsServer = Constants.WS_SERVER('anyenvironment', 'gll');
      assert.equal(wsServer, 'wss://global.vss.anyenvironment.twilio.com/signaling');
    });
```
where as similar test when written from connect.js need to  be bit vague about the actual format as they do not need to know what Constants.WS_SERVER does interenally. So I had something like:
```
        if (region === 'gll') {
          assert(options.wsServer.indexOf('global') > 0);
        } else {
          assert(options.wsServer.indexOf(region) > 0);
        }
```

test/integration/spec/connect.js
```diff
@@ -103,6 +104,52 @@ describe('connect', function() {
...
+    let token;
+    let cancelablePromise;
+
+    beforeEach(() => {
+      const identity = randomName();
+      token = getToken(identity);
+    });
+
+    context('with an invalid region', async () => {
```
manjeshbhargav: @makarandp0 You can simplify the cases here and re-use some of the test code:
```js
['without', 'global', 'specific', 'invalid'].forEach(regionType => {
  const regionOptions = {
    global: { region: 'gll' },
    invalid: { region: 'foo' },
    specific: { region: 'de1' },
    without: {}
  }[regionType];

  context(`when called ${regionType === 'without a' ? regionType : `with a(n) ${regionType}`} region`, () => {
    beforeEach(() => {
      cancelablePromise = connect(token, Object.assign({}, defaults, regionOptions, { tracks: [] }));
    });

    if (regionType === 'invalid') {
      it('should return a CancelablePromise that rejects with a SignalingConnectionError', async () => {
        ...
      });
    } else {
      it('should return a CancelablePromise that resolves with a Room', async () => {
        ...
      });
    }
  });
});
```

test/integration/spec/connect.js
```diff
@@ -103,6 +104,52 @@ describe('connect', function() {
...
+      const region = 'fictional';
+      it('should return a CancelablePromise that rejects with a SignalingConnectionError', async () => {
+        cancelablePromise = connect(token, Object.assign({}, defaults, { region, tracks: [] }));
+        try {
+          const room = await cancelablePromise;
+          room.disconnect();
+          throw new Error(`Connected to ${room.sid} with an invalid LogLevel`);
```
manjeshbhargav: @makarandp0 "with an invalid ~~LogLevel~~ region"?

test/integration/spec/connect.js
```diff
@@ -103,6 +104,52 @@ describe('connect', function() {
...
+        } catch (error) {
+          assert(error instanceof SignalingConnectionError);
+          assert(/Signaling connection error/.test(error.message));
```
manjeshbhargav: @makarandp0 Testing error message is not necessary.

test/unit/spec/connect.js
```diff
@@ -36,17 +36,66 @@ describe('connect', () => {
...
+        const signaling = sinon.spy(() => mockSignaling);
+
+        connect(token, {
+          iceServers: [],
+          signaling,
+          region
+        });
+
+        const options = signaling.args[0][1];
+        assert.equal(options.wsServer, WS_SERVER(options.environment, options.region));
```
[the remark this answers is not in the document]
makarandp0: I was debating about this when I added tests for `constants.WS_SERVER` in separate file. One option (a) was to let tests in connect.js know about implementation of WS_SERVER (like you are suggesting) or b)  add specific tests into constants.WS_SERVER. 
I chose b) but I am don't feel very strongly about it. I will delete the constants.js tests and add  the logic here.

test/integration/spec/connect.js
makarandp0: Thanks for the suggestion, that does look more compact.

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
...
+  [latency based routing](https://www.twilio.com/docs/video/ip-address-whitelisting#signaling-communication).
+  You can now override this behavior by using a new ConnectOptions flag called `region`. This will make
+  sure that your signaling traffic will terminate at the specified region. (JSDK-2338)
+
+  ```js
+  const { connect } = require('twilio-video');
+  const room = await connect(token, {
+    region: 'de1'
+  });
+  ```
```
manjeshbhargav: @makarandp0 Nitpick: Can you add a new line here?

makarandp0: @aymenn can you take a look at this changelog entry please, and suggest any updates.

ceaglest: Suggestion: "terminate at the" -> "terminate in the"

Overall, this sounds good to me, specifically leaving out mention of "gll" media behavior which is not ready yet. We will need to update the white listing guide to mention the us1, de1, and in1 VSS deployments sometime soon.

makarandp0: updated per your suggestion @ceaglest.

PR #641, opened 2019-05-14
JSDK 2337: media monitor

makarandp0:
This change adds Ice media monitor - to detect the disconnects, and start reconnect early. This is mostly based on corresponding CPP work in https://github.com/twilio/rtc-cpp/pull/640

Change:
1) added a new class IceMediaMonitor `lib/signaling/v2/icemediamonitor.js` that queries the peerConnection for media stats, and calls back when it detects inactivity. (inactivity is defined as no bytes received in 3 seconds)
2) PeerConnection (lib/signaling/v2/peerconnection.js) starts media monitor when ice state is `connected`, and stops it when ice state is `closed` or `failed`. 

Checklist before merge
- [x] ~merge https://github.com/twilio/twilio-webrtc.js/pull/102, and update this change to use `timestamp`~ n/a
- [x] remove debug console.log calls, added in the change 
- [x] resolve and incorporate questions identified in https://paper.dropbox.com/doc/Ice-Media-Monitor-for-JS--AdD_RBac6xRW8YBvKcLizZ68AQ-z2sUflMxe4dXBrjfyWl42
- [x] add more unit tests to verify Monitor callbacks
- [ ] test for network change on chrome/firefox and safari

### DIFF
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const { ICE_ACTIVITY_CHECK_PERIOD_MS, ICE_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
-const { getStats } = require('@twilio/webrtc');
 
 /**
  * Monitors a {@link RTCPeerConnection}'s media stats and notifies
@@ -15,23 +14,23 @@ class IceConnectionMonitor {
    */
   constructor(peerConnection, options) {
     options = Object.assign({
-      activityCheckPeriodMS: ICE_ACTIVITY_CHECK_PERIOD_MS,
-      inactivityThresholdMS: ICE_INACTIVITY_THRESHOLD_MS,
+      activityCheckPeriodMs: ICE_ACTIVITY_CHECK_PERIOD_MS,
+      inactivityThresholdMs: ICE_INACTIVITY_THRESHOLD_MS,
     }, options);
 
     Object.defineProperties(this, {
+      _activityCheckPeriodMs: {
+        value: options.activityCheckPeriodMs
+      },
+      _inactivityThresholdMs: {
+        value: options.inactivityThresholdMs
+      },
       _peerConnection: {
         value: peerConnection
       },
       _timer: {
         value: null,
         writable: true,
-      },
-      _activityCheckPeriodMS: {
-        value: options.activityCheckPeriodMS
-      },
-      _inactivityThresholdMS: {
-        value: options.inactivityThresholdMS
       }
     });
   }
@@ -43,26 +42,35 @@ class IceConnectionMonitor {
     console.log.apply(console.log, newArgs.concat(args));
   }
 
-  // TODO: simplify once timestamp is added to the webrtc getStat api.
+  /**
+   * Get ICE connection stats.
+   * @returns Promise<{?RTCIceCandidatePairStats}>
+   */
+  _getIceConnectionStats() {
+    return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
+      return stat.type === 'candidate-pair' && stat.nominated;
+    }));
+  }
+
+  /**
+   * returns media stats or null.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
+   */
   _getMediaStats() {
-    // returns stats needed for media monotor.
-    return getStats(this._peerConnection).then((response) => {
-      // TODO: update twilio-webRTC to contain timestamp in activeIceCandidatePair
-      if (response.activeIceCandidatePair) {
-        return this._peerConnection.getStats().then((stats) => {
-          var realActiveCandidatePairStats = Array.from(stats.values()).find(function(stat) {
-            return stat.type === 'candidate-pair' && stat.nominated;
-          });
-          return {
-            timestamp: realActiveCandidatePairStats.timestamp,
-            bytesReceived: response.activeIceCandidatePair.bytesReceived,
-            bytesSent: response.activeIceCandidatePair.bytesSent
-          };
-        });
+    return this._getIceConnectionStats().then((activeCandidatePairStats) => {
+      if (!activeCandidatePairStats) {
+        return null;
       }
+      return {
+        timestamp: activeCandidatePairStats.timestamp,
+        bytesReceived: activeCandidatePairStats.bytesReceived,
+        bytesSent: activeCandidatePairStats.bytesSent
+      };
+    }).catch(() => {
       return null;
     });
   }
+
   /**
    * Start monitoring.
    * @returns {void}
@@ -92,13 +100,13 @@ class IceConnectionMonitor {
         }
 
         // after inactivity threshold fire inactiveCallback.
-        if (mediaStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMS) {
+        if (mediaStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
           onInactiveCallback();
         }
       }).catch((err) => {
         this._makarandlog('error in _restartIceIfInactiveForLong', err);
       });
-    }, this._activityCheckPeriodMS);
+    }, this._activityCheckPeriodMs);
   }
 
   /**
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -1,23 +1,27 @@
 'use strict';
 
-const { ICE_MONITOR_ACTIVITY_CHECK_PERIOD_MS, ICE_MONITOR_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
+const { ICE_ACTIVITY_CHECK_PERIOD_MS, ICE_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
+const { getStats } = require('@twilio/webrtc');
 
 /**
- * @emits IceMediaMonitor#updated
+ * Monitors a {@link RTCPeerConnection}'s media stats and notifies
+ * caller when inactivity is detected.
  */
-class IceMediaMonitor {
+class IceConnectionMonitor {
   /**
-   * Construct a {@link IceMediaMonitor}.
+   * Construct an {@link IceConnectionMonitor}.
+   * @param {RTCPeerConnection} peerConnection
+   * @param {object} [options]
    */
-  constructor(pc, options) {
+  constructor(peerConnection, options) {
     options = Object.assign({
-      activityCheckPeriodMS: ICE_MONITOR_ACTIVITY_CHECK_PERIOD_MS,
-      inactivityThresholdMS: ICE_MONITOR_INACTIVITY_THRESHOLD_MS,
+      activityCheckPeriodMS: ICE_ACTIVITY_CHECK_PERIOD_MS,
+      inactivityThresholdMS: ICE_INACTIVITY_THRESHOLD_MS,
     }, options);
 
     Object.defineProperties(this, {
-      _pc: {
-        value: pc
+      _peerConnection: {
+        value: peerConnection
       },
       _timer: {
         value: null,
@@ -39,8 +43,25 @@ class IceMediaMonitor {
     console.log.apply(console.log, newArgs.concat(args));
   }
 
+  // TODO: simplify once timestamp is added to the webrtc getStat api.
   _getMediaStats() {
-    return this._pc.getMediaStats();
+    // returns stats needed for media monotor.
+    return getStats(this._peerConnection).then((response) => {
+      // TODO: update twilio-webRTC to contain timestamp in activeIceCandidatePair
+      if (response.activeIceCandidatePair) {
+        return this._peerConnection.getStats().then((stats) => {
+          var realActiveCandidatePairStats = Array.from(stats.values()).find(function(stat) {
+            return stat.type === 'candidate-pair' && stat.nominated;
+          });
+          return {
+            timestamp: realActiveCandidatePairStats.timestamp,
+            bytesReceived: response.activeIceCandidatePair.bytesReceived,
+            bytesSent: response.activeIceCandidatePair.bytesSent
+          };
+        });
+      }
+      return null;
+    });
   }
   /**
    * Start monitoring.
@@ -93,4 +114,4 @@ class IceMediaMonitor {
   }
 }
 
-module.exports = IceMediaMonitor;
+module.exports = IceConnectionMonitor;
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -80,7 +80,6 @@ class IceConnectionMonitor {
    * @returns {void}
    */
   start(onIceConnectionInactive) {
-    console.trace("makarand: start was called!");
     this._makarandlog('starting MediaTimer');
     this.stop();
     this._timer = setInterval(() => {
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -46,13 +46,7 @@ class IceConnectionMonitor {
   _getIceConnectionStats() {
     return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
       return stat.type === 'candidate-pair' && stat.nominated;
-    })).then((activeCandidatePairStats) => {
-      return {
-        timestamp: activeCandidatePairStats.timestamp,
-        bytesReceived: activeCandidatePairStats.bytesReceived,
-        bytesSent: activeCandidatePairStats.bytesSent
-      };
-    }).catch(() => {
+    })).catch(() => {
       return null;
     });
   }
@@ -85,7 +79,7 @@ class IceConnectionMonitor {
   }
 
   /**
-   * Stop monitoring the ICE connection state
+   * Stop monitoring the ICE connection state.
    * @returns {void}
    */
   stop() {
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -25,6 +25,10 @@ class IceConnectionMonitor {
       _inactivityThresholdMs: {
         value: options.inactivityThresholdMs
       },
+      _lastActiveStats: {
+        value: null,
+        writable: true
+      },
       _peerConnection: {
         value: peerConnection
       },
@@ -118,6 +122,7 @@ class IceConnectionMonitor {
       this._makarandlog('stopping MediaTimer');
       clearInterval(this._timer);
       this._timer = null;
+      this._lastActivity = null;
     }
   }
 }
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -57,10 +57,10 @@ class IceConnectionMonitor {
   }
 
   /**
-   * returns media stats or null.
+   * returns ICE connection stats or null.
    * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
    */
-  _getMediaStats() {
+  _getStats() {
     return this._getIceConnectionStats().then((activeCandidatePairStats) => {
       if (!activeCandidatePairStats) {
         return null;
@@ -79,33 +79,32 @@ class IceConnectionMonitor {
    * Start monitoring.
    * @returns {void}
    */
-  start(onInactiveCallback) {
-    if (typeof onInactiveCallback !== 'function') throw new Error('invalid callback');
-
+  start(onIceConnectionInactive) {
+    console.trace("makarand: start was called!");
     this._makarandlog('starting MediaTimer');
     this.stop();
     this._timer = setInterval(() => {
-      this._getMediaStats().then(mediaStats => {
-        if (!mediaStats) {
+      this._getStats().then(iceStats => {
+        if (!iceStats) {
           return;
         }
 
         if (this._lastActivity ) {
           this._makarandlog('Diff: ', {
-            time: mediaStats.timestamp - this._lastActivity.timestamp,
-            received: mediaStats.bytesReceived - this._lastActivity.bytesReceived,
-            sent: mediaStats.bytesSent - this._lastActivity.bytesSent
+            time: iceStats.timestamp - this._lastActivity.timestamp,
+            received: iceStats.bytesReceived - this._lastActivity.bytesReceived,
+            sent: iceStats.bytesSent - this._lastActivity.bytesSent
           });
         }
 
         // if change detected in bytesReceived mark it as lastActiveiy.
-        if (!this._lastActivity || this._lastActivity.bytesReceived !== mediaStats.bytesReceived) {
-          this._lastActivity = mediaStats;
+        if (!this._lastActivity || this._lastActivity.bytesReceived !== iceStats.bytesReceived) {
+          this._lastActivity = iceStats;
         }
 
         // after inactivity threshold fire inactiveCallback.
-        if (mediaStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
-          onInactiveCallback();
+        if (iceStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
+          onIceConnectionInactive();
         }
       }).catch((err) => {
         this._makarandlog('error in _restartIceIfInactiveForLong', err);
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -3,7 +3,7 @@
 const { ICE_ACTIVITY_CHECK_PERIOD_MS, ICE_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
 
 /**
- * Monitors a {@link RTCPeerConnection}'s media stats and notifies
+ * Monitors a {@link RTCPeerConnection}'s stats and notifies
  * caller when inactivity is detected.
  */
 class IceConnectionMonitor {
@@ -39,32 +39,14 @@ class IceConnectionMonitor {
     });
   }
 
-  // TODO: remove this before merging.
-  _makarandlog() {
-    var newArgs = ['makarand: '];
-    var args = Array.prototype.slice.call(arguments);
-    console.log.apply(console.log, newArgs.concat(args));
-  }
-
   /**
-   * Get ICE connection stats.
-   * @returns Promise<{?RTCIceCandidatePairStats}>
+   * Get ICE connection stats, and extract received and send bytes.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
    */
   _getIceConnectionStats() {
     return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
       return stat.type === 'candidate-pair' && stat.nominated;
-    }));
-  }
-
-  /**
-   * returns ICE connection stats or null.
-   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
-   */
-  _getStats() {
-    return this._getIceConnectionStats().then((activeCandidatePairStats) => {
-      if (!activeCandidatePairStats) {
-        return null;
-      }
+    })).then((activeCandidatePairStats) => {
       return {
         timestamp: activeCandidatePairStats.timestamp,
         bytesReceived: activeCandidatePairStats.bytesReceived,
@@ -76,24 +58,17 @@ class IceConnectionMonitor {
   }
 
   /**
-   * Start monitoring.
-   * @returns {void}
+   * Start monitoring the ICE connection.
+   * Monitors bytes received on active ice connection pair,
+   * invokes onIceConnectionInactive when inactivity is detected.
+   * @param {function} onIceConnectionInactive
    */
   start(onIceConnectionInactive) {
-    this._makarandlog('starting MediaTimer');
     this.stop();
     this._timer = setInterval(() => {
-      this._getStats().then(iceStats => {
+      this._getIceConnectionStats().then(iceStats => {
         if (!iceStats) {
           return;
-        }
-
-        if (this._lastActivity ) {
-          this._makarandlog('Diff: ', {
-            time: iceStats.timestamp - this._lastActivity.timestamp,
-            received: iceStats.bytesReceived - this._lastActivity.bytesReceived,
-            sent: iceStats.bytesSent - this._lastActivity.bytesSent
-          });
         }
 
         // if change detected in bytesReceived mark it as lastActiveiy.
@@ -105,19 +80,16 @@ class IceConnectionMonitor {
         if (iceStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
           onIceConnectionInactive();
         }
-      }).catch((err) => {
-        this._makarandlog('error in _restartIceIfInactiveForLong', err);
       });
     }, this._activityCheckPeriodMs);
   }
 
   /**
-   * Stop monitoring.
+   * Stop monitoring the ICE connection state
    * @returns {void}
    */
   stop() {
     if (this._timer !== null) {
-      this._makarandlog('stopping MediaTimer');
       clearInterval(this._timer);
       this._timer = null;
       this._lastActivity = null;
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -41,7 +41,7 @@ class IceConnectionMonitor {
 
   /**
    * Get ICE connection stats, and extract received and send bytes.
-   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
+   * @returns Promise<?RTCIceCandidatePairStats>
    */
   _getIceConnectionStats() {
     return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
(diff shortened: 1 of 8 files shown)


lib/signaling/v2/icemediamonitor.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  _getMediaStats() {
+    return this._pc.getMediaStats();
+  }
+  /**
+   * Start monitoring.
+   * @returns {void}
+   */
+  start(onInactiveCallback) {
```
makarandp0: I initially had  this class as EventEmitter (instead of taking callback param here), but decided against that, as callback works better in this case because 1) we have just one listener 2) the lifetime of callback is between start and stop, and it's easier for caller to call start/stop instead of an additional on/off for event listener.

test/unit/spec/signaling/v2/icemediamonitor.js
```diff
@@ -0,0 +1,94 @@
...
+      const monitor = new IceMediaMonitor(pc);
+      assert.equal(monitor._timer, null);
+      monitor.start(() => {});
+      assert.notEqual(monitor._timer, null);
+      monitor.stop();
+      assert.equal(monitor._timer, null);
+    });
+  });
+
+  describe('Callback', () => {
```
makarandp0: WIP

lib/signaling/v2/icemediamonitor.js
manjeshbhargav: @makarandp0 I think you can convert this to a Promise-based method and get rid of the `stop()` method 
 and the `_timer` property:
```js
class IceConnectionMonitor {
  ...
  /**
   * Check if the ICE connection is inactive.
   * @returns {Promise<void>}
   */
  checkIfInactive() {
    return wait(this._activityCheckPeriodMs).then(() => {
      return this._getIceConnectionStats();
    }).then(stats => {
      if (!stats) {
        return this.checkIfInactive();
      }
      if (!this._lastActiveStats || this._lastActiveStats.bytesReceived !== stats.bytesReceived) {
        this._lastActiveStats = stats;
      }
      if (stats.timestamp - this._lastActivity.timestamp < this._inactivityThresholdMs) {
        return this.checkIfInactive();
      }
    });
  }
  ...
}

// This can be moved to lib/utils/index.js

/**
 * Wait for a given period of time.
 * @param {number} delayMs
 * @returns {Promise<void>}
 */
function wait(delayMs) {
  return new Promise(resolve => setTimeout(resolve, delayMs));
}
```

lib/signaling/v2/icemediamonitor.js
```diff
@@ -0,0 +1,96 @@
...
+      _timer: {
+        value: null,
+        writable: true,
+      },
+      _activityCheckPeriodMS: {
+        value: options.activityCheckPeriodMS
+      },
+      _inactivityThresholdMS: {
+        value: options.inactivityThresholdMS
+      }
```
manjeshbhargav: @makarandp0 Can you define `_lastActiveStats` here?

lib/signaling/v2/peerconnection.js
```diff
@@ -185,6 +187,13 @@ class PeerConnectionV2 extends StateMachine {
       _rtpSenders: {
         value: new Map()
       },
+      _mediaMonitorTimer: {
```
manjeshbhargav: @makarandp0 Looks like this is not used anywhere.

lib/signaling/v2/peerconnection.js
```diff
@@ -99,6 +100,7 @@ class PeerConnectionV2 extends StateMachine {
       setBitrateParameters,
       setCodecPreferences,
       setSimulcast,
+      IceMediaMonitor,
```
manjeshbhargav: @makarandp0 `IceConnectionMonitor`?

lib/signaling/v2/peerconnection.js
```diff
@@ -7,6 +7,7 @@ const DefaultMediaStream = WebRTC.MediaStream;
 const DefaultRTCIceCandidate = WebRTC.RTCIceCandidate;
 const DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
 const DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
+const IceMediaMonitor = require('./icemediamonitor.js');
```
manjeshbhargav: @makarandp0 `iceconnectionmonitor.js`?

lib/signaling/v2/peerconnection.js
```diff
@@ -185,6 +187,13 @@ class PeerConnectionV2 extends StateMachine {
       _rtpSenders: {
         value: new Map()
       },
+      _mediaMonitorTimer: {
+        writable: true,
+        value: null
+      },
+      _mediaMonitor: {
```
manjeshbhargav: @makarandp0 Can you change this to:
```js
_iceConnectionMonitor: {
  value: new options.IceConnectionMonitor(peerConnection)
}
```
?

lib/signaling/v2/peerconnection.js
```diff
@@ -474,10 +483,15 @@ class PeerConnectionV2 extends StateMachine {
   _handleIceConnectionStateChange() {
     const { iceConnectionState } = this._peerConnection;
 
-    this._log.debug(`ICE connection state is "${iceConnectionState}"`);
+    this._log.debug(`makarand: ICE connection state is "${this._lastIceConnectionState}" => "${iceConnectionState}"`);
+    // makarand: TODO: is it okay to short-circuit here? or anybody depends on
+    // iceConnectionStateChanged firing on every non-change
+    if (this._lastIceConnectionState === iceConnectionState) {
```
manjeshbhargav: @makarandp0 Let's retain how it was before, unless you can demonstrate that "iceConnectionStateChanged" is emitted even when there is a non-change.

lib/signaling/v2/peerconnection.js
```diff
@@ -488,10 +502,46 @@ class PeerConnectionV2 extends StateMachine {
       this._log.info('ICE reconnected');
     }
 
+    if (iceConnectionState === 'connected') {
+      this._mediaMonitor.start(() => {
```
manjeshbhargav: @makarandp0 This can be:
```js
this._iceConnectionMonitor.checkIfInactive().then(() => {
  if (this._lastIceConnectionState !== 'disconnected') {
    return this._iceConnectionMonitor.checkIfInactive();
  }
  if (!this._shouldRestartIce && !this._isRestartingIce) {
    this._shouldRestartIce = true;
    this.offer();
  }
});
```

lib/signaling/v2/peerconnection.js
```diff
@@ -488,10 +502,46 @@ class PeerConnectionV2 extends StateMachine {
...
+        this._log.debug(`makarand: Media Monitor detected inactivity while in "${this._lastIceConnectionState}" state`);
+        if (this._lastIceConnectionState === 'disconnected') {
+          this._mediaMonitor.stop();
+          if (!this._shouldRestartIce && !this._isRestartingIce) {
+            this._shouldRestartIce = true;
+            this.offer();
+          }
+        }
+      });
+    } else if (iceConnectionState === 'failed' || iceConnectionState === 'closed') {
```
manjeshbhargav: @makarandp0 I don't think this will be necessary with the new way of checking inactivity, since we are using `setTimeout` instead of `setInterval`.

lib/signaling/v2/peerconnection.js
```diff
@@ -488,10 +502,46 @@ class PeerConnectionV2 extends StateMachine {
...
+      this._mediaMonitor.stop();
+    }
+
     this._lastIceConnectionState = iceConnectionState;
     this.emit('iceConnectionStateChanged');
   }
 
+  // TODO: simplify and move to IceMediaMonitor once timestamp is added to the webrtc getStat api.
+  getMediaStats() {
```
manjeshbhargav: @makarandp0 This method won't be necessary due to availability of `IceConnectionMonitor#_getIceConnectionStats()`.

lib/util/constants.js
```diff
@@ -71,3 +71,6 @@ module.exports.typeErrors = {
 module.exports.DEFAULT_NQ_LEVEL_LOCAL = 1;
 module.exports.DEFAULT_NQ_LEVEL_REMOTE = 0;
 module.exports.MAX_NQ_LEVEL = 3;
+
+module.exports.ICE_MONITOR_ACTIVITY_CHECK_PERIOD_MS = 1000;
```
manjeshbhargav: @makarandp0 Nitpick: I think we can shorten this to `ICE_ACTIVITY_CHECK_PERIOD_MS`.

lib/util/constants.js
```diff
@@ -71,3 +71,6 @@ module.exports.typeErrors = {
...
+module.exports.ICE_MONITOR_INACTIVITY_THRESHOLD_MS = 3000;
```
manjeshbhargav: @makarandp0 Nitpick: I think we can shorten this to `ICE_INACTIVITY_THRESHOLD_MS`.

test/unit/index.js
```diff
@@ -49,6 +49,7 @@ require('./spec/signaling/v2/localparticipant');
 require('./spec/signaling/v2/networkqualitymonitor');
 require('./spec/signaling/v2/networkqualitysignaling');
 require('./spec/signaling/v2/recording');
+require('./spec/signaling/v2/icemediamonitor');
```
manjeshbhargav: @makarandp0 `iceconnectionmonitor.js`.

lib/signaling/v2/icemediamonitor.js
makarandp0: @manjeshbhargav, I think your suggestion would work if we decided to only start the monitor when Ice goes into disconnected state, and simply wait for 3 seconds of non-activity. But since we are monitoring at connected state, we need this callback to fire every time it detects inactivity, and peerConnection object acts on it only when it finds itself in disconnected state when callback fires. Maybe I can setup a quick meeting to syncup on this.

lib/signaling/v2/icemediamonitor.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const { ICE_MONITOR_ACTIVITY_CHECK_PERIOD_MS, ICE_MONITOR_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
+
+/**
```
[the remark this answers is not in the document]
makarandp0: okay, will add line describing the class
 ```
  Monitors a {@link RTCPeerConnection}'s ICE media stats and notifies
 caller when inactivity is detected.
```
not mentioning the ICE connection state, as that logic is in PeerConnectionV2

File: lib/signaling/v2/icemediamonitor.js [hunk 16]
```diff
@@ -0,0 +1,96 @@
...
+ * @emits IceMediaMonitor#updated
+ */
+class IceMediaMonitor {
+  /**
+   * Construct a {@link IceMediaMonitor}.
+   */
+  constructor(pc, options) {
```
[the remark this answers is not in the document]
makarandp0: yes, that makes sense. I will update to use RTCPeerConnection, and also move the getMediaStats over to this file.

lib/signaling/v2/icemediamonitor.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  // TODO: remove this before merging.
+  _makarandlog() {
+    var newArgs = ['makarand: '];
+    var args = Array.prototype.slice.call(arguments);
+    console.log.apply(console.log, newArgs.concat(args));
+  }
+
+  _getMediaStats() {
```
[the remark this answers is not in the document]
makarandp0: Although it may not be necessary in this case - I thought using twilio-webrtc was preferred because it ensures standardized response across browsers. But I am open to using peerConnection directly.

lib/signaling/v2/peerconnection.js
makarandp0: Looks like the code previously assumed that event could be emitted for non-changes. Notice that https://github.com/twilio/twilio-video.js/blob/master/lib/signaling/v2/peerconnection.js#L494  
explicitely ensures that old state is not "failed" 
```
    // Case 1: Transition to "failed".
    if (this._lastIceConnectionState !== 'failed' && iceConnectionState === 'failed' && !this._shouldRestartIce && !this._isRestartingIce) {
```

This change allows to get rid of that check.

lib/signaling/v2/peerconnection.js
makarandp0: ah, Now I understand what you mean by using promise for this api -  Its using nested promises -  I think there are couple cons with this approach.

1. nested promises are making code bit complicated.
2. this usage makes it difficult to use "previous" state from the media monitor. For example handling of this case is not obvious.
a) connected: time 0 -  bytes = 0 - Api was called and it returned a promise..
b) connected: time 4 -  bytes = 0 - promise resolved. 
c) Api was was called again   because state is still connected.
d) disconnected:
e) disconnected: time 5 - bytes = 0.

At e) should the API resolve the promise? For callback it's obvious that it should fire again (as it was started, and callbacks on every inactivity). But since a new promise was started c) and it hasn't yet been 3 seconds of inactivity (although total inactivity is 5 seconds).


3. Also  with this promise API the promise never gets cleared (as there is no `stop` functionality) That means we will always have an outstanding promise (and associated timer) for peer connection - long after the peer connection is closed. This would lead to memory leaks, if there were lots of peerconnection on the app.

lib/signaling/v2/icemediamonitor.js
makarandp0: never mind my comment above - I think I better understand your proposal now, and how it would work, but I do have few concerns with it, as I commented earlier.

lib/signaling/v2/peerconnection.js
makarandp0: updated this now that we've decided to listen only after disconnect.

lib/signaling/v2/icemediamonitor.js
makarandp0: updated this now that we've decided to listen only after disconnect.

lib/signaling/v2/peerconnection.js
```diff
@@ -488,6 +493,19 @@ class PeerConnectionV2 extends StateMachine {
       this._log.info('ICE reconnected');
     }
 
+    if (iceConnectionState === 'disconnected') {
+      this._iceConnectionMonitor.start(() => {
```
manjeshbhargav: @makarandp0 We need to make sure that when we access `PeerConnectionV2#iceConnectionState`, it is set to 'failed' here, even though the actual state is disconnected. We can do something like this:
```js
this._iceConnectionMonitor.start(() => {
  this._isIceConnectionInactive = true;
  ...
});
```
Then, we can modify the `iceConnectionState` getter for PeerConnectionV2 like this:
```js
get iceConnectionState() {
  return this._peerConnection.iceConnectionState === 'disconnected' && this._isIceConnectionInactive
    ? 'failed'
    : this._peerConnection.iceConnectionState;
}
```

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
+'use strict';
+
+const assert = require('assert');
+const sinon = require('sinon');
+
+const IceConnectionMonitor = require('../../../../../lib/signaling/v2/iceconnectionmonitor');
+
+describe('IceConnectionMonitor', () => {
+
```
manjeshbhargav: @makarandp0 Nitpick: Remove extra line?

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+  describe('.constructor', () => {
```
manjeshbhargav: @makarandp0 Nitpick: Remove `.`?

manjeshbhargav: @makarandp0 I don't think we need to test for internal variables/methods unless absolutely required. Ideally we should just test public facing variables/methods like `start()` and `stop()`.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+    it('defaults to 1sec of check period', () => {
+      assert.equal(new IceConnectionMonitor(pc)._activityCheckPeriodMs, 1000);
+    });
+
+    it('defaults to 3 sec of inactivity threshold', () => {
+      assert.equal(new IceConnectionMonitor(pc)._inactivityThresholdMs, 3000);
+    });
+  });
+
+  describe('.start', () => {
```
manjeshbhargav: @makarandp0 Nitpick: `#start`?

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+    // eslint-disable-next-line no-undefined
+    ['foo', 45, { foo: 5 }, null].forEach((callback) => {
+      it(`throws if callback is ${typeof callback} `, () => {
```
manjeshbhargav: @makarandp0 Not needed.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+    it('starts the timer', () => {
+      var pc = { foo: 1 };
+      const monitor = new IceConnectionMonitor(pc);
+      assert.equal(monitor._timer, null);
+      monitor.start(() => {});
+      assert.notEqual(monitor._timer, null);
+    });
+  });
+
+  describe('._getMediaStats', () => {
```
manjeshbhargav: @makarandp0 Not necessary.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+      it(`throws if callback is ${typeof callback} `, () => {
+        const pc = { foo: 1 };
+        const monitor = new IceConnectionMonitor(pc);
+        assert.throws(() => {
+          monitor.start(callback);
+        });
+      });
+    });
+
+    it('starts the timer', () => {
```
manjeshbhargav: @makarandp0 I think we can test the callback firing here, which is the only public facing behavior. We should not test internal implementation details like starting/stopping the timer.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+      var pc = { foo: 1 };
+      const iceConnectionMonitor = new IceConnectionMonitor(pc);
+      sinon.stub(iceConnectionMonitor, '_getIceConnectionStats').returns(Promise.reject(new Error('boo')));
+      return iceConnectionMonitor._getMediaStats().then((mediaStats) => {
+        assert.strictEqual(mediaStats, null);
+      });
+    });
+  });
+
+  describe('.stop', () => {
```
manjeshbhargav: @makarandp0 Nitpick: `#stop`?

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+    it('stops the timer, and resets the state', () => {
```
manjeshbhargav: @makarandp0 Should not be testing implementation details, only expected behavior.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+      assert.equal(monitor._timer, null);
+      monitor.start(() => {});
+      assert.notEqual(monitor._timer, null);
+      monitor.stop();
+      assert.equal(monitor._timer, null);
+      assert.equal(monitor._lastActivity, null);
+    });
+  });
+
+  describe('on inactivity', () => {
```
manjeshbhargav: @makarandp0 This can move to the `#start` test.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,172 @@
...
+
+      setTimeout(() => {
+        assert(monitor._getMediaStats.callCount < 12, 'timeout should reach before reasonable callCount');
+        done();
+      }, 10);
+    });
+  });
+});
+
+function mockMediaStats(iceConnectionMonitor, mediaStatResults) {
```
manjeshbhargav: @makarandp0 `mockIceConnectionStats` and `statsResults`?

test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -5,12 +5,34 @@ const EventEmitter = require('events');
...
+    });
+    sinon.stub(IceConnectionMonitor.prototype, 'stop').callsFake(() => {
+      didStopMonitor = true;
+    });
+  });
+  afterEach(() => {
+    IceConnectionMonitor.prototype.start.restore();
+    IceConnectionMonitor.prototype.stop.restore();
+  });
+
```
manjeshbhargav: @makarandp0 Can you also add or update the test for `PeerConnectionV2#iceConnectionState` to return "failed" if the actual state is "disconnected" and ICE is inactive, and should return the underlying RTCPeerConnection's `iceConnectionState` otherwise?

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,130 @@
...
+      return null;
+    });
+  }
+
+  /**
+   * Start monitoring.
+   * @returns {void}
+   */
+  start(onInactiveCallback) {
+    if (typeof onInactiveCallback !== 'function') throw new Error('invalid callback');
```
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav, I am big fan of asserts - but did not see usage in our codebase. this looks like something that we could assert for (and not have an elaborate throw like I have here). Do we use asserts in our code? basically some code that is mostly for for helping debug, catch bugs early ?

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,130 @@
...
+    return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
+      return stat.type === 'candidate-pair' && stat.nominated;
+    }));
+  }
+
+  /**
+   * returns media stats or null.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
+   */
+  _getMediaStats() {
```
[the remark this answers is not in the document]
makarandp0: Re: introduction of _getMediaStats - I added it to  help isolate the peerConnection.getStats logic, and thus help write effective unit tests. like "getMediaStats works fine when _getIceConnectionStats throws or returns undefined"

test/unit/spec/signaling/v2/iceconnectionmonitor.js
makarandp0: it's a uni test, and I think it is useful to verify the internal state from unit test. But I agree overall, its better to test public interfaces. Here i did not want to add to the public interface to ensure that the timer is turned off.

lib/signaling/v2/iceconnectionmonitor.js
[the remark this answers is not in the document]
makarandp0: I will remove this check, but wanted to discuss this with you more about.

test/unit/spec/signaling/v2/iceconnectionmonitor.js
makarandp0: q: where do we use '.foo' v/s '#foo' ? didn't see any other usage of '#foo' in the code

test/unit/spec/signaling/v2/peerconnection.js
makarandp0: oh, I did not think about this - Do we need to keep track of the ice failure predicted by monitor and pretend that it was real failure to the callers of PeerConnectionV2 ? Let me sync up with you  understand this better.

makarandp0:
Thank you for the feedback and catching unit test failure issue @manjeshbhargav - I think my tests were leaking a timer which was causing failures during other test runs ). I have fixed the issue, and added some instrumentation to see if it happens again. i will remove the instrumentation before merge along with other debug code.

lib/signaling/v2/peerconnection.js
makarandp0: ah, I missed this comment. makes sense.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
+
+const { ICE_ACTIVITY_CHECK_PERIOD_MS, ICE_INACTIVITY_THRESHOLD_MS } = require('../../util/constants');
+
+/**
+ * Monitors a {@link RTCPeerConnection}'s media stats and notifies
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 "media" is still present.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,128 @@
...
+    return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
+      return stat.type === 'candidate-pair' && stat.nominated;
+    }));
+  }
+
+  /**
+   * returns ICE connection stats or null.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
+   */
+  _getStats() {
```
manjeshbhargav: @makarandp0 I still think we don't need this method. Why would we want to copy the stats values into a JSON instead of just using the RTCIceCandidateStats directly? You can also attach the `catch()` handler to `_getIceConnectionStats()`.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,128 @@
...
+        bytesReceived: activeCandidatePairStats.bytesReceived,
+        bytesSent: activeCandidatePairStats.bytesSent
+      };
+    }).catch(() => {
+      return null;
+    });
+  }
+
+  /**
+   * Start monitoring.
```
manjeshbhargav: @makarandp0 Can you add a more descriptive statement here along with specifying the parameter?
```js
/**
 * Start monitoring the ICE connection.
 * @param {function} onIceConnectionInactive
 * ...
 */
```

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,128 @@
...
+          onIceConnectionInactive();
+        }
+      }).catch((err) => {
+        this._makarandlog('error in _restartIceIfInactiveForLong', err);
+      });
+    }, this._activityCheckPeriodMs);
+  }
+
+  /**
+   * Stop monitoring.
```
manjeshbhargav: @makarandp0 Nitpick: `Stop monitoring the ICE connection state.`?

lib/signaling/v2/peerconnection.js
```diff
@@ -290,7 +299,7 @@ class PeerConnectionV2 extends StateMachine {
    * @property {RTCIceConnectionState}
    */
   get iceConnectionState() {
-    return this._peerConnection.iceConnectionState;
+    return (this._isIceConnectionInactive && this._peerConnection.iceConnectionState === 'disconnected') ? 'failed' : this._peerConnection.iceConnectionState;
```
manjeshbhargav: @makarandp0 Nitpick: can you make this multiline so that its not too long?

lib/signaling/v2/peerconnection.js
```diff
@@ -474,7 +483,7 @@ class PeerConnectionV2 extends StateMachine {
   _handleIceConnectionStateChange() {
     const { iceConnectionState } = this._peerConnection;
 
-    this._log.debug(`ICE connection state is "${iceConnectionState}"`);
+    this._log.debug(`ICE connection state is "${this._lastIceConnectionState}" => "${iceConnectionState}"`);
```
manjeshbhargav: @makarandp0 I think we should not change the log statement here since `_lastIceConnectionState` can be null.

lib/signaling/v2/iceconnectionmonitor.js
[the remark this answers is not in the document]
makarandp0: removed the method now.

lib/signaling/v2/iceconnectionmonitor.js
makarandp0: okay, I guess the method has served its purpose of being able to test. I will removed it and fold into `_getIceConnectionStats`

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Get ICE connection stats, and extract received and send bytes.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
+   */
+  _getIceConnectionStats() {
+    return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
+      return stat.type === 'candidate-pair' && stat.nominated;
+    })).then((activeCandidatePairStats) => {
```
manjeshbhargav: @makarandp0 Nitpick: I think we can return the whole RTCIceCandidatePairStats here and get rid of the `then()` callback. Also, I think you can get rid of the braces around `activeIceCandidatePairStats`.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,100 @@
...
+        // after inactivity threshold fire inactiveCallback.
+        if (iceStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
+          onIceConnectionInactive();
+        }
+      });
+    }, this._activityCheckPeriodMs);
+  }
+
+  /**
+   * Stop monitoring the ICE connection state
```
manjeshbhargav: @makarandp0 Nitpick: Add a `.` at the end?

lib/signaling/v2/peerconnection.js
```diff
@@ -290,7 +299,8 @@ class PeerConnectionV2 extends StateMachine {
    * @property {RTCIceConnectionState}
    */
   get iceConnectionState() {
-    return this._peerConnection.iceConnectionState;
+    return (this._isIceConnectionInactive && this._peerConnection.iceConnectionState === 'disconnected')
```
manjeshbhargav: @makarandp0 Nitpick: I don't think we need the `()` around the condition. Also, can you indent the `?` with just two spaces after `return` like this:
```js
return this._isIceConnectionInactive && this._peerConnection.iceConnectionState === 'disconnected'
  ? 'failed' : this._peerConnection.iceConnectionState;
```
?

lib/signaling/v2/iceconnectionmonitor.js
makarandp0: imo having a limited fields returned from _getIceConnectionStats, helps with unit test/mocking of this function. I will  update  to get rid of braces.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -46,13 +46,7 @@ class IceConnectionMonitor {
...
     return this._peerConnection.getStats().then(stats => Array.from(stats.values()).find(stat => {
       return stat.type === 'candidate-pair' && stat.nominated;
-    })).then((activeCandidatePairStats) => {
-      return {
-        timestamp: activeCandidatePairStats.timestamp,
-        bytesReceived: activeCandidatePairStats.bytesReceived,
-        bytesSent: activeCandidatePairStats.bytesSent
-      };
-    }).catch(() => {
+    })).catch(() => {
```
manjeshbhargav: @makarandp0 I'll leave it up to you, but the catch callback can also be the second argument for `then()`:
```js
.then(onResolved, onRejected);
```

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,94 @@
...
+      _timer: {
+        value: null,
+        writable: true,
+      }
+    });
+  }
+
+  /**
+   * Get ICE connection stats, and extract received and send bytes.
+   * @returns Promise<?{timestamp, bytesReceived, bytesSent}>
```
manjeshbhargav: @makarandp0 `@returns Promise<?RTCIceCandidatePairStats>`?

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,94 @@
...
+   */
+  start(onIceConnectionInactive) {
+    this.stop();
+    this._timer = setInterval(() => {
+      this._getIceConnectionStats().then(iceStats => {
+        if (!iceStats) {
+          return;
+        }
+
+        // if change detected in bytesReceived mark it as lastActiveiy.
```
manjeshbhargav: @makarandp0 I think the code is self-explanatory, so we don't need the comment here.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -0,0 +1,94 @@
...
+        if (!this._lastActivity || this._lastActivity.bytesReceived !== iceStats.bytesReceived) {
+          this._lastActivity = iceStats;
+        }
+
+        // after inactivity threshold fire inactiveCallback.
```
manjeshbhargav: @makarandp0 I think the code is self-explanatory, so we don't need the comment here.

lib/signaling/v2/iceconnectionmonitor.js
makarandp0: @manjeshbhargav, I do see usage of 2nd param of `.then` in our codebase, and thinking we should move to using .catch instead - 1st type of usage could be problematic. More about this antipattern at https://www.datchley.name/promise-patterns-anti-patterns/

PR #643, opened 2019-05-17
JSDK-2357 Filter codecs in unified plan local SDPs based on remote SDPs.

manjeshbhargav:
@syerrapragada @makarandp0 

This PR attempts to mitigate the issues found in [JSDK-2357](https://issues.corp.twilio.com/browse/JSDK-2357), where the signaling transport closes due to large SDP sizes if more than 10-12 Participants are in a Group Room with 1 audio and 1 video Track.

The main reason why this happens is the browser supports more codecs than those supported by VMS, and each of these extra codecs/profiles add significantly to the size of each m= section. The function `unifiedPlanFilterLocalCodecs` in `lib/util/sdp/index.js` filters the codecs in the m=sections of the local SDP based on the codecs present in the corresponding m= sections in the remote SDP.

According to @syerrapragada 's initial tests, the signaling transport can handle about 23 Participants in a Group Room with this change. Your feedback is greatly appreciated.

## TODO

- [x] Unit tests for `unifiedPlanFilterLocalCodecs`
- [x] Smoke tests on Simpler Signaling

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -9,7 +9,7 @@ const DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
 const DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
 const getStatistics = WebRTC.getStats;
 const createCodecMapForMediaSection = require('../../util/sdp').createCodecMapForMediaSection;
-const unifiedPlanfilterLocalCodecs = require('../../util/sdp').unifiedPlanFilterLocalCodecs;
+const unifiedPlanFilterLocalCodecs = require('../../util/sdp').unifiedPlanFilterLocalCodecs;
 const getMediaSections = require('../../util/sdp').getMediaSections;
 const oncePerTick = require('../../util').oncePerTick;
 const setBitrateParameters = require('../../util/sdp').setBitrateParameters;
@@ -613,7 +613,7 @@ class PeerConnectionV2 extends StateMachine {
       }
 
       const sdp = isUnifiedPlan && this._peerConnection.remoteDescription
-        ? unifiedPlanfilterLocalCodecs(offer.sdp, this._peerConnection.remoteDescription.sdp)
+        ? unifiedPlanFilterLocalCodecs(offer.sdp, this._peerConnection.remoteDescription.sdp)
         : offer.sdp;
 
       const updatedSdp = this._setCodecPreferences(
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -9,6 +9,7 @@ const DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
 const DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
 const getStatistics = WebRTC.getStats;
 const createCodecMapForMediaSection = require('../../util/sdp').createCodecMapForMediaSection;
+const filterLocalCodecs = require('../../util/sdp').filterLocalCodecs;
 const getMediaSections = require('../../util/sdp').getMediaSections;
 const oncePerTick = require('../../util').oncePerTick;
 const setBitrateParameters = require('../../util/sdp').setBitrateParameters;
@@ -622,7 +623,9 @@ class PeerConnectionV2 extends StateMachine {
       }
       return this._setLocalDescription({
         type: 'offer',
-        sdp: updatedSdp
+        sdp: this._peerConnection.remoteDescription ?
+          filterLocalCodecs(updatedSdp, this._peerConnection.remoteDescription.sdp)
+          : updatedSdp
       });
     });
   }
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -9,7 +9,7 @@ const DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
 const DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
 const getStatistics = WebRTC.getStats;
 const createCodecMapForMediaSection = require('../../util/sdp').createCodecMapForMediaSection;
-const filterLocalCodecs = require('../../util/sdp').filterLocalCodecs;
+const unifiedPlanfilterLocalCodecs = require('../../util/sdp').unifiedPlanFilterLocalCodecs;
 const getMediaSections = require('../../util/sdp').getMediaSections;
 const oncePerTick = require('../../util').oncePerTick;
 const setBitrateParameters = require('../../util/sdp').setBitrateParameters;
@@ -612,8 +612,12 @@ class PeerConnectionV2 extends StateMachine {
         offer = workaroundIssue8329(offer);
       }
 
+      const sdp = isUnifiedPlan && this._peerConnection.remoteDescription
+        ? unifiedPlanfilterLocalCodecs(offer.sdp, this._peerConnection.remoteDescription.sdp)
+        : offer.sdp;
+
       const updatedSdp = this._setCodecPreferences(
-        offer.sdp,
+        sdp,
         this._preferredAudioCodecs,
         this._preferredVideoCodecs);
 
@@ -623,9 +627,7 @@ class PeerConnectionV2 extends StateMachine {
       }
       return this._setLocalDescription({
         type: 'offer',
-        sdp: this._peerConnection.remoteDescription ?
-          filterLocalCodecs(updatedSdp, this._peerConnection.remoteDescription.sdp)
-          : updatedSdp
+        sdp: updatedSdp
       });
     });
   }
--- a/lib/util/sdp/index.js
+++ b/lib/util/sdp/index.js
@@ -290,6 +290,97 @@ function setSimulcast(sdp, sdpFormat, trackIdsToAttributes) {
 }
 
 /**
+ * Get the matching Payload Types in a unified plan local m= section for a particular remote codec.
+ * @param {string} remoteCodec
+ * @param {PayloadType} remotePt
+ * @param {Map<string, PayloadType>} localCodecsToPts
+ * @param {string} localSection
+ * @param {string} remoteSection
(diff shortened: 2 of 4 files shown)


lib/util/sdp/index.js
```diff
@@ -251,6 +289,84 @@ function setSimulcast(sdp, sdpFormat, trackIdsToAttributes) {
   })).concat('').join('\r\n');
 }
 
+/**
+ * Filter local codecs based on the remote unified plan SDP.
+ * @param {string} localSdp
+ * @param {string} remoteSdp
+ * @returns {string} - Updated local SDP
+ */
+function unifiedPlanFilterLocalCodecs(localSdp, remoteSdp) {
```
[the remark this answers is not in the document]
manjeshbhargav: @syerrapragada Unfortunately Chrome always generates the full unfiltered SDP in `createOffer()` every time. So we need to filter each time.

ramapal:
@manjeshbhargav Was wondering about the case when a new participant joins with fewer supported codecs. For eg. Alice and Bob in a hybrid room {H264 and VP8} and they both started using H264 (lets say). If Charlie joins the room and supports only VP8, what will happen? Charlie will have no video?

manjeshbhargav:
@ramapal In that case there will be a new RTCRtpTransceiver created for Charlie that will negotiate VP8 with VMS.

lib/util/sdp/index.js
```diff
@@ -251,6 +289,84 @@ function setSimulcast(sdp, sdpFormat, trackIdsToAttributes) {
...
+  const localMediaSections = getMediaSections(localSdp);
+  const localSession = localSdp.split('\r\nm=')[0];
+  const remoteMidsToMediaSections = createMidToMediaSectionMap(remoteSdp);
+  return [localSession].concat(localMediaSections.map(localSection => {
```
makarandp0: @manjeshbhargav, this function is getting pretty big - please consider breaking it down further. one idea is to move  the mapping of localMediaSection -> targetSecion into another function 
```
return [localSession].concat(localMediaSections.map(localSection => convertLocalSection(localSection, remoteMidsToMediaSections)));
```

lib/util/sdp/index.js
```diff
@@ -251,6 +289,84 @@ function setSimulcast(sdp, sdpFormat, trackIdsToAttributes) {
...
+    const localMid = getMidForMediaSection(localSection);
+    const remoteSection = localMid && remoteMidsToMediaSections.get(localMid);
+    // Do nothing if either the local m= section does not have an equivalent
+    // remote m= section or if it is neither an audio nor a video m= section.
+    if (!(remoteSection && /^m=(audio|video)/.test(localSection))) {
```
makarandp0: you can optimize a bit by returning early if `localSection` was not video or audio.

```
  // do nothing if neither an audio nor a video m= section.
  if (/^m=(audio|video)/.test(localSection) === false) {
     return localSection;
  }

   const localMid = getMidForMediaSection(localSection);
   const remoteSection = localMid && remoteMidsToMediaSections.get(localMid);
   if (!remoteSection) {
    // the local m= section does not have an equivalent
    // remote m= section
    return localSection;
   }

```

lib/util/sdp/index.js
```diff
@@ -251,6 +289,84 @@ function setSimulcast(sdp, sdpFormat, trackIdsToAttributes) {
...
+    // Construct a Map of the local codec names to their Payload Types.
+    const localCodecToPts = createCodecMapForMediaSection(localSection);
+    // Maintain a list of local Payload Types to retain.
+    let localPts = [];
+
+    remotePtToCodecs.forEach((remoteCodec, remotePt) => {
+      const matchingLocalPts = localCodecToPts.get(remoteCodec) || [];
+      // If there is only one local Payload Type that matches the remote codec,
+      // retain it.
+      if (matchingLocalPts.length === 1) {
```
makarandp0: not really sure if its a case to be considered, but maybe leave early here?
```
if (matchingLocalPts.length === 0) { 
   return; 
}
```

lib/util/sdp/index.js
makarandp0: lot of munging going in here that relies on sdp formats - would be useful to put in inside try catch block and log an error - this would let you know if any of the assumptions about format were wrong and thus led to crash.

lib/signaling/v2/peerconnection.js
```diff
@@ -611,8 +612,12 @@ class PeerConnectionV2 extends StateMachine {
         offer = workaroundIssue8329(offer);
       }
 
+      const sdp = isUnifiedPlan && this._peerConnection.remoteDescription
+        ? unifiedPlanfilterLocalCodecs(offer.sdp, this._peerConnection.remoteDescription.sdp)
+        : offer.sdp;
+
```
makarandp0: consider logging orignalSdp.length, and updatedSdp.length to insights publisher.

lib/util/sdp/index.js
manjeshbhargav: @makarandp0 I think if the length is 0, then the 2 `if` blocks are skipped and we have the desired effect that you have mentioned.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @makarandp0 In order to do this, we have to update the Insights Publisher payload spec and the back-end has to update their data checks so this payload doesn't get rejected.

manjeshbhargav:
@syerrapragada The integration test would do the same thing being done in the unit test. The SDPs used in the unit test are actual SDPs generated by the browser.

lib/util/sdp/index.js
makarandp0: Thanks for breaking down, It would be nice to get more unittest for the individual functions.

test/unit/spec/util/sdp/index.js
```diff
@@ -490,6 +491,222 @@ a=ssrc:0000000000 label:d8b9a935-da54-4d21-a8de-522c87258244\r
   });
 });
 
+describe('unifiedPlanFilterLocalCodecs', () => {
+  it('should filter codecs in a local SDP based on those advertised in the remote SDP', () => {
```
makarandp0: You could add few more unittests for the -ve cases where we do not expect any updates to the offer.sdp

test/unit/spec/util/sdp/index.js
```diff
@@ -490,6 +491,222 @@ a=ssrc:0000000000 label:d8b9a935-da54-4d21-a8de-522c87258244\r
...
+a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f\r
+a=ssrc-group:FID 0000000000 1111111111\r
+a=ssrc:0000000000 cname:s9hDwDQNjISOxWtK\r
+a=ssrc:0000000000 msid:7a9d401b-3cf6-4216-b260-78f93ba4c32e d8b9a935-da54-4d21-a8de-522c87258244\r
+a=ssrc:0000000000 mslabel:7a9d401b-3cf6-4216-b260-78f93ba4c32e\r
+a=ssrc:0000000000 label:d8b9a935-da54-4d21-a8de-522c87258244\r
+a=ssrc:1111111111 cname:s9hDwDQNjISOxWtK\r
+a=ssrc:1111111111 msid:7a9d401b-3cf6-4216-b260-78f93ba4c32e d8b9a935-da54-4d21-a8de-522c87258244\r
+a=ssrc:1111111111 mslabel:7a9d401b-3cf6-4216-b260-78f93ba4c32e\r
+a=ssrc:1111111111 label:d8b9a935-da54-4d21-a8de-522c87258244\r
```
makarandp0: it would be useful to see the resultant/updated sdp as part of this test.

test/unit/spec/util/sdp/index.js
makarandp0: I am sure you know this - but running `npm run cover` would give you an idea of missing code paths.

lib/util/sdp/index.js
```diff
@@ -75,6 +87,32 @@ function createPtToCodecName(mediaSection) {
...
+/**
+ * Get the associated fmtp attributes for the given Payload Type in an m= section.
+ * @param {PT} pt
+ * @param {string} mediaSection
+ * @returns {?object}
+ */
+function getFmtpAttributesForPt(pt, mediaSection) {
+  const fmtpRegex = new RegExp(`^a=fmtp:${pt} (.+)$`, 'm');
+  const matches = mediaSection.match(fmtpRegex);
+  return matches && matches[1].split(';').reduce((attrs, nvPair) => {
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Good catch! I will convert it to `(matches[1] || '').split(';')`.

test/unit/spec/util/sdp/index.js
manjeshbhargav: @makarandp0 We generally don't log intermediate results.

syerrapragada:
Let us get a green Travis build before we merge.

lib/signaling/v2/peerconnection.js
ceaglest: Should we limit the application of this filtering to Group Rooms only (ice-lite agents)? In a Peer-to-Peer Room, another Participant could change their codec support during the life of the PeerConnection (including in an Answer). A JS Client with these changes might artificially restrict the codecs available in their offer, resulting in a preferred codec not being selected (without an additional re-offer). I don't want to write re-offer logic if possible.

manjeshbhargav:
@ceaglest 

>  I agree with @makarandp0 that more negative test coverage would be nice to have.

The current unit test covers negative use-cases where a local m= section that is not in the remote sdp will be untouched.

> Should we limit the application of this filtering to Group Rooms only (ice-lite agents)?

I'm ok with testing for ice lite peers in the case of not forwarding candidates. But I'm reluctant to infer that ice lite agents are Group Rooms. Also, regarding changes in codec support in Peer to Peer Rooms,  in the case of the offerer doing that, then it will add a new transceiver instead of re-using an existing one. Regarding the answerer, won't the answer generated only contain those codecs matched with the offer?

PR #649, opened 2019-05-24
JSDK-2368 Consider only non-rtx payload types in the remote RTCSessionDescription while filtering local codecs.

manjeshbhargav:
@syerrapragada @makarandp0 

This PR fixes [JSDK-2368](https://issues.corp.twilio.com/browse/JSDK-2368), which was being caused due to a bug in the filtering logic where rtx payload types for filtered codecs were being retained. For ex, if the codec payload type 50 was removed, we would still retain the rtx payload type with `apt=50`.

The solution is to not consider rtx payload types in the remote RTCSessionDescription while constructing a list of local codecs to retain.

### DIFF
--- a/lib/util/sdp/index.js
+++ b/lib/util/sdp/index.js
@@ -354,13 +354,15 @@ function unifiedPlanFilterCodecsInLocalMediaSection(localSection, remoteMidsToMe
   const remotePtToCodecs = createPtToCodecName(remoteSection);
   // Construct a Map of the local codec names to their Payload Types.
   const localCodecsToPts = createCodecMapForMediaSection(localSection);
-  // Maintain a list of local Payload Types to retain.
-  let localPts = flatMap(Array.from(remotePtToCodecs), ([remotePt, remoteCodec]) => unifiedPlanGetMatchingLocalPayloadTypes(
-    remoteCodec,
-    remotePt,
-    localCodecsToPts,
-    localSection,
-    remoteSection));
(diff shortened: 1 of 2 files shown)


makarandp0:
@manjeshbhargav, Its amazing that you managed to debug and find the issue so fast. My knowledge of sdps is limited, and I do not  fully understand the change. @syerrapragada may have more insights.

But overall I think a unit test that shows the generated sdp would really help. such change would then show the before and after of the generated sdp. The unit test you have do not list the output sdp as part of the test. I know it would be lot of text, but I think its worth it. 

One question the initial test was valid - and you are updating it with more specific example right? maybe it warrants a seperate test? 

Since this change is so involved - maybe its worthwhile to have multiple tests with different kinds of sdps and their treatment (showing both input sdp and output sdp)

manjeshbhargav:
@makarandp0 ,

> But overall I think a unit test that shows the generated sdp would really help. such change would then show the before and after of the generated sdp. The unit test you have do not list the output sdp as part of the test. I know it would be lot of text, but I think its worth it.

I think we should have a more general way of running tests in a verbose manner. This can be something we can specify in a command line parameter while running tests.

> One question the initial test was valid - and you are updating it with more specific example right? maybe it warrants a separate test?

The previous mocked SDP did not account for RTX codecs being present in the remote SDP. So, this change updates the mock SDP to have RTX codecs.

PR #654, opened 2019-06-03
Prepping for 2.0.0-rc27.

manjeshbhargav:
@makarandp0 @syerrapragada 

Since we have decided to push Unified Plan support for Chrome further by at least a week or two, I'm bringing in Region Selection and ICE improvements for the next beta release. Please review these cherry-picked commits to see if the changes are ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta10 (in progress)
+==========================
+
+New Features
+------------
+- By default, twilio-video.js connects to your nearest signaling server, as determined by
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+  ```js
+  const { connect } = require('twilio-video');
+  const room = await connect(token, {
+    region: 'de1'
+  });
+  ```
+
+  This will guarantee that your signaling traffic will terminate in Germany. For other possible values
+  for region, please refer to this [table](https://www.twilio.com/docs/video/ip-address-whitelisting#media-servers).
```
syerrapragada: Link is pointing to media-servers. shouldn't it point to https://www.twilio.com/docs/video/ip-address-whitelisting#signaling-communication ?

manjeshbhargav: @syerrapragada I think at the time of implementing region selection, only the media servers section contained the list of all possible regions. I will make this change. Thanks for catching it.

Issue #656, opened 2019-06-05
Local participant's 'trackUnpublished' event not fired


dferrazm:
It seems that the `trackUnpublished` event is not being fired for the local participant whenever unpublishing a local video track (didn't test for the audio track, but perhaps has the same issue). The `trackPublished` is fired fine on the local participant whenever a new local video track is published.

Remote participants' `trackPublished` and `trackUnpublished` also work fine.

**Code to reproduce the issue:**

```js
localParticipant.on('trackPublished', ({ track }) => {
  console.log('track published:', track.id)
})

localParticipant.on('trackUnpublished', ({ track }) => {
  console.log('track unpublished:', track.id)
})

createLocalVideoTrack().then(track => localParticipant.publishTrack(track))
localParticipant.videoTracks.forEach(trackPublication => trackPublication.unpublish())
createLocalVideoTrack().then(track => localParticipant.publishTrack(track))
```

**Expected behavior:**

I'd expect the following outputs:
```
// track published: 111
// track unpublished: 111
// track published: 222
```

**Actual behavior:**

It seems that the `unpublishedTrack` event is not being fired.
```
// track published: 111
// track published: 222
```

**Software versions:**

 - [X] Browser(s): Chrome 74, Firefox 67
 - [X] Operating System: Ubuntu 18.04
 - [X] twilio-video.js: 2.0.0-beta9
 - [X] Third-party libraries (e.g., Angular, React, etc.): React

dferrazm:
I've checked the source code, and I noticed that in the `localparticipant.js`, the `trackUnpublished` event is non-existent. I'm not sure if that's intended.

Perhaps what we're missing here is on line `477`, right after the `_removeTrackPublication` call, having the following:

```js
if (self._signaling.state === 'connected') {
  setTimeout(() => {
    self.emit('trackUnpublished', localTrackPublication);
  });
}
```

(copied/based on how the `trackPublished` counter-part event is emitted)

dferrazm:
I can confirm that it worked. I could work on a PR if needed.

manjeshbhargav:
Hi @dferrazm ,

[...]

PR #657, opened 2019-06-05
Fix issue where the 'trackUnpublished' event was not being fired

dferrazm:
Fix issue #656 where the `trackUnpublished` event was not being fired for the local participant's tracks when tracks publications were removed.

**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.


dferrazm:
Lacking unit tests. Let me know if I can proceed co that I can add the tests.

manjeshbhargav:
@makarandp0 Right now, "trackUnpublished" is not part of the Video SDK spec. We would need to add it to the spec and implement it the right way, that is after the backend response. I will create an internal ticket to track this work. It looks like this event is useful in decoupling React Components for Participants from media manipulation actions.

dferrazm:
Alright. Thanks! I'll hold on this.

PR #659, opened 2019-06-06
Setting release date for 2.0.0-beta10.

manjeshbhargav:
@syerrapragada

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.0.0-beta10 (in progress)
-==========================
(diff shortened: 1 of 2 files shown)


smkrn110:
@manjeshbhargav 

Is beta10 is ready to use? as I am seeing inprogress 
https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md 

where according to this it was release on June 6.
https://github.com/twilio/twilio-video.js/releases

Also I will be upgrading from beta5 to beta9 or beta10 can you confirm that there are no breaking changes, as we have some when migrating from 1.x to 2.x?

Thanks

manjeshbhargav:
Hi @smkrn110,

Right now, we are releasing off of the "master-chrome-planb" branch because of some changes in master that are blocked from release. I will update master when those changes are unblocked.

PR #662, opened 2019-06-12
JSDK-2387 Consuming REGIONS environment variable in integration tests for signaling region.

manjeshbhargav:
@makarandp0 @syerrapragada 

This PR adds a new environment variable `REGIONS` to the integration test suite. In `stage`, you can use this to pass only those signaling regions that are supported in this environment like this:

```
... ENVIRONMENT=stage REGIONS=au1,in1,us1 ... npm run test:integration 
```

If you don't specify `REGIONS`, then all the signaling regions will be tested.

**Contributing to Twilio**

### DIFF
--- a/test/integration/spec/connect.js
+++ b/test/integration/spec/connect.js
@@ -132,8 +132,8 @@ describe('connect', function() {
         });
 
         it(`should return a CancelablePromise that ${region === 'invalid' ? 'rejects with a SignalingConnectionError' : 'resolves with a Room'}`, async () => {
-          let error;
-          let room;
+          let error = null;
+          let room = null;
           try {
             room = await cancelablePromise;
           } catch (error_) {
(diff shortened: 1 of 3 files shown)


test/integration/spec/connect.js
```diff
@@ -104,41 +104,53 @@ describe('connect', function() {
...
           assert(cancelablePromise instanceof CancelablePromise);
         });
 
-        if (regionType === 'invalid') {
-          it('should return a CancelablePromise that rejects with a SignalingConnectionError', async () => {
-            try {
-              const room = await cancelablePromise;
+        it(`should return a CancelablePromise that ${region === 'invalid' ? 'rejects with a SignalingConnectionError' : 'resolves with a Room'}`, async () => {
+          let error;
+          let room;
```
makarandp0: since you are using room in finally block, I suggest explicitly initialize to `null` here

test/integration/spec/connect.js
```diff
@@ -104,41 +104,53 @@ describe('connect', function() {
...
+          try {
+            room = await cancelablePromise;
+          } catch (error_) {
+            error = error_;
+          } finally {
+            if (room) {
```
makarandp0: you could move this `if` block after the assertions. 
```
            if (region === 'invalid') {
              assert(!room, `Connected to Room ${room && room.sid} with an invalid signaling region "${region}"`);
              assert(error instanceof SignalingConnectionError);
            } else {
              assert(!error);
              assert(room instanceof Room);
            }
            if (room) {
              room.disconnect();
            }
```

manjeshbhargav: @makarandp0 I would leave it where it is because if any assertion fails for some reason, then we won't disconnect from the Room.

PR #663, opened 2019-06-12
Updating "region" docs and setting release date.

manjeshbhargav:
@ceaglest @anna-vasilko @aymenn 

Please take a look at these changes to the documentation for `region`, and let me know what you think.

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.0.0-beta11 (June 13, 2019)
+2.0.0-beta11 (June 12, 2019)
 ============================
 
 New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.0.0-beta11 (in progress)
-==========================
+2.0.0-beta11 (June 13, 2019)
+============================
 
 New Features
 ------------
@@ -20,6 +20,20 @@ New Features
 
   This will guarantee that your signaling traffic will terminate in Germany. For other possible values
   for region, please refer to this [table](https://www.twilio.com/docs/video/ip-address-whitelisting#signaling-communication).
+  If you specify an invalid value for `region`, `connect` will raise a [SignalingConnectionError](https://www.twilio.com/docs/api/errors/53000):
+
+  ```js
+  const { connect } = require('twilio-video');
+
+  try {
+    const room = await connect(token, {
+      region: 'foo'
+    });
+  } catch (error) {
+    assert.equal(error.code, 53000);
+    assert.equal(error.message, 'Signaling connection error');
+  }
+  ```
 
 Bug Fixes
 ---------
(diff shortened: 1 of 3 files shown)


lib/connect.js
```diff
@@ -304,12 +304,13 @@ function connect(token, options) {
...
- * @property {string} [region='gll'] - Preferred region; By default, the client will connect to the
+ * @property {string} [region='gll'] - Preferred signaling region; By default, you will be connected to the
  *   nearest signaling server determined by latency based routing. Setting a value other
  *   than <code style="padding:0 0">gll</code> bypasses routing and guarantees that signaling traffic will be
- *   terminated in the region that you prefer. If you are connecting to a Group Room created with the <code style="padding:0 0">gll</code>
+ *   terminated in the region that you prefer. If you are the first to join a Group Room created with the <code style="padding:0 0">gll</code>
  *   Media Region (either Ad-Hoc or via the REST API), then the Room's Media Region will be selected
- *   based upon your client's region
+ *   based upon your signaling region. Please refer to this <a href="https://www.twilio.com/docs/video/ip-address-whitelisting#signaling-communication" target="_blank">table</a>
+ *   for the list of supported signaling regions.
```
ceaglest: I would remove the sentence about GLL media behavior until we are ready with the feature. This would match the iOS docs.

Other than that it looks good to me. I think we should take another pass at the IDL to arrive on final terminology before GA.

CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ New Features
...
+
+  try {
+    const room = await connect(token, {
+      region: 'foo'
+    });
+  } catch (error) {
+    assert.equal(error.code, 53000);
+    assert.equal(error.message, 'Signaling connection error');
+  }
+  ```
```
aymenn: Some small suggestions below:

By default, twilio-video.js connects to your nearest signaling server by automatically finding the signaling server connection with the lowest latency. ﻿Howe﻿ver﻿,﻿ in some cases,﻿ you may ﻿have a need to ﻿lock in to a specific region. Y﻿ou can﻿ now override this behavior by using a﻿ new ConnectOptions flag called region﻿.﻿ This will make sure that your signaling traffic will terminate in the region you specify. The default value for region is “gll“ (Global Low Latency). With this setting, twilio-video.js automatically finds the signalling server with the lowest latency. Please refer to this table for a complete regio﻿﻿ns list ﻿(JSDK-2338)

For example, to connect to “de1”:﻿
const { connect } = require(‘twilio-video‘);
const room = await connect(token, {
  region: ‘de1‘
});

Note, if you specify an invalid value for region, connect will raise a SignalingConnectionError:
const { connect } = require(‘twilio-video‘);

try {
  const room = await connect(token, {
    region: ‘foo‘
  });
} catch (error) {
  assert.equal(error.code, 53000);
  assert.equal(error.message, ‘Signaling connection error‘);
}

lib/connect.js
innerverse: Although I prefer names like "auto" or "global" we decided to converge on the same API term that Voice uses in their APIs for consistency across Voice and Video, which is "gll" (not my favorite, but we felt consistency was more useful to customers than a new, better name...)

aymenn: If you asked me 2 months ago I would say let's use "gll" for consistency :) However, now I would not consider this API consistency to be a factor in deciding on the right API given how far apart Voice and Video are.  Everyone Ive spoken to internally has said gll is not a good name. We have an opportunity to give it a better name! Again, this would only be public facing. Internally, we can call it whatever you like.

PR #665, opened 2019-06-15
JSDK-2307 MSP objects do not recover on JSDK side after SFU Failover	

makarandp0:
For network quality updates client creates a `NetworkQualityMonitor` object listen on a datachannel for network quality monitor. The id for the channel is obtained from RSP field (`roomState.media_signaling.network_quality.transport.label`). 

During the call in a group room if the VMS process gets killed, a new instance gets created - This results in an `update` message on signaling channel with a new peerconnection id. Client responds by creating a new peer connection.

The problem was we were not rehooking our NetoworkMonitor to the dataChannel from the new peerconnection in such cases. 

With this change, we fix this by updating the dataChannel that we listen on everytime a new dataChannel with desired id is detected. As part of NetoworkMonitor creation, client now listens for any `trackAdded` messages from peerConnectionManager. Whenever a  track with the id that we were interested in gets added, it will start listening on new track.

I have verified the changes by killing the VMS and ensuring that we continue to update networkQualityLevelChanged.

Still TODO:
- [ ] Add tests
- [x] Incorporate similar changes for DominantSpeaker channel.

### DIFF
--- a/lib/signaling/v2/networkqualitysignaling.js
+++ b/lib/signaling/v2/networkqualitysignaling.js
@@ -109,6 +109,7 @@ class NetworkQualitySignaling extends EventEmitter {
         }
       }
     });
+
     mediaSignalingTransport.on('message', message => {
       switch (message.type) {
         case 'network_quality':
--- a/lib/signaling/v2/networkqualitysignaling.js
+++ b/lib/signaling/v2/networkqualitysignaling.js
@@ -109,9 +109,7 @@ class NetworkQualitySignaling extends EventEmitter {
         }
       }
     });
-    let internal = 0;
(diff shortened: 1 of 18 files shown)


lib/signaling/v2/networkqualitysignaling.js
```diff
@@ -109,8 +109,9 @@ class NetworkQualitySignaling extends EventEmitter {
         }
       }
     });
-
+    let internal = 0;
```
makarandp0: TODO: remove before merge.

makarandp0:
Closing this and reopening as https://github.com/twilio/twilio-video.js/pull/669. I have incorporated feedback in the new PR.

PR #667, opened 2019-06-18
JSDK-2385 Add or rewrite Track IDs in local SDPs even if "appdata" field is not present.

manjeshbhargav:
@makarandp0 @syerrapragada 

This PR has two sections:

1. As of now, each m= section has this line `a=msid:<streamId> <appdata>`. The `appdata` field represents the Track ID associated with the m= section. This need not be the same as the ID of the actual MediaStreamTrack that is assigned to this m= section. So we have been rewriting the `appdata` field with the actual MediaStreamTrack ID. In this PR, we make sure that we write the MediaStreamTrack ID even if `appdata` is not present in the m= section, because WebRTC spec has been updated to remove `appdata`.

2. A new (or rather a previously available) `sdpSemantics` ConnectOptions flag is (re)introduced to switch easily between `plan-b` and `unified-plan`. Now, we don't need to maintain a separate branch for the Plan B implementation.

NOTE: Because we now calculate the SDP format after connect() is called in Chrome, JSDK-2392 is fixed.

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -142,6 +142,8 @@ function connect(token, options) {
     preferredVideoCodecs: [],
     realm: constants.DEFAULT_REALM,
     region: constants.DEFAULT_REGION,
+    // TODO(mmalavalli): Remove once we decide to support Unified Plan on Chrome 72+
+    sdpSemantics: constants.DEFAULT_CHROME_SDP_SEMANTICS,
     signaling: SignalingV2
   }, util.filterObject(options));
 
--- a/lib/data/sender.js
+++ b/lib/data/sender.js
@@ -98,7 +98,11 @@ class DataTrackSender extends DataTrackTransceiver {
       }
     });
     this._clones.forEach(clone => {
-      clone.send(data);
+      try {
+        clone.send(data);
+      } catch (error) {
+        // Do nothing.
+      }
     });
     return this;
   }
--- a/lib/data/transport.js
+++ b/lib/data/transport.js
@@ -48,7 +48,11 @@ class DataTransport extends EventEmitter {
    */
   _publish(message) {
     const data = JSON.stringify(message);
-    this._dataChannel.send(data);
+    try {
+      this._dataChannel.send(data);
+    } catch (error) {
+      // Do nothing.
+    }
   }
 
   /**
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -36,12 +36,13 @@ const guess = guessBrowser();
 const isChrome = guess === 'chrome';
 const isFirefox = guess === 'firefox';
 const isSafari = guess === 'safari';
-const sdpFormat = getSdpFormat();
-const isUnifiedPlan = sdpFormat === 'unified';
 
 const firefoxMajorVersion = isFirefox
   ? parseInt(navigator.userAgent.match(/Firefox\/(\d+)/)[1], 10)
   : null;
+
+let isUnifiedPlan;
+let sdpFormat;
 
 /*
 PeerConnectionV2 States
@@ -109,10 +110,13 @@ class PeerConnectionV2 extends StateMachine {
       RTCSessionDescription: DefaultRTCSessionDescription
     }, options);
 
-    const RTCPeerConnection = options.RTCPeerConnection;
     const configuration = getConfiguration(options);
+    sdpFormat = sdpFormat || getSdpFormat(configuration.sdpSemantics);
+    isUnifiedPlan = typeof isUnifiedPlan === 'boolean' ? isUnifiedPlan : sdpFormat === 'unified';
+
     const localMediaStream = isUnifiedPlan ? null : new options.MediaStream();
     const logLevels = buildLogLevels(options.logLevel);
+    const RTCPeerConnection = options.RTCPeerConnection;
     const peerConnection = new RTCPeerConnection(configuration, options.chromeSpecificConstraints);
 
     if (options.dummyAudioMediaStreamTrack) {
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -1,31 +1,44 @@
 'use strict';
 
-const WebRTC = require('@twilio/webrtc');
+const {
+  MediaStream: DefaultMediaStream,
+  RTCIceCandidate: DefaultRTCIceCandidate,
+  RTCPeerConnection: DefaultRTCPeerConnection,
+  RTCSessionDescription: DefaultRTCSessionDescription,
+  getStats: getStatistics
+} = require('@twilio/webrtc');
+
 const { guessBrowser } = require('@twilio/webrtc/lib/util');
 const { getSdpFormat } = require('@twilio/webrtc/lib/util/sdp');
-const DefaultMediaStream = WebRTC.MediaStream;
-const DefaultRTCIceCandidate = WebRTC.RTCIceCandidate;
-const DefaultRTCPeerConnection = WebRTC.RTCPeerConnection;
-const DefaultRTCSessionDescription = WebRTC.RTCSessionDescription;
-const getStatistics = WebRTC.getStats;
-const createCodecMapForMediaSection = require('../../util/sdp').createCodecMapForMediaSection;
-const unifiedPlanFilterLocalCodecs = require('../../util/sdp').unifiedPlanFilterLocalCodecs;
-const getMediaSections = require('../../util/sdp').getMediaSections;
-const oncePerTick = require('../../util').oncePerTick;
-const setBitrateParameters = require('../../util/sdp').setBitrateParameters;
-const setCodecPreferences = require('../../util/sdp').setCodecPreferences;
-const setSimulcast = require('../../util/sdp').setSimulcast;
-const revertSimulcastForNonVP8MediaSections = require('../../util/sdp').revertSimulcastForNonVP8MediaSections;
-const unifiedPlanAddOrRewriteNewTrackIds = require('../../util/sdp').unifiedPlanAddOrRewriteNewTrackIds;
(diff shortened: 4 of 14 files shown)


lib/data/sender.js
```diff
@@ -98,7 +98,11 @@ class DataTrackSender extends DataTrackTransceiver {
       }
     });
     this._clones.forEach(clone => {
-      clone.send(data);
+      try {
+        clone.send(data);
+      } catch (error) {
```
makarandp0: curious- why was this catch necessary? in what cases do we expect errors? Maybe worth adding a console.log here?

lib/data/transport.js
```diff
@@ -48,7 +48,11 @@ class DataTransport extends EventEmitter {
    */
   _publish(message) {
     const data = JSON.stringify(message);
-    this._dataChannel.send(data);
+    try {
+      this._dataChannel.send(data);
+    } catch (error) {
+      // Do nothing.
```
makarandp0: same. when is this expected - how about logging the errors?

lib/signaling/v2/peerconnection.js
```diff
@@ -16,7 +16,8 @@ const setBitrateParameters = require('../../util/sdp').setBitrateParameters;
 const setCodecPreferences = require('../../util/sdp').setCodecPreferences;
 const setSimulcast = require('../../util/sdp').setSimulcast;
 const revertSimulcastForNonVP8MediaSections = require('../../util/sdp').revertSimulcastForNonVP8MediaSections;
-const unifiedPlanRewriteTrackIds = require('../../util/sdp').unifiedPlanRewriteTrackIds;
+const unifiedPlanAddOrRewriteNewTrackIds = require('../../util/sdp').unifiedPlanAddOrRewriteNewTrackIds;
```
makarandp0: maybe it's worth group these with
``` 
const {
  createCodecMapForMediaSection,
  unifiedPlanFilterLocalCodecs,
  getMediaSections,
  setBitrateParameters,
  setCodecPreferences,
  setSimulcast,
  revertSimulcastForNonVP8MediaSections,
  unifiedPlanRewriteTrackIds,
  unifiedPlanAddOrRewriteNewTrackIds,
} = require('../../util/sdp');
```

lib/signaling/v2/peerconnection.js
```diff
@@ -682,19 +690,34 @@ class PeerConnectionV2 extends StateMachine {
   }
 
   /**
-   * Rewrite local MediaStreamTrack IDs in the given Unified Plan RTCSessionDescription.
+   * Add or rewrite Rewrite local MediaStreamTrack IDs in the given Unified Plan RTCSessionDescription.
```
makarandp0: Add or rewrite ~Rewrite~ local MediaStreamTrack IDs in the given Unified Plan RTCSessionDescription.

lib/util/sdp/index.js
```diff
@@ -437,25 +437,64 @@ function revertSimulcastForNonVP8MediaSections(localSdp, localSdpWithoutSimulcas
...
-      }
+function unifiedPlanAddOrRewriteTrackIds(sdp, midsToTrackIds) {
+  const mediaSections = getMediaSections(sdp);
+  const session = sdp.split('\r\nm=')[0];
+  return [session].concat(mediaSections.map(mediaSection => {
+    // Do nothing if the m= section represents neither audio nor video.
+    if (!/^m=(audio|video)/.test(mediaSection)) {
+      return mediaSection;
+    }
+    // This shouldn't happen, but in case there is no MID for the m= section, do nothing.
```
makarandp0: since it shouldn't happen...lets add some log/telemetry to see if it happens.

lib/util/sdp/index.js
```diff
@@ -437,25 +437,64 @@ function revertSimulcastForNonVP8MediaSections(localSdp, localSdpWithoutSimulcas
...
+      return mediaSection;
     }
-    return sdp;
-  }, sdp);
+    // In case there is no Track ID for the given MID in the map, do nothing.
+    const trackId = midsToTrackIds.get(mid);
+    if (!trackId) {
+      return mediaSection;
+    }
+    // This shouldn't happen, but in case there is no a=msid: line, do nothing.
```
makarandp0: since it shouldn't happen...lets add some log/telemetry to see if it happens.

package.json
```diff
@@ -117,7 +117,7 @@
     "cover": "istanbul cover node_modules/mocha/bin/_mocha -- ./test/unit/index.js"
   },
   "dependencies": {
-    "@twilio/webrtc": "github:twilio/twilio-webrtc.js#4.1.0-rc1",
+    "@twilio/webrtc": "github:twilio/twilio-webrtc.js#87f7e61106f58e29364999d8753c773e5602536f",
```
makarandp0: is this temporary right ?

test/unit/spec/util/sdp/index.js
```diff
@@ -846,16 +847,46 @@ describe('revertSimulcastForNonVP8MediaSections', () => {
...
-      assert.equal(section.match(/^a=ssrc:.+ msid:.+ (.+)$/m)[1], trackId);
+describe('unifiedPlanAddOrRewriteNewTrackIds', () => {
+  [true, false].forEach(withAppData => {
+    context(`when the Unified Plan SDP ${withAppData ? 'has' : 'does not have'} MediaStreamTrack IDs in a=msid: lines`, () => {
+      it('should rewrite Track IDs with the IDs of MediaStreamTracks associated with unassigned RTCRtpTransceivers', () => {
+        const sdp = makeSdpWithTracks('unified', {
+          audio: ['foo', 'bar'],
+          video: ['baz', 'zee']
+        }, null, null, withAppData);
+        const newTrackIdsByKind = new Map([['audio', ['yyy']], ['video', ['zzz']]]);
```
makarandp0: consider  using multiple ids for one of the tests to make sure they are handled correctly. maybe just 
```
 const newTrackIdsByKind = new Map([['audio', ['yyy', 'xxx']], ['video', ['zzz']]]);
```

lib/data/sender.js
manjeshbhargav: @makarandp0 In integration tests, sometime when you disconnect from a Room, if the timing is right, trying to send data on RTCDataChannels whose RTCPeerConnection is closed throws an error. This makes sure that the tests don't fail because of it. Regarding logging, we should do it once we thread the logging module down to this level.

lib/data/transport.js
manjeshbhargav: @makarandp0 In integration tests, sometime when you disconnect from a Room, if the timing is right, trying to send data on RTCDataChannels whose RTCPeerConnection is closed throws an error. This makes sure that the tests don't fail because of it. Regarding logging, we should do it once we thread the logging module down to this level.

package.json
manjeshbhargav: @makarandp0 Yes. Before I release the next RC, I will also release the corresponding twilio-webrtc.js RC and update package.json.

lib/connect.js
```diff
@@ -142,6 +142,8 @@ function connect(token, options) {
     preferredVideoCodecs: [],
     realm: constants.DEFAULT_REALM,
     region: constants.DEFAULT_REGION,
+    // TODO(mmalavalli): Remove once we decide to support Unified Plan on Chrome 72+
+    sdpSemantics: constants.DEFAULT_CHROME_SDP_SEMANTICS,
```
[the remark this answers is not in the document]
manjeshbhargav: @syerrapragada This is internal only, just to make it to switch to default SDP format later.

lib/signaling/v2/peerconnection.js
```diff
@@ -682,19 +690,34 @@ class PeerConnectionV2 extends StateMachine {
...
+    const midsToTrackIds = new Map(assignedTransceivers.map(({ mid, sender }) => [mid, sender.track.id]));
+    const sdp1 = unifiedPlanAddOrRewriteTrackIds(description.sdp, midsToTrackIds);
+
+    // NOTE(mmalavalli): Chrome and Safari do not apply the offer until they get an answer.
+    // So, we add or re-write the actual MediaStreamTrack IDs to the unassigned m= sections here.
+    const unassignedTransceivers = activeTransceivers.filter(({ mid }) => !mid);
+    const newTrackIdsByKind = new Map(['audio', 'video'].map(kind => [
+      kind,
+      unassignedTransceivers.filter(({ sender }) => sender.track.kind === kind).map(({ sender }) => sender.track.id)
+    ]));
```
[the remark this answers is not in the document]
manjeshbhargav: @syerrapragada I think `track` won't be null unless we've called `sender.replaceTrack(null)`, which we don't.

lib/util/sdp/index.js
```diff
@@ -437,25 +437,64 @@ function revertSimulcastForNonVP8MediaSections(localSdp, localSdpWithoutSimulcas
...
+ * @param {Map<Track.Kind, Array<Track.ID>>} trackIdsByKind
+ * @returns {string}
+ */
+function unifiedPlanAddOrRewriteNewTrackIds(sdp, trackIdsByKind) {
+  // NOTE(mmalavalli): The m= sections for the new MediaStreamTracks are usually
+  // present after the m= sections for the existing MediaStreamTracks, in order
+  // of addition.
+  const newMidsToTrackIds = Array.from(trackIdsByKind).reduce((midsToTrackIds, [kind, trackIds]) => {
+    const mediaSections = getMediaSections(sdp, kind);
+    const newMids = mediaSections.slice(mediaSections.length - trackIds.length).map(getMidForMediaSection);
```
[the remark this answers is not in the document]
manjeshbhargav: @syerrapragada No. Since `setLocalDescription` is not called, we can't rely on `transceiver.mid`, which is null. We have to use the fact that new m= sections are created in the order of calls to `RTCPeerConnection.addTransceiver()`.

PR #669, opened 2019-06-19
JSDK-2307 MSP objects do not recover on JSDK side after SFU Failover

makarandp0:
Note: This PR was originally opened as https://github.com/twilio/twilio-video.js/pull/665, 

For network quality and dominant speaker updates client  listen on a datachannel. The id for the channels is obtained from RSP field (roomState.media_signaling.network_quality.transport.label, roomState.media_signaling.network_quality.transport.label).

During the call in a group room if the VMS process gets killed, a new instance gets created - This results in an update message on signaling channel with a new peerconnection id. Client responds by creating a new peer connection.

The problem was we were not rehooking our NetoworkMonitor, Dominant Spekear dataChannels from the new peerconnection in such cases.

With this change, we fix this by tearing down the monitors after close event on datachannel. They get recreated when a new update message is received.
I have verified the changes by killing the VMS and ensuring that we continue to update networkQualityLevelChanged, and Dominant Speaker changes.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #671, opened 2019-06-21
JSDK-2307 MSP objects do not recover on JSDK side after SFU Failover …

makarandp0:
Poriting #669 into `support-1.x-chrome-planb` branch 

### DIFF
--- a/test/unit/spec/signaling/v2/room.js
+++ b/test/unit/spec/signaling/v2/room.js
@@ -1171,7 +1171,7 @@ describe('RoomV2', () => {
     });
   });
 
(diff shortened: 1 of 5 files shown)


test/unit/spec/signaling/v2/room.js
```diff
@@ -1170,8 +1171,115 @@ describe('RoomV2', () => {
     });
   });
 
-  // Network Quality Signaling
-  // -------------------------
+  describe('Dominant Speaker', () => {
```
manjeshbhargav: @makarandp0 "Dominant Speaker **Signaling**"

PR #683, opened 2019-07-02
BandwidthProfileOptions: mode, maxSubscriptionBitrate and maxTracks.

manjeshbhargav:
@makarandp0 @innerverse 

This PR implements the BandwidthProfileOptions specified as a ConnectOptions parameter `bandwidthProfile`. Please take a look when you can.

**Contributing to Twilio**

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -521,6 +521,40 @@ function valueToJSON(value) {
 }
 
 /**
+ * Create the bandwidth profile payload included in an RSP connect message.
+ * @param {BandwidthProfileOptions} bandwidthProfile
+ * @returns {object}
+ */
+function createBandwidthProfilePayload(bandwidthProfile) {
+  const payload = {};
+  if (typeof bandwidthProfile.video !== 'object') {
+    return payload;
+  }
+
+  const mode = bandwidthProfile.video.mode;
+  if (typeof mode === 'string') {
+    payload.mode = mode;
+  }
+
+  const maxSubscriptionBitrate = bandwidthProfile.video.maxSubscriptionBitrate;
+  if (typeof maxSubscriptionBitrate === 'number') {
+    payload.video = Object.assign({
+      // eslint-disable-next-line
+      max_subscription_bandwidth: Math.round(maxSubscriptionBitrate / 1024)
+    }, payload.video);
+  }
+
+  const maxTracks = bandwidthProfile.video.maxTracks;
+  if (typeof maxTracks === 'number') {
+    payload.video = Object.assign({
+      // eslint-disable-next-line
+      max_tracks: maxTracks
+    }, payload.video);
+  }
+  return payload;
+}
+
+/**
  * Create the Media Signaling payload included in an RSP connect message.
  * @param {boolean} dominantSpeaker - whether to enable the Dominant Speaker
  *   protocol or not
@@ -581,6 +615,7 @@ function inRange(num, min, max) {
 }
 
 exports.constants = constants;
+exports.createBandwidthProfilePayload = createBandwidthProfilePayload;
 exports.createMediaSignalingPayload = createMediaSignalingPayload;
 exports.createSubscribePayload = createSubscribePayload;
 exports.asLocalTrack = asLocalTrack;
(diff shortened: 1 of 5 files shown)


lib/util/index.js
```diff
@@ -520,6 +520,40 @@ function valueToJSON(value) {
...
+  }
+
+  const mode = bandwidthProfile.video.mode;
+  if (typeof mode === 'string') {
+    payload.mode = mode;
+  }
+
+  const maxSubscriptionBitrate = bandwidthProfile.video.maxSubscriptionBitrate;
+  if (typeof maxSubscriptionBitrate === 'number') {
+    payload.video = Object.assign({
```
makarandp0: do we need `Object.assign` here? Seems like 
```
payload.video = {
   max_subscription_bandwidth: Math.round(maxSubscriptionBitrate / 1024)
};
```
would just do the job, and is more readable.

lib/util/index.js
```diff
@@ -520,6 +520,40 @@ function valueToJSON(value) {
...
+      // eslint-disable-next-line
+      max_subscription_bandwidth: Math.round(maxSubscriptionBitrate / 1024)
+    }, payload.video);
+  }
+
+  const maxTracks = bandwidthProfile.video.maxTracks;
+  if (typeof maxTracks === 'number') {
```
makarandp0: same thing - not sure what `Object.assign` is needed.

lib/util/index.js
```diff
@@ -520,6 +520,40 @@ function valueToJSON(value) {
...
+ */
+function createBandwidthProfilePayload(bandwidthProfile) {
+  const payload = {};
+  if (typeof bandwidthProfile.video !== 'object') {
+    return payload;
+  }
+
+  const mode = bandwidthProfile.video.mode;
+  if (typeof mode === 'string') {
+    payload.mode = mode;
```
makarandp0: do we need extra validation on valid values `mode ` for this ? I am okay with not having more validations, but just wondering.

test/unit/spec/signaling/v2/twilioconnectiontransport.js
```diff
@@ -12,8 +12,40 @@ const TwilioError = require('../../../../../lib/util/twilioerror');
 const { combinations } = require('../../../../lib/util');
 
 describe('TwilioConnectionTransport', () => {
-  combinations([[true, false], [true, false], [true, false]]).forEach(([networkQuality, dominantSpeaker, automaticSubscription]) => {
-    describe(`constructor, called with .networkQuality flag ${networkQuality ? 'enabled' : 'disabled'}, .dominantSpeaker flag ${dominantSpeaker ? 'enabled' : 'disabled'} and .automaticSubscription flag ${automaticSubscription ? 'enabled' : 'disabled'}`, () => {
+  combinations([
+    [true, false],
```
makarandp0: not sure if its worth trying various combinations of networkQuality, dominentSpeaker and all such values. maybe breaking up this test would make it more readable. Not a big concern though, as its just a unit test.

lib/util/index.js
manjeshbhargav: @makarandp0 This would overwrite other properties in `payload.video`.

lib/util/index.js
manjeshbhargav: @makarandp0 See my comment above.

test/unit/spec/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 It helps to be thorough in unit tests, since integration tests are more costly.

lib/util/index.js
makarandp0: but given that  payload.video is `undefined` before this assignment we do not have to worry about its existing values. IMO Something like this would look simpler. But now I kind of understand your motivation to use `Object.assign`
```
payload.video = {};
if (typeof bandwidthProfile.video.maxSubscriptionBitrate === 'number') {
  payload.video.max_subscription_bandwidth =  Math.round(bandwidthProfile.video.maxSubscriptionBitrate / 1024)
}

if (typeof bandwidthProfile.video.maxTracks === 'number') {
  payload.video.maxTracks = bandwidthProfile.video.maxTracks;
}
```

lib/util/index.js
manjeshbhargav: @makarandp0 I don't think so. We can add validation once we have the latest RSP spec implemented by Room Service.

lib/util/index.js
manjeshbhargav: @makarandp0 In your example, The `.video` property is created even when both `maxSubscriptionBitrate` and `maxTracks` are not present, which shouldn't happen. So, using `Object.assign()` is the most straight-forward way.

PR #686, opened 2019-07-05
JSDK-2400 invalid characters in region parameter should return correct errorcode.

makarandp0:
Instead of using region specified directly, we now `encodeURIComponent` the region, This ensures that if `region` contains invalid characters for URI, it will fail with correct error code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta13 (in progress)
+============================
+
+Bug Fixes
+---------
+
+- Instead of using region specified directly, we now `encodeURIComponent` the region,
+This ensures that if `region` contains invalid characters for URI, it will fail with
+correct error code.
+
 2.0.0-beta12 (July 12, 2019)
 ============================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
 2.0.0-beta13 (in progress)
-============================
+==========================
 
 Bug Fixes
 ---------
 
-- Instead of using region specified directly, we now `encodeURIComponent` the region,
-This ensures that if `region` contains invalid characters for URI, it will fail with
-correct error code.
+- Fixed a bug where connecting to a Room with a `region` containing special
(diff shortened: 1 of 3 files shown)


test/integration/spec/connect.js
```diff
@@ -146,6 +146,35 @@ describe('connect', function() {
...
+    let sid;
+    let token;
+    beforeEach(async () => {
+      const identity = randomName();
+      token = getToken(identity);
+      sid = await createRoom(randomName(), defaults.topology);
+    });
+
+    // few examples of invalid regions.
+    ['invalid', '34', '$abc', 'abc!'].forEach((region) => {
```
manjeshbhargav: @makarandp0 Instead of a separate test here, I would include them in the existing test as follows:
```js
const invalidRegions = ['foo', '34', '$abc', 'abc!'];

const validRegions = defaults.regions ? defaults.regions.split(',') : [
  'au1', 'br1', 'de1', 'ie1', 'in1', 'jp1', 'sg1', 'us1', 'us2'
];

const regions = ['without', 'gll', ...invalidRegions, ...validRegions];

regions.forEach(region => {
  const regionOptions = region === 'without' ? {} : { region };
  const scenario = `when called ${region === 'without'
    ? 'without a region'
    : `with a region that is ${invalidRegions.includes(region) ? 'invalid' : 'valid'} (${region})`};

  ...

  it(`should return a CancelablePromise that ${invalidRegions.includes(region)' ? 'rejects with a SignalingConnectionError' : 'resolves with a Room'}`, async () => {...});
});
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta13 (in progress)
+============================
+
+Bug Fixes
+---------
+
+- Instead of using region specified directly, we now `encodeURIComponent` the region,
```
manjeshbhargav: @makarandp0 
```
- Fixed a bug where connecting to a Room with a `region` containing special
  characters in ConnectOptions failed with an Error other than
  [SignalingConnectionError](https://www.twilio.com/docs/api/errors/53000). (JSDK-2400)
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta13 (in progress)
+============================
```
manjeshbhargav: @makarandp0 Nitpick: remove the last two `==`.

PR #690, opened 2019-07-16
JSDK-2440 dscp tagging option experiment

makarandp0:
This change is only for demonstration purpose. It sets `chromeSpecificConstraints` to `{"optional":[{"googDscp":true}]}` 

I was hoping to DSCP field set with this change. Although I do see that chrome understands this parameter (as seen from chrome://webrtc-internals, I do not see any difference in DSCP header as reported by wireshark.


makarandp0:
this was just an experiment - real PR is opened as: https://github.com/twilio/twilio-video.js/pull/704

PR #693, opened 2019-07-18
Handle Firefox 68+'s empty ice candidate in Chrome and Safari.

manjeshbhargav:
@makarandp0 

This PR fixes JSDK-2412 and JSDK-2415. The reason for these bugs is that Firefox 68+ generates an RTCIceCandidate with an empty candidate string to signal the end of candidate gathering. It also generates a null candidate (legacy way of signaling end of candidate gathering). But Chrome and Safari do not yet accept the empty candidate. So, we now just log warnings and suppress the error since it does not affect media connection in any way.

The reason I wanted to log warnings is that we should be able to make out when Chrome and Safari start accepting the empty candidate. When they do, we should no longer see these console warnings.

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
 Bug Fixes
 ---------
 
+- Worked around a minor interop issue between Chrome/Safari Participants and Firefox 68+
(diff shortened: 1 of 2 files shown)


manjeshbhargav:
@makarandp0 I don't think we need a unit test for this. I was able to confirm this using Simpler Signaling.

PR #696, opened 2019-07-22
JSDK-2448 Type checks for TrackPriority and BandwidthProfileOptions

manjeshbhargav:
@makarandp0 

This PR adds type checking logic to make sure only valid values are specified for TrackPriority and BandwidthProfileOptions.

TODO
=====

- [x] Add CHANGELOG.md for Bandwidth Profile API.

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,34 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
 New Features
 ------------
 
-- In a **Group Room**, you can now have more control over how your available bandwidth
-  is distributed among the RemoteVideoTracks that you have subscribed to. twilio-video.js
-  introduces the [Bandwidth Profile API](TODO(mmalavalli)). Note that this feature is
+- In a **Group Room**, you can now control how your available downlink bandwidth is
+  distributed among the RemoteVideoTracks that you have subscribed to. twilio-video.js
+  introduces the [Bandwidth Profile APIs](TODO(mmalavalli)). Note that this feature is
   currently in **private beta** and hence will be **opt-in**. Please reach out to
   [email@address.com](TODO(mmalavalli)) for more information about how to enable these
   APIs for your Twilio Account. **Using these APIs in a Peer-to-Peer Room will have no effect**.
-  
+
+  ### Bandwidth Profile
+
+  You can now configure how your available downlink bandwidth will be distributed
+  among your subscribed RemoteVideoTracks by using a new optional ConnectOptions
+  parameter `bandwidthProfile`. For more details, please refer to the `BandwidthProfileOptions`
+  [documentation](//media.twiliocdn.com/sdk/js/video/releases/2.0.0-beta13/docs/global.html#BandwidthProfileOptions).
+  Here is an example:
+
+  ```js
+  const { connect } = require('twilio-video');
+  const room = await connect(token, {
+    bandwidthProfile: {
+      video: {
+        maxSubscriptionBitrate: 150000, // Max. bandwidth (bps) to be allocated to subscribed RemoteVideoTracks.
+        maxTracks: 3, // Max. number of visible RemoteVideoTracks. Other RemoteVideoTracks will be switched off.
+        mode: 'collaboration', // Subscription mode: "collaboration", "grid" or "presentation".
+      }
+    }
+  });  
+  ```
+
   ### Track Priority
 
   While publishing a LocalTrack, you can now optionally specify its publish priority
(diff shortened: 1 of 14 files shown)


lib/connect.js
```diff
@@ -593,4 +598,32 @@ function normalizeVideoCodecSettings(nameOrSettings) {
   }
 }
 
+function validateBandwidthProfile(bandwidthProfile) {
+  if (typeof bandwidthProfile !== 'object' || Array.isArray(bandwidthProfile)) {
```
makarandp0:
```suggestion
  if (bandwidthProfile === null || typeof bandwidthProfile !== 'object' || Array.isArray(bandwidthProfile) ) {
```

also since we are using such check in lots of places, lets add `isObject` function in utils and use it here.

lib/connect.js
```diff
@@ -593,4 +598,32 @@ function normalizeVideoCodecSettings(nameOrSettings) {
...
+    return E.INVALID_TYPE('options.bandwidthProfile', 'object');
+  }
+  if (!('video' in bandwidthProfile)) {
+    return null;
+  }
+  const video = bandwidthProfile.video;
+  if (typeof video !== 'object' || Array.isArray(video)) {
```
makarandp0: same as last suggestion - need to add check against null, or use utils function

lib/localparticipant.js
```diff
@@ -371,6 +375,12 @@ class LocalParticipant extends Participant {
       return Promise.reject(error);
     }
 
+    const priorityValues = Object.values(trackPriority);
+    if (typeof options.priority !== 'string' || !priorityValues.includes(options.priority)) {
```
makarandp0: ist part of the check is not necessary as `includes` cover it. 
```suggestion
    if (!priorityValues.includes(options.priority)) {
```

lib/util/index.js
```diff
@@ -278,6 +278,15 @@ function delegateMethods(source, wrapper, target) {
   }
 }
 
+/**
+ * Whether the given argument is a non-array object.
+ * @param {*} object
+ * @return {boolean}
+ */
+function isNonArrayObject(object) {
```
makarandp0: Most of the times, we would want to also exclude `null` value as an object. It would be useful if we folded null check inside this function as well. 

not sure what would you call the function. how about `isSimpleObject` 
```
function isSimpleObject(subject) {
   return subject !== null && typeof subject === 'object' && !Array.isArray(subject);
}
```

manjeshbhargav: @makarandp0 I have used this method in places where `null` is a valid value. So, for now, I think I'm more comfortable with the null check being outside the function.

CHANGELOG.md
```diff
@@ -1,5 +1,91 @@
...
+
+  ```js
+  remoteParticipant.on('trackPublished', remoteTrackPublication => {
+    console.log(`RemoteParticipant published a Track with priority ${remoteTrackPublication.publishPriority}`);
+  });
+  ```
+
+  ### Switching on/off RemoteVideoTracks
+
+  When a subscribing Participant's bandwidth is considerably low, the media server
```
aymenn: maybe change "is considerably low" to something like "is insufficient"

CHANGELOG.md
```diff
@@ -1,5 +1,91 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta13 (in progress)
+==========================
+
+New Features
+------------
+
+- In a **Group Room**, you can now have more control over how your available bandwidth
```
aymenn: For readability, "you can now have more control over" -> "you can control"

CHANGELOG.md
```diff
@@ -1,5 +1,91 @@
...
+    });
+
+    remoteTrack.on('switchedOn', () => {
+      assert.equal(remoteTrack.isSwitchedOff, false);
+      console.log(`The RemoteTrack ${remoteTrack.name} was switched on`);
+    });
+  });
+  ```
+
+  ### Bandwidth Profile
```
aymenn: I think make this the first item (before track prio). This is the "easy" API and is used during connect.

CHANGELOG.md
```diff
@@ -1,5 +1,91 @@
...
+  const localTrackPublication = await room.localParticipant.publishTrack(localTrack, {
+    priority: 'high' // LocalTrack's publish priority - "low", "standard" or "high"
+  });
+
+  // LocalTrackPublication has a new property "priority" which stores the publish
+  // priority set while publishing the corresponding LocalTrack.
+  assert.equal(localTrackPublication.priority, 'high');
+  ```
+
+  This signals to the media server that this LocalTrack is more important than other
```
lwilio: The track priority can be used to indicate that a track is more important and also less important. I would not constraint the explanation to the 'high' case. In addition, the current explanation seems to make understand that priorities work "per-participant". Hence, I would be saying instead something like "This signals to the media server the relevance of a track with respect to other tracks published to the Room"

test/unit/spec/media/track/remotedatatrack.js
```diff
@@ -53,6 +57,62 @@ describe('RemoteDataTrack', () => {
...
 
+  describe('#_setSwitchedOff', () => {
+    [
+      [true, true],
+      [true, false],
+      [false, true],
+      [false, false]
+    ].forEach(([isSwitchedOff, newIsSwitchedOff]) => {
+      context(`when .isSwitchedOff is ${isSwitchedOff} and the new value is ${newIsSwitchedOff}`, () => {
+        let arg;
```
makarandp0:
```suggestion
        let arg = null
```

test/unit/spec/media/track/remotedatatrack.js
```diff
@@ -53,6 +57,62 @@ describe('RemoteDataTrack', () => {
...
+        let track;
+        let trackSwitchedOff;
```
makarandp0:
```suggestion
        let trackSwitchedOff = false;
```

manjeshbhargav:
@makarandp0 All your suggestions are great, but for now I want to follow the pattern in the tests so that we don't have any inconsistencies. We should keep this in mind once we get a chance to refactor our tests.

PR #698, opened 2019-07-29
Prepping for 2.0.0-rc35

manjeshbhargav:
@makarandp0 

This PR makes sure that we don't send the `bandwidth_profile` RSP payload if `bandwidthProfile` is not specified in ConnectOptions.

## TODO

- [x] Document `maxSubscriptionBitrate` behavior for values `<= 0`.
- [x] Modify CHANGELOG.md entry if required.

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -130,7 +130,6 @@ function connect(token, options) {
   options = Object.assign({
     abortOnIceServersTimeout: false,
     automaticSubscription: true,
-    bandwidthProfile: {},
     createLocalTracks,
     dominantSpeaker: false,
     environment: constants.DEFAULT_ENVIRONMENT,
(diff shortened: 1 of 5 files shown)


lib/connect.js
```diff
@@ -605,7 +604,10 @@ function normalizeVideoCodecSettings(nameOrSettings) {
 }
 
 function validateBandwidthProfile(bandwidthProfile) {
-  if (bandwidthProfile === null || !isNonArrayObject(bandwidthProfile)) {
+  if (typeof bandwidthProfile === 'undefined' || bandwidthProfile === null) {
```
makarandp0: I am debating if  we should allow explicit `null` and `undefined` as valid values. Not specifying bandwidthProfile is one thing, but if specified maybe we should require valid object.

Do we have other properites that allow  `{}` as well as `undefined` but they behave differently ?

manjeshbhargav: @makarandp0 I think not specifying the property, setting it to null or undefined is essentially the same thing. We do have `EncodingParameters` which is passed to `LocalParticipant.setParameters` where passing an `{}` does not change existing behavior.

makarandp0: @manjeshbhargav  in this case  caller would get different behavior when they pass  `{}` (default bwprofile) v/s `null` or `undefined` (no bandwidth profile). I do not have any concerns with `{}` - but just wondering what is the reason to accept `null` and `undefined` values.

manjeshbhargav: @makarandp0 The `undefined` check is not for explicitly set `undefined` value, since we filter those props out. It is to check whether the property exists. I will change the code to reject `null` in order to remove any confusion.

PR #705, opened 2019-08-01
JSDK-2463 Undoing updates to "_trackIdsToAttributes" when an offer is rolled back.

manjeshbhargav:
@makarandp0 

This PR consumes https://github.com/twilio/twilio-webrtc.js/pull/109 and makes sure updates to "_trackIdsToAttributes" are undone when an offer is rolled back.

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -614,9 +614,7 @@ class PeerConnectionV2 extends StateMachine {
   _handleSignalingStateChange() {
     if (this._peerConnection.signalingState === 'stable') {
       this._appliedTrackIdsToAttributes = new Map(this._trackIdsToAttributes);
-      return;
-    }
-    if (this._peerConnection.signalingState === 'closed' && this.state !== 'closed') {
+    } else if (this._peerConnection.signalingState === 'closed' && this.state !== 'closed') {
       this.preempt('closed');
     }
   }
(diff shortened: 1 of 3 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -606,6 +612,10 @@ class PeerConnectionV2 extends StateMachine {
    * @returns {void}
    */
   _handleSignalingStateChange() {
+    if (this._peerConnection.signalingState === 'stable') {
+      this._appliedTrackIdsToAttributes = new Map(this._trackIdsToAttributes);
+      return;
+    }
```
makarandp0: nit: suggest use `else if` and remove return from the previous block

lib/signaling/v2/peerconnection.js
```diff
@@ -606,6 +612,10 @@ class PeerConnectionV2 extends StateMachine {
    * @returns {void}
    */
   _handleSignalingStateChange() {
+    if (this._peerConnection.signalingState === 'stable') {
+      this._appliedTrackIdsToAttributes = new Map(this._trackIdsToAttributes);
```
makarandp0: just curious - why update  this._appliedTrackIdsToAttributes here, instead of saving them before calling `_setSimulcast` in offer and answer?

manjeshbhargav: @makarandp0 We only update the map after both offer and answer are applied. If we have to roll back the offer, then we would just discard the changes in `_trackIdsToAttributes`.

PR #708, opened 2019-08-02
JSDK-2441 - dominantSpeakerPriority and renderDimensions

manjeshbhargav:
@makarandp0 

_WIP_

## Changes

- [x] Adding dominantSpeakerPriority + unit tests.
- [x] Adding integration tests for dominantSpeakerPriority.
- [x] Adding renderDimensions + unit tests.
~~Adding integration tests for renderDimensions.~~ - *This only affects bandwidth allocation and cannot be deterministically tested*.

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -371,8 +371,11 @@ function connect(token, options) {
  * bandwidth is shared among the {@link RemoteVideoTrack}s you have subscribed to in a Group Room.
  * @typedef {object} VideoBandwidthProfileOptions
  * @property {Track.Priority} [dominantSpeakerPriority="standard"] - Optional parameter to
- *   specify the default subscribe priority of the Dominant Speaker's {@link RemoteVideoTrack}s.
- *   This defaults to "standard".
+ *   specify the minimum subscribe {@link Track.Priority} of the Dominant Speaker's {@link RemoteVideoTrack}s.
+ *   This means that the Dominant Speaker's {@link RemoteVideoTrack}s that are published with
+ *   lower {@link Track.Priority} will be subscribed to with the {@link Track.Priority} specified here.
+ *   This has no effect on {@link RemoteVideoTrack}s published with higher {@link Track.Priority}, which will
+ *   still be subscribed to with with the same {@link Track.Priority}. If not specified, this defaults to "standard".
  * @property {number} [maxSubscriptionBitrate] - Optional parameter to specify the maximum
  *   downlink video bandwidth in bits per second (bps). By default, there are no limits on
  *   the downlink video bandwidth.
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -382,6 +382,28 @@ function connect(token, options) {
  * @property {BandwidthProfileMode} [mode="grid"] - Optional parameter to specify how the {@link RemoteVideoTrack}s'
  *   TrackPriority values are mapped to bandwidth allocation in Group Rooms. This defaults to "grid",
  *   which results in equal bandwidth share allocation to all {@link RemoteVideoTrack}s.
+ * @property {VideoRenderDimensions} [renderDimensions] - Optional parameter to specify the desired
+ *   render dimensions of {@link RemoteVideoTrack}s based on {@link Track.Priority} and the
+ *   {@link RemoteVideoTrack}s of the Dominant Speaker.
+ */
+
+/**
+ * {@link VideoRenderDimensions} allows you to specify the desired render dimensions of {@link RemoteVideoTrack}s
+ * based on {@link Track.Priority} and the {@link RemoteVideoTrack}s of the Dominant Speaker. The bandwidth allocation
+ * algorithm will distribute the available downlink bandwidth proportional to the requested render dimensions.
+ * @typedef {object} VideoRenderDimensions
+ * @property {VideoTrack.Dimensions} [dominantSpeaker] - Optional parameter to specify the desired rendering dimensions
+ *   of the Dominant Speaker's {@link RemoteVideoTrack}s. 0 or a negative value will result in the lowest possible
+ *   resolution. This defaults to 640 x 480 (VGA).
+ * @property {VideoTrack.Dimensions} [high] - Optional parameter to specify the desired rendering dimensions of
+ *   {@link RemoteVideoTrack} whose {@link Track.Priority} is "high". 0 or a negative value will result in the lowest
+ *   possible resolution. This defaults to 1280 x 720 (HD).
+ * @property {VideoTrack.Dimensions} [low] - Optional parameter to specify the desired rendering dimensions of
+ *   {@link RemoteVideoTrack} whose {@link Track.Priority} is "low". 0 or a negative value will result in the lowest
+ *   possible resolution. This defaults to 176 x 144 (QCIF).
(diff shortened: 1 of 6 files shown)


lib/util/index.js
```diff
@@ -535,32 +535,21 @@ function valueToJSON(value) {
...
+  return typeof bandwidthProfile.video === 'object' ? [
+    ['dominantSpeakerPriority', 'string', 'active_speaker_priority'],
+    ['maxSubscriptionBitrate', 'number', 'max_subscription_bandwidth', x => Math.round(x / 1024)],
+    ['maxTracks', 'number', 'max_tracks'],
+    ['mode', 'string'],
+    ['renderDimensions', 'object', 'render_dimensions', createRenderDimensionsPayload]
+  ].reduce((payload, [prop, dataType, payloadProp = prop, transform = x => x]) => {
+    const value = bandwidthProfile.video[prop];
+    if (typeof value === dataType) {
+      payload.video = Object.assign({
```
makarandp0:
```suggestion
      payload.video[payloadProp] = transform(value)
```

lib/util/index.js
```diff
@@ -591,6 +580,44 @@ function createMediaSignalingPayload(dominantSpeaker, networkQuality, trackSwitc
...
+/**
+ * Create {@link VideoTrack.Dimensions} RSP payload.
+ * @param {VideoTrack.Dimensions} [dimensions]
+ * @returns {object}
+ */
+function createDimensionsPayload(dimensions) {
+  return ['height', 'width'].reduce((payload, dimension) => {
+    const value = dimensions[dimension];
+    if (typeof value === 'number') {
+      payload = Object.assign({
```
makarandp0:
```suggestion
      payload[dimension] = value;
```

lib/util/index.js
```diff
@@ -591,6 +580,44 @@ function createMediaSignalingPayload(dominantSpeaker, networkQuality, trackSwitc
...
+ */
+function createRenderDimensionsPayload(renderDimensions) {
+  return [
+    ['dominantSpeaker', 'active_speaker']
+  ].concat(Object.values(constants.trackPriority).map(priority => {
+    return [priority];
+  })).reduce((payload, [prop, payloadProp = prop]) => {
+    const value = renderDimensions[prop];
+    if (typeof value === 'object') {
+      payload = Object.assign({
```
makarandp0:
```suggestion
      payload[payloadProp] = createDimensionsPayload(value);
```

lib/util/index.js
```diff
@@ -591,6 +580,44 @@ function createMediaSignalingPayload(dominantSpeaker, networkQuality, trackSwitc
...
+}
+
+/**
+ * Create {@link VideoRenderDimensions} RSP payload.
+ * @param renderDimensions
+ * @returns {object}
+ */
+function createRenderDimensionsPayload(renderDimensions) {
+  return [
+    ['dominantSpeaker', 'active_speaker']
```
makarandp0: since `dominantSpeaker` is the only one that has `payloadProp != prop`  - I would take it out and make the rest of the processing bit easier to read.

lib/connect.js
```diff
@@ -633,7 +658,34 @@ function validateBandwidthProfile(bandwidthProfile) {
...
+}
+
+function validateDimensions(prop, dimensions) {
+  if (typeof dimensions === 'undefined') {
+    return null;
+  }
+  if (dimensions === null || !isNonArrayObject(dimensions)) {
+    return E.INVALID_TYPE(`options.bandwidthProfile.video.renderDimensions.${prop}`, 'object');
+  }
+  return ['height', 'width'].reduce((error, dimension) => {
```
makarandp0: I think using reduce for two properties is overkill. Would be easier to read w/o reduce.

```
if (typeof dimension.height !== `number`) {
   return E.INVALID_TYPE(`options.bandwidthProfile.video.renderDimensions.${prop}.height`, 'number');
}
if (typeof dimension.width !== `number`) {
   return E.INVALID_TYPE(`options.bandwidthProfile.video.renderDimensions.${prop}.width`, 'number');
}
```

lib/connect.js
```diff
@@ -633,7 +658,34 @@ function validateBandwidthProfile(bandwidthProfile) {
...
+    return typeof dimensions[dimension] === 'number' ? null
+      : E.INVALID_TYPE(`options.bandwidthProfile.video.renderDimensions.${prop}.${dimension}`, 'number');
   }, null);
 }
 
+function validateRenderDimensions(renderDimensions) {
+  if (typeof renderDimensions === 'undefined') {
+    return null;
+  }
+  if (renderDimensions === null || !isNonArrayObject(renderDimensions)) {
```
makarandp0: I see that pretty much all usages of `isNonArrayObject` are preceded by checking for null. Consider moving it inside the function.

lib/util/index.js
manjeshbhargav: @makarandp0 Then we would have to add `payload.video = payload.video || {}` everywhere.

lib/util/index.js
manjeshbhargav: @makarandp0 See my previous comment.

lib/util/index.js
manjeshbhargav: @makarandp0 I think this is fairly easy to read because we've set the initial value of `payloadProp` to `prop` in the `reduce` callback, which means that if the object prop and RSP prop are the same, then `payloadProp` will be equal to `prop`.

lib/util/index.js
manjeshbhargav: @makarandp0 Then we would have to add payload.video = payload.video || {} everywhere.

lib/connect.js
manjeshbhargav: @makarandp0 We use `isNonArrayObject` in other places where `null` is a valid parameter.

lib/util/index.js
```diff
@@ -603,18 +630,26 @@ function createDimensionsPayload(dimensions) {
...
+  ]);
+}
+
+/**
+ * Create an RSP payload for the given object.
+ * @param {object} object
+ * @param {Array<object>} propConversions
+ * @returns {*}
+ */
+function createRSPPayload(object, propConversions) {
```
makarandp0: nice refactor, but can use more beefed up description and documentation on parameters

makarandp0: nit: rename `object` to `inputPayload`

test/unit/spec/signaling/v2/twilioconnectiontransport.js
```diff
@@ -20,7 +20,7 @@ describe('TwilioConnectionTransport', () => {
     [              // bandwidthProfile
       [undefined],
       [{}, {}],
-      [{ video: {} }, {}],
+      [{ video: {} }, { video: {} }],
```
makarandp0: looks like this is producing different output than last revision - is  this intentional ?  `{ video: {} }` => `{}` seems okay too.

lib/util/index.js
manjeshbhargav: @makarandp0 `payload` implies RSP, where as the object is part of ConnectOptions.

test/unit/spec/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 I did this to maintain consistency between the options object and RSP.

PR #709, opened 2019-08-07
generate and upload integration test results from Travis CI

makarandp0:
This change 
a) adds plugins to generate html and junit results from integration tests
b) instructs Travis CI to upload results  to S3. 

The build results would look something like [this](https://twilio-video-logs.s3.amazonaws.com/twilio/twilio-video.js/1835/1835.4/2019-08-07T191404423Z/results.html). Note: I have chose to only show failed tests to avoid noise.

These results will be uploaded only for triggered or scheduled builds.  It does not support the uploading artifacts for pull request builds ( https://github.com/travis-ci/travis-ci/issues/5579 and https://docs.travis-ci.com/user/uploading-artifacts/ )  

also Travis CI does not provide way to send file to email/slack ( https://travis-ci.community/t/sending-files-as-attachment-on-notifications/1259 ) 
 

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #712, opened 2019-08-12
JSDK-2469: run integration tests inside container

makarandp0:
This change runs integration tests inside docker containers. It also generates structured report files for the tests and asks travis to upload to S3. The S3 upload does not work for pull request builds. But works for scheduled builds and kicked off build.

one such build in Progress: https://travis-ci.org/twilio/twilio-video.js/builds/571064856 
 
1. DockerFile and Docker-compose.yml file
2. .travis.yml - Update to to enable docker, and execute tests in container
3.  karma/makeconf.js - 
Changes to update browser launch options when running in docker.
Changes to add structured reporting for test results.
Additional option process.env.DEBUG that can help debug by preserving line numbers, and keeping browser open after failures for further inspection. 
4. scripts/karma.js
Changes to reliably make `FILE:foo npm run test:integration` work 

This change folds in changes from https://github.com/twilio/twilio-video.js/pull/709 - Generate and upload build results to S3

Some results from above job 

Firefox Disconnect failures: [1](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.8/testresults-BROWSER-firefox-BVER-stable-TOPOLOGY-group-0.html), [2](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.7/testresults-BROWSER-firefox-BVER-stable-TOPOLOGY-peer-to-peer-0.html)

Successes: [1](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.1/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-peer-to-peer-2.html), [2](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.1/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-peer-to-peer-1.html), [3](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.2/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-group-1.html), [4](https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.2/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-group-2.html)

**Contributing to Twilio**


.travis.yml
```diff
@@ -80,31 +83,16 @@ matrix:
...
+  - docker-compose run test npm run test:integration
 
 notifications:
   slack:
     on_pull_requests: false
     secure: jxmuMo2n/OCaUg6DMZ/fahcxoCmkyOt4eD5pWgIqinxL0INyByZhw3EyAItueuXo4+7oh2KZj78ryf2bChSMsGGXXF25Vdy154bGsFXS7xPH/ars7yTo2Cjry8j3ki0ve7aqH3GjMlBVuJRA3ps2w/bJjw9oq4JDiibDRsDPawcS3Xda0tHqc022lJer9jHaF+zBrjobo2xAZ8dFhMmx50wsY5tALq5xoiiAAfmARm8ZIZu6hZpwdbpC6pM9SBd1RNMxnzQrlFPqA25FYPZSaD4/5wjKDEHTO+dfOEW79sgZUuohE/AC+zKd3wiYxxDaxrkPkzijmYs39LyDqm5Vr6eS9OSrZptgGni6G/2J/dBW52MLlxCDB+SqpbaPuLVKsvPdVAwydMTYo02fJpxa3purUdNM9GFPZCZC4KGvht2cSvM1txUNFkpkAp8RCxyUGpVw58S4gnAp2sYmRR22i42nZJunR7myXiMV+KZcaS7BJ2Mf6Wo1YwSnHSP4xwuaBPBY0L86w/gyd9tABpEIamOwKkFiLmOKr1l8aCcuGrEHnBB2L5EpiNfxl93DjrApzwR3FqCeSzfwHgaZ6EIGtRqPPXUJgM4DtYt9udGO4c9fLr6QbDgyS4swn3uUiUmeI3i9UUg92jvv7OCYoe80DfsF5RVjZxXcQM+WMMsJ7os=
+
+addons:
+  artifacts: true
+
```
manjeshbhargav: @makarandp0 Nitpick: remove extra line.

karma/makeconf.js
```diff
@@ -1,26 +1,32 @@
...
+  let files = [];
+  if (config.files && config.files.length) {
+    files = config.files;
+  } else if (defaultFile) {
+    files = [defaultFile];
+  }
+  return files;
+}
+
+let testRun = 0;
```
manjeshbhargav: @makarandp0 Nitpick: add newline between variable and function.

.env
```diff
@@ -0,0 +1,6 @@
+BROWSER=chrome
```
manjeshbhargav: @makarandp0 Is this file necessary to be committed, since you're adding it to `.gitignore`?

docker/test/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+# It's a good idea to use dumb-init to help prevent zombie chrome processes.
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
+RUN chmod +x /usr/local/bin/dumb-init
+
+ADD ./ /opt/app
+WORKDIR /opt/app
+RUN npm install
+
+# Build Args
+ARG USE_BROWSER='firefox'
```
manjeshbhargav: @makarandp0 Why are we hard-coding this value?

docker/test/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+ARG USE_BVER='unstable'
```
manjeshbhargav: @makarandp0 Why are we hard-coding this value?

docker/test/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+
+ENV ENABLE_REST_API_TESTS=1
```
manjeshbhargav: @makarandp0 Why are we hard-coding this value? This should be enabled only for non peer-to-peer rooms.

karma/makeconf.js
```diff
@@ -29,12 +35,26 @@ function makeConf(defaultFile, browserNoActivityTimeout, requires) {
...
       }
+    } else if (isDocker) {
+      browsers = ['ChromeInDocker', 'FirefoxInDocker'];
     } else if (process.platform === 'darwin') {
       browsers = ['ChromeWebRTC', 'ElectronWebRTC', 'FirefoxWebRTC', 'Safari'];
     } else {
       browsers = ['ChromeWebRTC', 'FirefoxWebRTC'];
     }
 
+    const strTestRun = (testRun++).toString();
```
manjeshbhargav: @makarandp0 Do we need to append the test run number? I think the combination of browser/bver/topology would be unique enough no? Also, I think we should enable test reporting only for docker container. I don't want to push reports when I run integration tests locally.

karma/makeconf.js
```diff
@@ -46,19 +66,48 @@ function makeConf(defaultFile, browserNoActivityTimeout, requires) {
...
+        debug: !!process.env.DEBUG,
         transform: [
           'envify'
         ]
       },
-      reporters: ['spec'],
+      reporters: ['spec', 'junit', 'html'],
+      htmlReporter: { // configuration for karma-htmlfile-reporter
+        outputFile: htmlReport,
+        pageTitle: 'Twilio-Video Integration Tests',
```
manjeshbhargav: @makarandp0 Nitpick: `s/Twilio-Video/twilio-video.js`

karma/makeconf.js
manjeshbhargav: @makarandp0 Maybe instead of `testRun`, we can use the Travis build number if we have access to it?

scripts/karma.js
```diff
@@ -10,6 +10,9 @@ const configFile = join(__dirname, '..', process.argv[2]);
 const integrationTests = join(__dirname, '..', 'test', 'integration', 'spec');
 
 function getTestPaths(path) {
+  if (process.env.FILE) {
```
manjeshbhargav: @makarandp0 I'm not entirely clear why this is added?

.env
makarandp0: This file is optional - but  helps with local runs, so I am committing it as a recommended template.  I added it gitignore so that git won't track any local changes to this file.

docker/test/Dockerfile
makarandp0: it's default value for args and not strictly necessary. It will be overwritten by the parameter passed at build time - and would then be baked into the docker image.

The idea is each docker image is built for specific combination of BROWSER+version 
the BROWSER and version to use is determined by the parameter. If you forget to pass the browser+version, it would default to these values. 

Docker-compose passes these arguments at build time by looking at your env variables. 
https://github.com/twilio/twilio-video.js/blob/JSDK-2469_dockerized_tests/docker-compose.yml#L13

but somebody were to use this DockerFile directly (manually) then its good to have some defaults (instead of building image w/o browser)

docker/test/Dockerfile
makarandp0: same as above

docker/test/Dockerfile
makarandp0: I didn't find a nice way to conditionally declare this parameter...but we are okay here, because the test themselves skip if the build is not using p2p rooms. https://github.com/twilio/twilio-video.js/blob/JSDK-2469_dockerized_tests/test/integration/spec/rest.js#L36

karma/makeconf.js
makarandp0: unfortunately, the combination of browser+bver+topology is not unique, as we have some custom logic to split tests per file to workaround some memory leaks (https://github.com/twilio/twilio-video.js/blob/JSDK-2469_dockerized_tests/scripts/karma.js#L32). 

I considered using date/time stamp instead of the testRun, but decided against it, as I found simple suffixes makes more deterministic and more readable 

Travis build number does not help either, as they are part of the same travis build (btw build is already encoded in the uri as seen in example files below) 

https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.2/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-group-2.html
https://s3.amazonaws.com/twilio-video-logs/twilio/twilio-video.js/1848/1848.2/testresults-BROWSER-chrome-BVER-stable-TOPOLOGY-group-1.html

scripts/karma.js
makarandp0: We had a similar check earlier in https://github.com/twilio/twilio-video.js/blob/master/karma/makeconf.js#L9 which did not quite work for some files. Moving it at this level makes it works better.

karma/makeconf.js
makarandp0: about pushing reports - the reports are generated all the time, but pushed to S3 by Travis only for CI runs. They don't get pushed for local runs.

.env
manjeshbhargav: @makarandp0 Then I think we should rename it to `.env.template`.

.gitignore
```diff
@@ -13,5 +13,6 @@ npm-debug.log
 package-lock.json
 test.json
 yarn.lock
+.env
```
syerrapragada: Nitpick: .env -> .env.template

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: '3'
+# to run test in docer:
```
syerrapragada: Nitpick: docer -> docker

docker/test/Dockerfile
syerrapragada: Default in Karma config is chrome. Should we default this to chrome as well?

karma/makeconf.js
```diff
@@ -71,6 +121,19 @@ function makeConf(defaultFile, browserNoActivityTimeout, requires) {
...
+          base: 'Firefox',
+          flags: [
+            '-headless',
+          ],
+          prefs: {
+            'media.gstreamer.enabled': false,
+            'media.navigator.permission.disabled': true,
+            'media.navigator.streams.fake': true,
+            'media.autoplay.enabled.user-gestures-needed': false,
+            'media.block-autoplay-until-in-foreground': false
```
syerrapragada: You may want to consider setting following 2 values as well:
'media.getusermedia.insecure.enabled'
'media.devices.insecure.enabled'

Future builds of FF does not support connecting to insecure URLs, I assume that including localhost.

docker/test/Dockerfile
makarandp0: Its best to setup env variables/.env files. I intentionally kept defaults to atypical  ff/unstable so that the developer does not forget them. ( tests are more likely to fail in firefox/unstable :)

karma/makeconf.js
makarandp0: Thanks, I will experiment with those and add them if they work well.

.gitignore
makarandp0: @syerrapragada , since we expect everybody to have their version of `.env` - .env needs to go in `.gitignore` to not include track any changes to it. We do want to track any changes to .env.template!

PR #720, opened 2019-08-20
JSDK-2481 migrate to json config for eslit + lint clean

makarandp0:
This change updates our eslint config to latest version + marks all existing issues so that we are lint clean. This will help us 
a) catch any newer issues. 
b) ensure that modern tools work well with our codebase.

Note: I have made line based exception for existing innocent violations instead of fixing them. I have also disabled 'no-undefined' rule for test code as there were just too many such instances.
  
Let's plan to stay lint clean completely.

~Update: This change also addresses a regression caused when I moved the integration tests to docker ( https://github.com/twilio/twilio-video.js/pull/712 ) That change caused some of the tests (lint, test:umd, test:framework) to not run on travis anymore. This change fixes that by adding a variation to the matrix that runs outside the docker, and runs all the tests it used to run previously.~

- [x] merge after https://github.com/twilio/twilio-video.js/pull/721

**Contributing to Twilio**


.eslintrc.json
```diff
@@ -0,0 +1,341 @@
+{
```
manjeshbhargav: @makarandp0 Can we indent with 2 spaces? Also, can you make sure that we retain only those rules present in the `.eslintrc` file? That way we don't have to make any (or very little) modifications in the source code.

test/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
```
manjeshbhargav: @makarandp0 Indent with 2 spaces. Also, extend the the root `.eslintrc.json` here.

test/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
...
+    "env": {
+        "browser": true,
+        "es6": true,
+        "node": true,
+        "mocha": true
+    },
+    "rules": {
+        "no-undefined": "off"
+    }
+}
```
manjeshbhargav: @makarandp0 Add newline at EOF.

.eslintrc.json
makarandp0: @manjeshbhargav - I will update with 2 spaces indentation for this json file. 


About retaining previous rule - I have ensured that all previous rules are carried over. 
Upgrading to newer eslint gives us an opportunity to catch more issues that were not in older version. Looks like our code is mostly clean, and it required only few exceptions that too mostly in test code.  I think it would be good to keep new checks by default, and remove new rules on case-by-case basis as we find them bothersome. Right now I have not noticed any that deserves blocking.  

Yes down side is some more files are modified in this PR, but right now they are all done, and it's more work to remove those changes.

makarandp0:
Hey @manjeshbhargav - I have addressed your feedback, Can you take another look. 

I also noticed a regression caused by my earlier change (https://github.com/twilio/twilio-video.js/pull/712/files#diff-354f30a63fb0907d4ad57269548329e3R86) which caused travis to skip some of the steps it used to run previously (`lint test:umd test:framework`)
To fix this, I have added a variation to our test matrix that will run outside docker - and run all these steps.

test/.eslintrc.json
manjeshbhargav: @makarandp0 Can you make sure you extend the root `.eslintrc.json` here?

makarandp0: It's not needed, because file hierarchy automatically enforces this. (https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy) I have verified that rules from root are applied in this folder.

PR #724, opened 2019-08-23
JSDK-2487 Removing "dominantSpeaker" from VideoRenderDimensions.

manjeshbhargav:
@makarandp0 

This PR removes the "dominantSpeaker" VideoRenderDimensions because VMS will now use
the RenderDimensions corresponding to the Dominant Speaker's priority instead of a dedicated RenderDimensions.

**Contributing to Twilio**


test/unit/spec/connect.js
```diff
@@ -86,9 +86,6 @@ describe('connect', () => {
       [{ video: { renderDimensions: null } }, 'whose .video.renderDimensions is null', TypeError, 'object'],
       [{ video: { renderDimensions: true } }, 'whose .video.renderDimensions is not an object', TypeError, 'object'],
       [{ video: { renderDimensions: ['foo'] } }, 'whose .video.renderDimensions is an Array', TypeError, 'object'],
-      [{ video: { renderDimensions: { dominantSpeaker: null } } }, 'whose .video.renderDimensions.dominantSpeaker is null', TypeError, 'object'],
```
makarandp0: maybe it would be good to keep one test where renderdimentions does have `dominantSpeaker` and verify that it gets ignored.

manjeshbhargav: @makarandp0 I don't think that is necessary. I already verified that the previous unit tests with "dominantSpeaker" failed.

PR #716, opened 2019-08-18
JSDK-2464 Update BandwidthProfileOptions after joining a Group Room.

manjeshbhargav:
_WIP_

## Changes

- [x] Implementing `LocalParticipant.setBandwidthProfile()`
- [ ] Unit tests
- [ ] Integration tests
- [ ] Update CHANGELOG.md

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #731, opened 2019-08-29
JSDK-2464 Implementing LocalParticipant.setBandwidthProfile().

manjeshbhargav:
@makarandp0 

Re-opening #716 here after re-base of `master` onto `feature/bandwidth-profile-api`.

This PR implements `LocalParticipant.setBandwidthProfile()`, which internally sends an RSP update message for changes to the BandwidthProfileOptions.

## Changes

- [x] Implementing `LocalParticipant.setBandwidthProfile()`
- [x] Unit tests
- [x] Integration tests
- [x] Update CHANGELOG.md

**Contributing to Twilio**


lib/signaling/v2/localparticipant.js
```diff
@@ -53,6 +75,21 @@ class LocalParticipantV2 extends LocalParticipantSignaling {
...
+  /**
+   * Update the {@link BandwidthProfileOptions}.
+   * @param {BandwidthProfileOptions} bandwidthProfile
+   */
+  setBandwidthProfile(bandwidthProfile) {
+    if (!isDeepEqual(this._bandwidthProfile, bandwidthProfile)) {
+      // NOTE(mmalavalli): Object.assign() copies the values of only
+      // the top level properties. In order to deep copy the object, we
+      // stringify and parse the object.
+      this._bandwidthProfile = JSON.parse(JSON.stringify(bandwidthProfile));
```
manjeshbhargav: @makarandp0 I want your opinion on the way I perform a deep copy of the object here. I'm not to sure about it, so any thoughts you may have will be helpful here.

makarandp0: this looks fine, as it will get rid of any non-value types from the bandwidthProfile. 
This code won't work if  `bandwidthProfile = undefined`. But I guess that won't happen because validateBandwidthProfile would reject it. I suggest add a comment saying why  `bandwidthProfile = undefined` is not possible here.  Or even better add an assert 
```
assert(bandwidthProfile !== undefined, 'something went wrong undefined should not be here`);
```

test/unit/spec/localparticipant.js
```diff
@@ -151,6 +155,123 @@ describe('LocalParticipant', () => {
...
 
+  describe('#setBandwidthProfile', () => {
+    const subscriptionModes = Object.values(subscriptionMode);
+    const trackPriorities = [PRIORITY_HIGH, PRIORITY_LOW, PRIORITY_STANDARD];
+
+    [
+      [null, 'that is null', TypeError, 'object'],
+      ['foo', 'that is not an object', TypeError, 'object'],
+      [['bar'], 'that is an Array', TypeError, 'object'],
+      [{ video: null }, 'whose .video is null', TypeError, 'object'],
```
makarandp0: please add `{ video: undefined }` as a test case to make sure that gets handled as expected.

test/unit/spec/localparticipant.js
```diff
@@ -151,6 +155,123 @@ describe('LocalParticipant', () => {
...
+      [{ video: 'baz' }, 'whose .video is not an object', TypeError, 'object'],
+      [{ video: ['zee'] }, 'whose .video is an Array', TypeError, 'object'],
+      [{ video: { dominantSpeakerPriority: 2 } }, `whose .video.dominantSpeakerPriority is not one of ${trackPriorities.join(', ')}`, RangeError, trackPriorities],
+      [{ video: { maxSubscriptionBitrate: false } }, 'whose .video.maxSubscriptionBitrate is not a number', TypeError, 'number'],
+      [{ video: { maxTracks: {} } }, 'whose .video.maxTracks is not a number', TypeError, 'number'],
+      [{ video: { mode: 'foo' } }, `whose .video.mode is not one of ${subscriptionModes.join(', ')}`, RangeError, subscriptionModes],
+      [{ video: { renderDimensions: null } }, 'whose .video.renderDimensions is null', TypeError, 'object'],
+      [{ video: { renderDimensions: true } }, 'whose .video.renderDimensions is not an object', TypeError, 'object'],
```
makarandp0: please add `{ video: { renderDimentions: undefined } }` as a test case

lib/signaling/v2/localparticipant.js
```diff
@@ -53,6 +75,21 @@ class LocalParticipantV2 extends LocalParticipantSignaling {
     });
   }
 
+  /**
+   * Update the {@link BandwidthProfileOptions}.
+   * @param {BandwidthProfileOptions} bandwidthProfile
+   */
+  setBandwidthProfile(bandwidthProfile) {
+    if (!isDeepEqual(this._bandwidthProfile, bandwidthProfile)) {
+      // NOTE(mmalavalli): Object.assign() copies the values of only
```
makarandp0: no need to mention `Object.assign` as you are not using it here :)

lib/signaling/v2/room.js
```diff
@@ -225,11 +239,24 @@ class RoomV2 extends RoomSignaling {
     };
   }
 
+  /**
+   * @private
+   */
+  _maybeAddBandwidthProfile(update) {
+    const { bandwidthProfile, bandwidthProfileRevision } = this.localParticipant;
+    if (bandwidthProfile && this._lastBandwidthProfileRevision < bandwidthProfileRevision) {
```
makarandp0: curious - why do we need to maintain the `_lastBandwidthProfileRevision` - What if we published the bandwithProfile with every update ? It would increase a payload a bit, but would make our code simpler.

lib/localparticipant.js
```diff
@@ -406,6 +407,43 @@ class LocalParticipant extends Participant {
...
+   *     video: {
+   *       dominantSpeakerPriority: 'high',
+   *       maxTracks: 1
+   *     }
+   *   }
+   * }).then(function(room) {
+   *   // Update the value for "maxTracks". The value for "dominantSpeakerPriority"
+   *   // will remain unchanged.
+   *   room.localParticipant.setBandwidthProfile({
+   *     video: {
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Right now, there isn't. The developer would have to explicitly set it back to the default value specified in the docs.

lib/signaling/v2/room.js
manjeshbhargav: @makarandp0 That was how I had implemented it at first. But it doesn't make sense to keep sending the payload even if it hasn't changed.

lib/signaling/v2/localparticipant.js
manjeshbhargav: @makarandp0 I think it's good to specify why we can't use `Object.assign()` here.

test/unit/spec/localparticipant.js
manjeshbhargav: @makarandp0 `JSON.stringify()` will remove properties explicitly set to `undefined`. So I think we are good here.

makarandp0: @manjeshbhargav, I am sure it works now, but adding a test would ensure that if we manage to break the behavior it will continues to work.

manjeshbhargav: @makarandp0 I've already added unit tests for the cases you mentioned in the last commit.

test/unit/spec/localparticipant.js
manjeshbhargav: @makarandp0 I've already added unit tests for the cases you mentioned in the last commit.

PR #733, opened 2019-09-03
JSDK-2442 Update LocalTrack's publish priority

manjeshbhargav:
@makarandp0 

Re-opening #717 here after rebasing `feature/bandwidth-profile-api`.

This PR contains the changes specified below. Once the `track_priority` MSP channel is available on VMS, I will add integration tests.

## Changes

- [x] Implementing `LocalTrackPublication.setPriority()`
- [x] Implementing `RemoteTrackPublication#publishPriorityChanged`
- [x] Unit tests
- [ ] Integration tests
- [x] Update CHANGELOG.md

**Contributing to Twilio**


CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
...
+
+  // After a while, change the priority to "low".
+  localTrackPublication.setPriority(low);
+  ```
+
+  This will update `publishPriority` on all corresponding RemoteTrackPublications and
+  emit a new event "publishPriorityChanged" to notify the user:
+
+  ```js
+  remoteTrackPublication.on('publishPriorityChanged', priority => {
```
makarandp0: This need not be fixed in this change but it looks like the `priority` parameter here is redundant (since its same as  `remoteTrackPublication.publishPriority`). Caller would find it more useful if we passed oldPriority instead of newPriority as a parameter. That way this event would become more easily action-able for the consumer.

lib/media/track/remotetrackpublication.js
```diff
@@ -89,6 +95,10 @@ class RemoteTrackPublication extends TrackPublication {
           this.track._setSwitchedOff(signaling.isSwitchedOff);
         }
       }
+      if (priority !== signaling.priority) {
+        priority = signaling.priority;
+        this.emit('publishPriorityChanged', priority);
```
makarandp0: I suggest we should pass oldPriority to the even (instead of / along with) new priority - But this need not hold of this PR

lib/signaling/v2/room.js
```diff
@@ -346,11 +379,41 @@ class RoomV2 extends RoomSignaling {
...
+    const trackPriorityPromise = this._getTrackReceiver(id).then(receiver => {
+      if (receiver.kind !== 'data') {
+        throw new Error('Expected a DataTrackReceiver');
+      } if (this._trackPriorityPromise !== trackPriorityPromise) {
+        return;
+      }
+
+      // NOTE(mmalavalli): The underlying RTCDataChannel is closed whenever
+      // the VMS instance fails over, and a new RTCDataChannel is created in order
+      // to resume sending Dominant Speaker updates.
```
makarandp0: Dominant Speaker => track Priority

CHANGELOG.md
manjeshbhargav: @makarandp0 Typically, any event that is fired to notify a change in value should have the new value as the parameter in the event handler. The developer usually stores the old value in a separate variable in his/her application logic.

lib/media/track/remotetrackpublication.js
manjeshbhargav: @makarandp0 See my previous comment.

CHANGELOG.md
makarandp0: maye we should pass both the values then - That could help developers avoid maintaining the old value themselves.

makarandp0:
Reopened as https://github.com/twilio/twilio-video.js/pull/774

PR #735, opened 2019-09-04
Merge feature/bandwidth-profile-api to master.

manjeshbhargav:
@makarandp0 

We are code-complete for phase 1. In order to cut the final RC before releasing 2.0.0-beta14, we need to merge the feature branch to master.

**Contributing to Twilio**


.eslintrc.json
```diff
@@ -133,7 +133,7 @@
     "no-bitwise": "off",
     "no-buffer-constructor": "error",
     "no-caller": "error",
-    "no-catch-shadow": "error",
```
makarandp0: why was this necessary ? It's okay to disable it for now, but we should file a task to re-enable it by turning disabling this for offending existing line

manjeshbhargav: @makarandp0 This is an error regarding Internet Explorer 8 and earlier:
```
error  Value of 'error' may be overwritten in IE 8 and earlier
```
So I don't think we need this check.

PR #737, opened 2019-09-05
JSDK-2500 breakup connect tests

makarandp0:
Firefox (especially when running inside docker) disconnects if too many tests are run in the same browser instance. It's mostly because of leaks in firefox. For a workaround we break up our tests into multiple file with each file is executed in new instance of browser. This changes breaks one of our larger test file `connect.js` into two. 

**Contributing to Twilio**


makarandp0:
closing the PR. Will take this up after BW Profile tests are merged in.

PR #739, opened 2019-09-10
JSDK-2347 subscriber side trackpriority

makarandp0:
@manjeshbhargav 

This change implements `RemoteTrack.setPriority` thus letting subscribers set priority for the tracks. 

For this picked up changes from #733 needed for trackPrioritySignaling MSP object.

TODO:
- [x] unit tests
- [x] integration tests
- [x] changelog update
- [ ] enable integration test for prod when VMS changes go in prod.
- [x] ~update base once https://github.com/twilio/twilio-video.js/pull/733 merges~ 
- [x] update VBT/Simpler-signaling to consume this API (https://code.hq.twilio.com/client/video.bytwilio.com/pull/96)

**Contributing to Twilio**


lib/signaling/participant.js
```diff
@@ -152,6 +156,31 @@ class ParticipantSignaling extends StateMachine {
...
   }
 
+
+  /**
+   * @param {TrackSid} trackSid
+   * @param {Priority?} subscriberPriority
+   * @returns {void}
+   */
+  updateTrackSubscriberPriority(trackSid, subscriberPriority) {
+    if (this._trackPrioritySignaling) {
```
makarandp0: I am also bit concerned about a possible race condition - if user happen to set track priority while MSP object was null, because it was not yet negotiated or because VMS was transitioning from failover. One solution is to always queue the update priority request behind `trackPriorityPromise`. But that adds bit of complications. BTW I think  same potential problem exists in https://github.com/twilio/twilio-video.js/pull/733 as well.

lib/signaling/v2/room.js
```diff
@@ -434,6 +435,9 @@ class RoomV2 extends RoomSignaling {
 
       const trackPrioritySignaling = new this._TrackPrioritySignaling(receiver.toDataTransport());
       this.localParticipant.setTrackPrioritySignaling(trackPrioritySignaling);
+      this.participants.forEach(participant => {
```
makarandp0: is there better way than having up update all participant signaling ?

lib/participant.js
```diff
@@ -267,7 +267,7 @@ class Participant extends EventEmitter {
     const RemoteVideoTrackPublication = this._RemoteVideoTrackPublication;
     const RemoteDataTrack = this._RemoteDataTrack;
     const RemoteDataTrackPublication = this._RemoteDataTrackPublication;
-    const signaling = this._signaling;
+    const participantSignaling = this._signaling;
```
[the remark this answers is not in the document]
makarandp0: renaming was necessary here, because of we were shadowing this variable at local scope, and I needed access to shadowed variable.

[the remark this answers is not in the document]
makarandp0: the variable was declared as `signaling` at line 270. But then there was local scoped variable  @ line 315 was shadowing it.  In order to access variable on line 335, I had to rename it. Actually there is a lint rule 'eslint-no-shadow' that would prevent such shadowing - but its turned off for our repo,  And I think we should enable it.

test/integration/spec/bandwidthprofile.js
```diff
@@ -0,0 +1,128 @@
...
+  trackSwitchedOff,
+  trackSwitchedOn
+} = require('../../lib/util');
+
+
+describe('bandwidth profile', function() {
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(60000);
+
+  describe('RemoteTrack.setPriority', () => {
```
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav - we do not have a file for RemoteTrack  integration tests specifically. Are you suggesting that we rename this to 'test/integration/spec/remoteTrack.js`?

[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Yes, it should be `test/integration/spec/remotetracks.js`. We should write two tests:
1. Test for making sure that RemoteTrack.priority is set when we call `setPriority`. This should be for audio, video and data tracks.
2. Test for making sure that RemoteVideoTrack.priority has the desired effect with VMS. (video only).

lib/participant.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I think we can avoid this by defining the `setPriority` callback outside the `trackSignalingSubscribed` method. That we we don't need the renaming:
```js
const setPriority = priority => signaling.updateTrackPriority(priority);

function trackSignalingSubscribed(signaling) {
  ...
  const options = { log, name, setPriority };
  const track = kind === 'data'
        : new RemoteTrack(sid, trackTransceiver, isEnabled, options);
  ...
}
```
What do you think?

makarandp0:
@manjeshbhargav I have incorporated your feedback. Can you take another look.

lib/media/track/remotemediatrack.js
```diff
@@ -42,13 +52,41 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
         sid: {
           enumerable: true,
           value: sid
         }
       });
     }
 
+    /**
+     * Update the subscribe {@link Track.Priority} of the {@link RemoteMediaTrack}.
+     * @param {?Track.Priority} priority - the new subscribe {@link Track.Priority}
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Can you incorporate this change?

lib/media/track/remotemediatrack.js
```diff
@@ -42,13 +52,41 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+     * @returns {this}
+     * @throws {RangeError}
+     */
+    setPriority(priority) {
+      const priorityValues = [null, ...Object.values(trackPriority)];
+      if (!priorityValues.includes(priority)) {
+        // eslint-disable-next-line new-cap
+        throw E.INVALID_VALUE('priority', priorityValues);
+      }
+
```
manjeshbhargav:
```suggestion
```

lib/media/track/remotevideotrack.js
```diff
@@ -22,11 +23,12 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
    * Construct a {@link RemoteVideoTrack}.
    * @param {Track.SID} sid - The {@link RemoteVideoTrack}'s SID
    * @param {MediaTrackReceiver} mediaTrackReceiver - A video MediaStreamTrack container
-   * @param {boolean} isEnabled - whether the {@link RemoteAudioTrack} is enabled
+   * @param {boolean} isEnabled - whether the {@link RemoteVideoTrack} is enabled
+   * @param {function(?Track.Priority): void} setPriority - callback to set subscriber priority for the track
```
manjeshbhargav:
```suggestion
   * @param {function(?Track.Priority): void} setPriority - Set or clear the subscribe {@link Track.Priority} of the {@link RemoteVideoTrack}

```

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+
+function getTracksOfKind(participant, kind) {
+  return [...participant.tracks.values()].filter(remoteTrack => remoteTrack.kind !== kind).map(({ track }) => track);
+}
+
+describe('RemoteParticipant', function() {
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(60000);
+
+  describe('.setPriority', () => {
```
manjeshbhargav:
```suggestion
  describe('#setPriority', () => {
```

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    afterEach(async () => {
+      [thisRoom, ...thoseRooms].forEach(room => room && room.disconnect());
+      if (thisRoom) {
+        await completeRoom(thisRoom.sid);
+      }
+    });
+
+    ['audio', 'video', 'data'].forEach(kind => {
```
manjeshbhargav: @makarandp0 
```js
combinationContext([
  [
    ['audio', 'data', 'video'],
    x => `when called on a Remote${capitalize(x)}Track`
  ],
  [
    [null, ...Object.values(trackPriority), undefined, 'foo', 42, true, {}, []]
    x => `when called with priority ${x}`
  ]
], ([kind, priority]) => {
  it(`should return the Remote${capitalize(kind)}Track`, () => {
    ...
  });
  
  if (![null, ...Object.values(trackPriority)].includes(priority) {
    it(`should set .priority`, () => {
      ..
    });
  } else {
    it(`should not set .priority`, () => {
      ..
    });
  }
});
```

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+        tracksSubscribed(aliceRemote, 3),
+        tracksSubscribed(bobRemote, 3)
+      ]);
+
+      [aliceRemoteVideoTrack, bobRemoteVideoTrack] = [aliceRemote, bobRemote].map(({ videoTracks }) => {
+        return [...videoTracks.values()][0].track;
+      });
+    });
+
+    afterEach(async () => {
```
manjeshbhargav: @makarandp0 Stop Alice's and Bob's LocalTracks.

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+function getTracksOfKind(participant, kind) {
+  return [...participant.tracks.values()].filter(remoteTrack => remoteTrack.kind !== kind).map(({ track }) => track);
+}
+
+describe('RemoteParticipant', function() {
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(60000);
+
+  describe('.setPriority', () => {
+    let thisRoom; let thoseRooms;
```
manjeshbhargav: @makarandp0 Nitpick: Please declare only one variable per line.

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+      [aliceTracks, bobTracks] = await Promise.all([1, 2, 3].map(async () => [
+        createSyntheticAudioStreamTrack() || await createLocalAudioTrack({ fake: true }),
+        await createLocalVideoTrack(smallVideoConstraints),
+      ]));
+
+      [aliceLocal, bobLocal] = thoseRooms.map(room => room.localParticipant);
+      [aliceRemote, bobRemote] = [thisRoom.participants.get(aliceLocal.sid), thisRoom.participants.get(bobLocal.sid)];
+
+      // Alice publishes her tracks at low priority
+      // Bob publishes his tracks at medium priority
```
manjeshbhargav:
```suggestion
      // Bob publishes his tracks at standard priority
```

test/integration/spec/remotetracks.js
```diff
@@ -0,0 +1,136 @@
...
+          assert(error instanceof RangeError);
+          assert(/priority must be one of/.test(error.message));
+          errorWasThrown = true;
+        }
+        assert.equal(errorWasThrown, true, 'was expectiing an error to be thrown, but it was not');
+      });
+    });
+
+    if (defaults.topology !== 'peer-to-peer') {
+      it('subscriber can update track\'s effective priority', async () => {
```
manjeshbhargav: @makarandp0 Can you make sure you have tests for 3 scenarios:
1. Upgrading the effective RemoteVideoTrack priority.
2. Downgrading the effective RemoteVideoTrack priority.
3. Reverting to the publisher priority by passing `null`. (You can disable this test with a NOTE pointing to the VMS bug).

test/unit/spec/media/track/remotedatatrack.js
```diff
@@ -113,6 +114,29 @@ describe('RemoteDataTrack', () => {
     });
   });
 
+  describe('.setPriority', () => {
+    [null, ...Object.values(trackPriority)].forEach((priorityValue) => {
```
manjeshbhargav: @makarandp0 Test for verifying return value?

test/unit/spec/media/track/remotedatatrack.js
```diff
@@ -113,6 +114,29 @@ describe('RemoteDataTrack', () => {
     });
   });
 
+  describe('.setPriority', () => {
```
manjeshbhargav:
```suggestion
  describe('#setPriority', () => {
```

test/unit/spec/media/track/remotemediatrack.js
```diff
@@ -59,6 +60,42 @@ const { FakeMediaStreamTrack } = require('../../../../lib/fakemediastream');
       });
     });
 
+    const { trackPriority } = require('../../../../../lib/util/constants');
+
+    describe('.setPriority', () => {
```
manjeshbhargav:
```suggestion
    describe('#setPriority', () => {
```

test/unit/spec/media/track/remotemediatrack.js
```diff
@@ -59,6 +60,42 @@ const { FakeMediaStreamTrack } = require('../../../../lib/fakemediastream');
...
+      let options;
+      let track;
+      let onPriorityChange;
+      beforeEach(() => {
+        onPriorityChange = sinon.spy();
+        track = makeTrack('foo', 'bar', kind, true, options, RemoteTrack, onPriorityChange);
+      });
+
+      [null, ...Object.values(trackPriority)].forEach((priorityValue) => {
```
manjeshbhargav: @makarandp0 Test for verifying return value?

test/integration/spec/remotetracks.js
makarandp0: @manjeshbhargav - I think separating the valid and invalid test cases makes it much more readable. There isn't much duplication of code to warrant consolidating them.

test/integration/spec/remotetracks.js
makarandp0: okay, I will update this, but to catch such things as errors in future how about enabling this rule: https://eslint.org/docs/2.13.1/rules/max-statements-per-line ?

test/integration/spec/remotetracks.js
makarandp0: I have a test for #1 - but #2 is no-op. As I understand subscriber can not lower effctive  priority of the track - They can only increase the effective priority.

manjeshbhargav: @makarandp0 According to David (who implemented subscriber priority), what you said only applies to Dominant Speaker tracks. Downgrading should work with `RemoteTrack.setPriority`. You can confirm with him if you want.

test/integration/spec/remotetracks.js
makarandp0: looks like we have 152 violations of this rule. I will get to it later.

test/integration/spec/remotetracks.js
makarandp0: okay, cool, I will add a test to downgrade priority and see it passes

test/unit/spec/media/track/remotemediatrack.js
makarandp0: this test does assert at the end that value equals.

PR #749, opened 2019-09-18
JSDK-2490 enable/disable MediaStream track

makarandp0:
This change  workarounds an [issue in chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=749928) where it continues to play audio even when audio element is not part of the DOM.

The matters were more complicated because we always create a dummy media element (audio/video)  at the start of media track creation (https://github.com/twilio/twilio-video.js/blob/JSDK_2490_audio_detached/lib/media/track/mediatrack.js#L102)

With this fix we will use `enabled` [property of MediaStreamTrack](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled) 

We will mark the `track.enabled = false` when its not attached to any element - This prevents it from rendering into the MediaStream.

To see this fix in action: checkout https://github.com/twilio/video-quickstart-js/tree/JSDK-2490-detach-tracks,  which lets you attach/detach tracks individually. Notice that before the fix, detached audio track continue to play.

TODO
- [x] update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.0.0-beta15 (In Progress)
+2.0.0-beta15 (In progress)
 ==========================
 
 Bug Fixes
 ---------
 
-- Worked around an [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=749928)
-  in Chrome and Safari where browser continued to play WebRTC-based MediaStreamTrack even
-  after corresponding `audio` element was removed from the DOM. With this fix twilio-video.js
+- Worked around a bug in [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=749928)
+  and Safari where browser continued to play WebRTC-based MediaStreamTrack even after
+  corresponding `audio` element was removed from the DOM. With this fix twilio-video.js
   now disables any RemoteMediaTrack when it's not attached to any media elements. (JSDK-2490)
 
 2.0.0-beta14 (September 17, 2019)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta15 (In Progress)
+==========================
+
+Bug Fixes
+---------
+
+- Worked around an [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=749928)
+  in Chrome and Safari where browser continued to play WebRTC-based MediaStreamTrack even
+  after corresponding `audio` element was removed from the DOM. With this fix twilio-video.js
+  now disables any RemoteMediaTrack when it's not attached to any media elements. (JSDK-2490)
+
 2.0.0-beta14 (September 17, 2019)
 =================================
 
@@ -46,7 +57,7 @@ New Features
         }
       }
     }
-  });  
+  });
   ```
 
   ### Track Priority (private beta)
@@ -68,7 +79,7 @@ New Features
   to other Tracks that may be published to the Room. The media server takes this into
   account while allocating a subscribing RemoteParticipant's bandwidth to the corresponding
   RemoteTrack. If you do not specify a priority, then it defaults to `standard`.
-  
+
   You can also find out about the priorities of RemoteTracks published by other
   RemoteParticipants by accessing a new property `publishPriority` on the corresponding
   RemoteTrackPublications:
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -221,7 +221,6 @@ class MediaTrack extends Track {
     }
 
     this._attachments.delete(el);
-
     return el;
   }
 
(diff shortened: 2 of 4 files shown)


lib/media/track/mediatrack.js
```diff
@@ -159,6 +159,9 @@ class MediaTrack extends Track {
       this._attachments.add(el);
     }
 
+    // enable the track;
+    this.mediaStreamTrack.enabled = true;
```
manjeshbhargav: @makarandp0 What if the publisher has disabled the Track? Can we make sure that we take that into account?

lib/media/track/mediatrack.js
```diff
@@ -218,17 +221,20 @@ class MediaTrack extends Track {
...
     this._attachments.delete(el);
+
+    if (this._attachments.size === 0) {
+      // NOTE(mpatwardhan): chrome continues playing audio track
+      // even after audio element is removed from the DOM.
+      // Removing track from the stream does not work either.
+      // https://bugs.chromium.org/p/chromium/issues/detail?id=749928
+      // So here we mark the track.enabled = false when there are
+      // no elements attached to it.
+      this.mediaStreamTrack.enabled = false;
```
manjeshbhargav: @makarandp0 What if we just set `el.srcObject = null` here instead, and set it back to a MediaStream when `attach()` is called? Wouldn't that work as well?

makarandp0: @manjeshbhargav, I experimented with this. It would typically work, but not in our case because we have  '_dummyEl'  which also would need to be fixed to point its srcObject to `null`

It seems like  `enabled` property fits our purpose really well here.

manjeshbhargav: @makarandp0 I'm worried about disabling the MediaStreamTrack because it might affect the emission of the "started" event on the RemoteTrack. Also, this does not cover the case when the RemoteTrack is created and no attachments are made yet. Plus this code is shared with LocalTrack as well.

lib/media/track/mediatrack.js
makarandp0: not sure what you mean - Shouldn't this be independent of what publisher does with the track.  This property is only controlling if the track is allowed to render or not. 

```
The enabled property on the MediaStreamTrack interface is a Boolean value which is true if the track is allowed to render the source stream or false if it is not. This can be used to intentionally mute a track. When enabled, a track's data is output from the source to the destination; otherwise, empty frames are output.
```

lib/media/track/mediatrack.js
makarandp0: Thanks, very good points - I will check about the `started` property to make sure its not affected by setting 'enable' to false. Also I will look into moving this for remote track only.

makarandp0: Thanks @manjeshbhargav for raising these concerns - I have now verified that 

1) mediatrack do get started notification - I forced the track `enabled=false`, before receiving the 'started' event, and verified that we still get `started` event.  I think that's because when marked as disabled, mediaStreamTrack still pumps frames into mediaStream but just empty ones. 

2)  I also verified that the track does not play when there are no attachments are made. I guess the problem is mainly about chrome not handling dom removals of audio elements, and maintaining some kind of global state for webrtc tracks. 

3) Third part - about this code handling LocalTrack - thats good point but it's  matter of refactoring the code. If we agree to this solution then I can go tackle 3. 

I agree that although this solution is simple and elegant, it may have side effects, but this property is really promising and documentation does indicate this to be correct usage of the property - I am hoping that we can discover any issues in our testing.

makarandp0:
@manjeshbhargav, I have updated this to limit changes only for Remote Tracks as we discussed. Its ready for review now.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.0.0-beta15 (In Progress)
```
manjeshbhargav:
```suggestion
2.0.0-beta15 (in progress)
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+==========================
+
+Bug Fixes
+---------
+
+- Worked around an [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=749928)
```
manjeshbhargav:
```suggestion
- Worked around a bug in [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=749928) and Safari
```

lib/media/track/remotemediatrack.js
```diff
@@ -72,6 +72,34 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
     }
+
+    attach(el) {
+      const result = super.attach(el);
+      if (this.mediaStreamTrack.enabled !== true) {
+        // NOTE(mpatwardhan): we disable mediaStreamTrack when there
+        // are no attchments to it (see notes below). Now that there are attachments.
+        // reenable the track.
+        // eslint-disable-next-line no-console
+        console.log('enabling the track: ', this.sid);
```
manjeshbhargav:
```suggestion
```

lib/media/track/remotemediatrack.js
```diff
@@ -72,6 +72,34 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+        this.mediaStreamTrack.enabled = true;
+      }
+      return result;
+    }
+
+    detach(el) {
+      const result = super.detach(el);
+      if (this._attachments.size === 0) {
+        // eslint-disable-next-line no-console
+        console.log('disabling the track: ', this.sid);
```
manjeshbhargav:
```suggestion
```

Issue #748, opened 2019-09-18
Twilio.Video.isSupported flag returns true on Edge, but Edge is not supported


dambusm:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

```js
console.log(Video.isSupported)
```

**Expected behavior:**

When running on Microsoft Edge on Windows 10 (EdgeHTML), should return false as Edge is not supported (https://github.com/twilio/twilio-video.js#browser-support)

**Actual behavior:**

Returns true

**Software versions:**

 - [x] Browser(s): Microsoft Edge 44.18362.329.0
 - [x] Operating System: Windows 10 Pro 10.0.18362
 - [x] twilio-video.js: 2.0.0-beta14
 - [x] Third-party libraries (e.g., Angular, React, etc.): React 16.8.6

makarandp0:
Hello @dambusm,  Thank you for writing in regarding this issue. Yes this flag checks for functionality provided by the browser, than actual checks against browser version.

https://github.com/twilio/twilio-video.js/blob/master/lib/util/support.js#L31

Can you tell us how are you using or planning to use this flag in your application, and maybe we can suggest some workaround, or see if we can update this flag to report false for anything bug supported browsers.

dambusm:
Hi @makarandp0, in our application we need to display a notice to the user and prevent them from attempting to join calls if they're running in an unsupported browser. On Edge our Twilio calls all fail when the local participant connects, with this error:
```
code: 53405
description: "Media connection failed"
message: "Media connection failed"
name: "TwilioError"
[...]

PR #752, opened 2019-09-25
JSDK-2515 - Video.isSupported should return false for Edge

makarandp0:
This came up as a github [issue](https://github.com/twilio/twilio-video.js/issues/748). Since we do not officially support Edge, `isSupported` should return false for edge. 

**Contributing to Twilio**

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -25,10 +25,10 @@ function isGetUserMediaSupported() {
 }
 
 /**
- * @returns {boolean} - true if the browser is explicitely unsupported.
+ * @returns {boolean} - true if the browser is explicitly unsupported.
  *  Note: this function returning false does not mean that browser is supported.
  */
-function isExplicitelyUnsupportedBrowser() {
+function isExplicitlyUnsupportedBrowser() {
   if (typeof navigator !== 'undefined' && typeof navigator.userAgent === 'string') {
     return /Edge/.test(navigator.userAgent);
   }
@@ -40,7 +40,7 @@ function isExplicitelyUnsupportedBrowser() {
  * @returns {boolean}
  */
 function isSupported() {
-  return !isExplicitelyUnsupportedBrowser()
+  return !isExplicitlyUnsupportedBrowser()
     && !!guessBrowser()
     && isGetUserMediaSupported()
     && isRTCPeerConnectionSupported();
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -25,11 +25,23 @@ function isGetUserMediaSupported() {
 }
 
 /**
+ * @returns {boolean} - true if the browser is explicitely unsupported.
+ *  Note: this function returning false does not mean that browser is supported.
+ */
+function isExplicitelyUnsupportedBrowser() {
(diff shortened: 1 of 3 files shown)


lib/util/support.js
```diff
@@ -24,12 +24,24 @@ function isGetUserMediaSupported() {
...
+}
+
 /**
  * Check if the current environment is supported by the SDK.
  * @returns {boolean}
  */
 function isSupported() {
-  return !!guessBrowser()
+  return !isExplicitelyUnsupportedBrowser()
+    && !!guessBrowser()
```
manjeshbhargav: @makarandp0 Wouldn't it be simpler to modify `guessBrowser()` to detect Edge? Then the above line can be refactored like so:
```suggestion
    && guessBrowser() && guessBrowser() !== 'edge'
```

test/unit/spec/util/support.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const assert = require('assert');
+const isSupported = require('../../../../lib/util/support');
+
+describe.only('isSupported', () => {
```
manjeshbhargav:
```suggestion
describe('isSupported', () => {
```

lib/util/support.js
makarandp0: @manjeshbhargav  I thought about it initially, but decided against it because 
1) modifying `guessBrowser` requires changes to @twilio/webrtc - and  
2) `guessbrowser` is widely used, I didn't want it to affect any other callers. if it starts returning `edge` where it returned `chrome`, other callers would need to handle this case.

test/unit/spec/util/support.js
makarandp0: thanks for catching this. Maybe I should add an eslint rule that prevents such mistakes.

lib/util/support.js
```diff
@@ -24,12 +24,24 @@ function isGetUserMediaSupported() {
     || !!(navigator.mozGetUserMedia);
 }
 
+/**
+ * @returns {boolean} - true if the browser is explicitely unsupported.
+ *  Note: this function returning false does not mean that browser is supported.
+ */
+function isExplicitelyUnsupportedBrowser() {
```
manjeshbhargav:
```suggestion
function isExplicitlyUnsupportedBrowser() {
```

lib/util/support.js
```diff
@@ -24,12 +24,24 @@ function isGetUserMediaSupported() {
     || !!(navigator.mozGetUserMedia);
 }
 
+/**
+ * @returns {boolean} - true if the browser is explicitely unsupported.
```
manjeshbhargav:
```suggestion
 * @returns {boolean} - true if the browser is explicitly unsupported.
```

makarandp0:
ported as https://github.com/twilio/twilio-video.js/pull/756 for 1.x

makarandp0:
CI build failed for known unstable test:"LocalTrackPublication #unpublish called with an enabled LocalVideoTrack that was published, unpublished, and then published again" merging now.

PR #780, opened 2019-10-18
JSDK-2550 test to verify creation of Room with recording enabled. 

makarandp0:
a small test that creates a room and connects to it to verify `recording.enabled`

**Contributing to Twilio**

### DIFF
--- a/test/integration/spec/rest.js
+++ b/test/integration/spec/rest.js
@@ -36,6 +36,27 @@ describe('', () => {
 (defaults.topology !== 'peer-to-peer' && enableRestApiTests ? describe : describe.skip)('REST APIs', function() {
   // eslint-disable-next-line no-invalid-this
   this.timeout(60000);
+  describe('recording', () => {
+    let sid;
+    let aliceRoom;
+    let bobRoom;
+    before(async () => {
+      sid = await createRoom(randomName(), 'group-small', { RecordParticipantsOnConnect: true });
+      const options = Object.assign({ name: sid, tracks: [] }, defaults);
+      [aliceRoom, bobRoom] = await Promise.all(['alice', 'bob'].map(getToken).map(token => connect(token, options)));
+    });
+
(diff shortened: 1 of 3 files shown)


test/integration/spec/rest.js
```diff
@@ -36,6 +36,18 @@ describe('', () => {
 (defaults.topology !== 'peer-to-peer' && enableRestApiTests ? describe : describe.skip)('REST APIs', function() {
   // eslint-disable-next-line no-invalid-this
   this.timeout(60000);
+  describe('recording', () => {
```
manjeshbhargav: @makarandp0 This should be in `test/integration/spec/room.js`:
```js
describe('Room', () => {
  ...
  (defaults.topology === 'peer-to-peer' ? describe.skip : describe)('.isEnabled', () => {
    [true, false].forEach(enableRecording => {
      context(`when recording is ${enableRecording ? 'enabled' : 'disabled'}`, () => {
        let room;
        let sid;

        before(async () => {
          sid = await createRoom(randomName(), defaults.topology, { RecordParticipantsOnConnect: true });
          const options = Object.assign({ name: sid, tracks: [] }, defaults);
          room = await connect(getToken(randomName()), options);
        });

        it(`should be set to ${enableRecording}`, () => {
          assert.equal(room.isRecording, enableRecording);
        });

        after(() => {
          if(room) {
            room.disconnect();
          }
          return completeRoom(sid);
        });
      });
    });
  });
  ...
});
```

manjeshbhargav: @makarandp0 This file also has empty tests for "recordingStarted" and "recordingStopped" events. Can you work with @innerverse (Seth) in order to see if we can write tests for those events?

makarandp0: @manjeshbhargav - I had a discussion with Seth, and unfortunately there isn't a way to invoke these events right now

PR #782, opened 2019-10-21
JSDK-2501 integration test for datatrack publish failures

makarandp0:
We have been  seeing intermittent failures on firefox when publishing data tracks. This issue was fixed as VMS-2005, but has regressed since then. I found that the failures are timing sensitive, and have something to do with MSP objects. 

These tests help reproduce them more regularly, and also captures room sid on repro. They will prevent such regressions going forward.

Example failure: https://circleci.com/gh/twilio/twilio-video.js/2020#tests/containers/1

**Contributing to Twilio**

TODO: 
- [ ] Merge after [VMS-2202](https://issues.corp.twilio.com/browse/VMS-2202) is fixed

### DIFF
--- a/test/integration/spec/datatrackPublish.js
+++ b/test/integration/spec/datatrackPublish.js
@@ -15,7 +15,7 @@ const {
   waitFor
 } = require('../../lib/util');
 
-describe('JSDK-2477: data track not getting published in firefox', function() {
+describe('JSDK-2501: data track not getting published in firefox', function() {
   // eslint-disable-next-line no-invalid-this
   this.timeout(60000);
   [true, false].forEach((dominantSpeaker) => {
--- a/test/integration/spec/datatrackPublish.js
+++ b/test/integration/spec/datatrackPublish.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const {
+  connect,
+  createLocalAudioTrack,
+  LocalDataTrack,
+} = require('../../../lib');
+
+const defaults = require('../../lib/defaults');
+const getToken = require('../../lib/token');
+
+const {
+  participantsConnected,
+  randomName,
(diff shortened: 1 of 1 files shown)


test/integration/spec/datatrackPublish.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
```
manjeshbhargav: @makarandp0 Isn't there already a test for this:
```
#publishTrack
  called with an enabled
    LocalAudioTrack
      that has never been published
        should raise a "trackPublished" event on the corresponding RemoteParticipant with a RemoteTrackPublication
```
The `defaults` already has `dominantSpeaker: true`.

makarandp0: @manjeshbhargav - This test is specifically for data tracks publishing failures. I have seen it has something to do with MSP objects (hence `[true, false].forEach((dominantSpeaker) `) and also is timing sensitive and thus `[0, 5000].forEach((delay) `

manjeshbhargav: @makarandp0 For the LocalDataTrack case, there is this scenario:
```
#publishTrack
  called with an enabled
    LocalDataTrack
      that has never been published
        should raise a "trackPublished" event on the corresponding RemoteParticipant with a RemoteTrackPublication
```
Also, I'm guessing that the `0` delay scenario is the problematic one? In that case, the above scenario should cover it no?

makarandp0: @manjeshbhargav - I saw that adding a  delay of 5000 ms between Alice and bob connecting reproduces issue more frequently. I am hoping that we would know more once somebody investigates VMS-2202  and readjust the tests accordingly.

Issue #736, opened 2019-09-04
Huge delay between trackSubscribed and trackStarted events.


sandeepnandula:
Hi, @twilio_team our clients are facing a screen share delay issue on group calls.

When a person does a screen share in group call on the receivers end it's taking a while to see that screen track.

Recently it occurred more than`5mins` delay in a group room having only two people, I’ve attached the room details & logs below for reference. Please let us know if you need any further information regarding this.

Twilio_sdk_version: 2.0.0-beta11

RoomSID: RM15f80605578b419ab91605cf2284d96a

No of participant : 2
log: [twilio_logs](https://connect-feedback-logs.storage.googleapis.com/generallog-f74cfc1f-6fad-4081-81f2-d79a8dff9faa.log)

**Note: We saw there is a huge delayed is occurring frequently between the trackSubscribed & trackStarted twilio_events for the tracks.**

makarandp0:
Thank you for writing about the issue @sandeepnandula, I do see the delayed start event from the logs you provided. I will look further into this and get back to you.

makarandp0:
@sandeepnandula, we are still investigating this. I found that some logs are missing for this room from our servers. Have you by any chance specified `insights: false` as one of the  [connectOption](https://media.twiliocdn.com/sdk/js/video/releases/2.0.0-beta13/docs/global.html#ConnectOptions) to `Video.connect` ? 

Thanks,
Makarand

sandeepnandula:
Hi @makarandp0 thank you for the quick response. Yeah, we made `insights: false` in the Twilio connection options.

makarandp0:
Thanks again for reporting this issue. We have identified the root cause, and filed a ticket to get it fixed. (VMS-2096). Will update when it gets resolved.

sandeepnandula:
Thank you @makarandp0 for the update 👍

sandeepnandula:
Hi @makarandp0, any updates regarding this bug?

makarandp0:
Hello @sandeepnandula, we are actively working on this - I will update once we have fix ready.

sandeepnandula:
Thank you for the update @makarandp0 👍

kamesh-a:
Hi @makarandp0 

We are getting more feedbacks related delayed starting of screen shares, even in our group call many screen shares didn't start immediately, we had to stop and start it again. Can we please take a look into this issue.

[...]

PR #790, opened 2019-10-30
JSDK-2557 Chrome screen share does not start

makarandp0:
This works around an issue in chrome where it sometimes do not send any data on screenshare track if there was a maxBitate value on the RTCRtpEncodingParameters for the RTCRtpSender associated with the screen share Track: 

This resulted in some issues like the one reported by this github issue: https://github.com/twilio/twilio-video.js/issues/736
**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -1393,7 +1393,7 @@ function updateEncodingParameters(pcv2) {
     if (maxBitrate === null || maxBitrate === 0) {
       removeMaxBitrate(params);
     } else if (utils.isChromeScreenShareTrack(sender.track)) {
-      // JSDK-2557: sometimes chrome does not send any bytes on screen track if MaxBitRate is set on it via setParameters,
+      // Note(mpatwardhan): Sometimes (JSDK-2557) chrome does not send any bytes on screen track if MaxBitRate is set on it via setParameters,
       // To workaround this issue we will not apply maxBitrate if the track appears to be a screen share track created by chrome
       pcv2._log.warn(`Not setting maxBitrate for ${sender.track.kind} Track ${sender.track.id} because it appears to be screen share track: ${sender.track.label}`);
     } else {
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -29,11 +29,12 @@ const {
   MediaClientRemoteDescFailedError
 } = require('../../util/twilio-video-errors');
 
+const utils = require('../../util');
 const {
   buildLogLevels,
   makeUUID,
   oncePerTick
-} = require('../../util');
+} = utils;
 
 const IceBox = require('./icebox');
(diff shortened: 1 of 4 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -1391,6 +1392,10 @@ function updateEncodingParameters(pcv2) {
 
     if (maxBitrate === null || maxBitrate === 0) {
       removeMaxBitrate(params);
+    } else if (utils.isChromeScreenShareTrack(sender.track)) {
```
syerrapragada: Curious Question: We could continue to use destructure this function from utils at line 37. Just curious to know why this is done differently here?

lib/util/index.js
```diff
@@ -692,6 +694,20 @@ function inRange(num, min, max) {
...
 }
 
+/**
+ * returns true if given MediaStreamTrack is a screen share track
+ * @private
+ * @param {MediaStreamTrack} track
+ * @returns {boolean}
+ */
+function isChromeScreenShareTrack(track) {
+  // Chrome creates screen share tracks with label like: "screen:69734272*"
```
syerrapragada: Nitpick: NOTE(<name>)

lib/signaling/v2/peerconnection.js
makarandp0: @syerrapragada - I chose to not destructure, so that it can be mocked easily in unit tests  `utils.isChromeScreenShareTrack`. With destructuring the function had to be mocked before this file is loaded. w/o it it can be mocked w/o reloading the file.

lib/signaling/v2/peerconnection.js
```diff
@@ -1391,6 +1392,10 @@ function updateEncodingParameters(pcv2) {
...
+      // Note(mpatwardhan): Sometimes (JSDK-2557) chrome does not send any bytes on screen track if MaxBitRate is set on it via setParameters,
```
syerrapragada: Nitpick: Please use `NOTE` all caps to be consistent with other comments.

lib/util/index.js
```diff
@@ -692,6 +694,19 @@ function inRange(num, min, max) {
...
 }
 
+/**
+ * returns true if given MediaStreamTrack is a screen share track
+ * @private
+ * @param {MediaStreamTrack} track
+ * @returns {boolean}
+ */
+function isChromeScreenShareTrack(track) {
+  // Note(mpatwardhan): Chrome creates screen share tracks with label like: "screen:69734272*"
```
syerrapragada: Same here

lib/util/index.js
```diff
@@ -692,6 +694,19 @@ function inRange(num, min, max) {
...
+  // we will check for label that starts with "screen:D" where D being a digit.
+  const isChrome = util.guessBrowser() === 'chrome';
+  return isChrome && track.kind === 'video' && track.label && /^screen:[0-9]+/.test(track.label);
```
syerrapragada: Hacky but I guess that is the only way to do it I guess.

PR #792, opened 2019-10-30
JSDK-2347, JSDK-2442: stale cached publisher priority prevented updates

makarandp0:
This fixes an issue where we weren't updating cached value of `updatedPrioirty` causing priority updates to not have an effect.

This was the root cause for both JSDK-2347 and JSDK-2442

our integration tests did not catch this because we did not have a test that made multiple priority updates on the same track. (1st priority update works fine, but subsequent would not). Added a test to cover this scenario.

Note: there are lots of indentation changes. Pls turn off whitespace to see just relevant change: https://github.com/twilio/twilio-video.js/pull/792/files?utf8=%E2%9C%93&diff=split&w=1

**Contributing to Twilio**

### DIFF
--- a/test/integration/spec/localtrackpublication.js
+++ b/test/integration/spec/localtrackpublication.js
@@ -247,7 +247,7 @@ describe('LocalTrackPublication', function() {
 
   // eslint-disable-next-line no-warning-comments
   // TODO: enable these tests when track_priority MSP is available in prod
-  (defaults.topology === 'peer-to-peer' ? describe.skip : describe.only)('#setPriority', () => {
+  (defaults.topology === 'peer-to-peer' ? describe.skip : describe)('#setPriority', () => {
     describe('three participant tests', () => {
       let thisRoom;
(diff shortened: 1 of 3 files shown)


test/integration/spec/localtrackpublication.js
```diff
@@ -246,160 +247,220 @@ describe('LocalTrackPublication', function() {
...
-    let aliceRemoteVideoTrackPublication;
-    let bobRemoteVideoTrackPublication;
-
-    beforeEach(async () => {
-      const dataTrack = new LocalDataTrack();
-      [, thisRoom, thoseRooms] = await setup({
-        testOptions: {
-          bandwidthProfile: {
-            video: { maxTracks: 1, dominantSpeakerPriority: 'low' }
+  (defaults.topology === 'peer-to-peer' ? describe.skip : describe.only)('#setPriority', () => {
```
syerrapragada: Please change describe.only -> describe

PR #794, opened 2019-11-04
regression test for VMS-2231 standard priority updates missing

makarandp0:
This adds some integration tests that verify that when publisher changes priority from a->b, 
subscriber gets notified of the change. Currently they are failing when b=standard.

TODO:
- [x] remove console.logs
- [x] merge when [VMS-2231](https://issues.corp.twilio.com/browse/VMS-2231) is fixed

**Contributing to Twilio**

### DIFF
--- a/test/integration/spec/localtrackpublication.js
+++ b/test/integration/spec/localtrackpublication.js
@@ -249,7 +249,7 @@ describe('LocalTrackPublication', function() {
   // TODO: enable these tests when track_priority MSP is available in prod
   (defaults.topology === 'peer-to-peer' ? describe.skip : describe)('#setPriority', function() {
     // eslint-disable-next-line no-invalid-this
-    this.retries(2);
+    // this.retries(2);
 
     describe('three participant tests', () => {
       let thisRoom;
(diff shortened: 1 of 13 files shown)


test/integration/spec/localtrackpublication.js
```diff
@@ -407,6 +405,55 @@ describe('LocalTrackPublication', function() {
       });
     });
 
+
+    [PRIORITY_LOW, PRIORITY_STANDARD, PRIORITY_HIGH].forEach(beforePriority => {
+      [PRIORITY_LOW, PRIORITY_STANDARD, PRIORITY_HIGH].forEach(afterPriority => {
+        if (beforePriority !== afterPriority) {
```
syerrapragada: We could remove this check here and have a test for updating the same priority as before. WDYT?

makarandp0: if priority does not change then we won't receive `publishPriorityChanged`

syerrapragada: Yes, we could have that check too in the test?

makarandp0: Thanks for the suggestion - I have updated the test to handle this case.

PR #796, opened 2019-11-05
JSDK-2561" wait for initial DLTS handshake before handling glare

makarandp0:
We did not handle glare when `this._descriptionRevision === 1` - This ensured that DTLS role was negotiated. But that works only for 1st peer connection, when a new peer connection is created in response to VMS failover - the `descriptionRevision` would not be 1. To ensure that every new peerConnection does not handle glare before DLTS negotiation is finished, we will wait for `this._lastStableDescriptionRevision` to be set.

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -954,17 +954,12 @@ class PeerConnectionV2 extends StateMachine {
           return Promise.resolve();
         }
         if (this._peerConnection.signalingState === 'have-local-offer') {
-          // For a peer connection -
+          // Note(mpatwardhan): For a peer connection
           // 1) createOffer always generate SDP with `setup:actpass`
           // 2) when remote description is set `setup:active`  - the answer generated selects the dtls role of setup:passive
           // 3) when remote description is set `setup:passive` - the answer generated selects the dtls role of setup:active
           // 4) when remote description is set `setup:actpass` - the answer generated uses the previously negotiated role (if not negotiated previously setup:active is used)
-          // This test makes the behavior super clear.
-          // https://github.com/twilio/twilio-webrtc.js/blob/master/test/integration/spec/rtcpeerconnection.js#L936
-          // if (this._needsAnswer && this._descriptionRevision === 1) {
-          //   this._queuedDescription = description;
-          //   return Promise.resolve();
-          // }
+          // This test shows the  behavior: https://github.com/twilio/twilio-webrtc.js/blob/master/test/integration/spec/rtcpeerconnection.js#L936
           // with glare handling (if dtls role was not negotiated before ) the generated answer will set setup:active.
           // we do not want that. lets wait for "initial negotiation" before attempting glare handling.
           if (this._needsAnswer && this._lastStableDescriptionRevision === 0) {
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -777,7 +777,6 @@ class PeerConnectionV2 extends StateMachine {
         };
         updatedSdp = this._setSimulcast(updatedSdp, this._sdpFormat, this._trackIdsToAttributes);
       }
-
       return this._setLocalDescription({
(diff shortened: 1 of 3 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -955,7 +954,20 @@ class PeerConnectionV2 extends StateMachine {
...
+          // 4) when remote description is set `setup:actpass` - the answer generated uses the previously negotiated role (if not negotiated previously setup:active is used)
+          // This test makes the behavior super clear.
+          // https://github.com/twilio/twilio-webrtc.js/blob/master/test/integration/spec/rtcpeerconnection.js#L936
+          // if (this._needsAnswer && this._descriptionRevision === 1) {
+          //   this._queuedDescription = description;
+          //   return Promise.resolve();
+          // }
+          // with glare handling (if dtls role was not negotiated before ) the generated answer will set setup:active.
+          // we do not want that. lets wait for "initial negotiation" before attempting glare handling.
+          if (this._needsAnswer && this._lastStableDescriptionRevision === 0) {
```
syerrapragada: Question: Would `this._lastStableDescriptionRevision` be 0 after VMS Failover if at least 1 SDP negotiation already happened? I do not see it being reset to 0 anywhere in the code. More unit/integration tests would help find the answer, I guess

makarandp0: Yes. For new PC, we set this 0.  https://github.com/twilio/twilio-video.js/blob/f300dc0e161b09d4e77d85405b2c95e7834f529c/lib/signaling/v2/peerconnection.js#L207, 

it will be updated only after new PC completes a negotiation.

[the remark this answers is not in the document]
ceaglest: Yes. We had the same bug in C++ where we interpreted revision 1 to mean "DTLS role negotiation" but this had a whole host of problems. We now have a boolean to track the DTLS negotiation process that is independent of revision. I think the usage of _lastStableDescriptionRevision would be equivalent here. 👍

makarandp0:
@ceaglest - I looked at adding more unit tests, but it involves too much code refactoring to be worth it. I think the tests that I have already added are sufficient because we have pretty good test coverage for all new and existing code  (https://3920-46595751-gh.circle-artifacts.com/0/home/circleci/project/coverage/lcov-report/lib/signaling/v2/peerconnection.js.html)
 
But while looking at the tests I realized the pattern we are using for unit tests here (`combinationContext` is causing the test time to grow exponentially. And now causing failures on travis. I have now added some checks to reduce redundant tests.

test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -874,7 +898,18 @@ describe('PeerConnectionV2', () => {
             break;
         }
 
-        rev = test.pcv2._lastStableDescriptionRevision;
+        if (vmsFailOver && initial && type === 'offer' && signalingState === 'have-local-offer') {
+          // in case of vms fail-over, new PC get to 'have-local-offer' state
+          // by VMS sending create-offer message. Which ends up setting a
+          // test.pcv2._descriptionRevision. lets simulate that.
+          lastOfferRevision = 25; // even though its 'initial' state - last offer will not be 1.
```
ceaglest: What if `lastOfferRevision = 0`?

lib/signaling/v2/peerconnection.js
makarandp0: Good point. @innerverse - is revision 0 in RSP a possibility?

innerverse: We always start `create-offer` at rev1.  You should be fine with this.

test/unit/spec/signaling/v2/peerconnection.js
makarandp0: @ceaglest - @innerverse confirmed that RSP starts at revision 0, so we should be safe with assumptions here.

PR #815, opened 2019-11-22
JSDK-2588-filtering out unstable tests (2.x - master)

makarandp0:
This adds support in our workflows to be able to filter-out unstable tests. It also marks currently unstable tests in master as such. 

it uses mocha's `grep` option + a `@unstable` tag to mark tests as unstable. This `grep` option is controlled from the workflow using environment variable  `TEST_STABILITY` which can have values of `all`, `unstable` or `stable`

Note: this just marks the tests in master - we will also repeat this for `1.x` and `develop` branches. 

Example build that run with `TEST_STABLITY=stable` : https://app.circleci.com/github/twilio/twilio-video.js/pipelines/79e2d3c4-3053-4f56-babd-57f0e892b28e/workflows/b039633e-9324-4498-8575-fc1bd31bcfa0

**Contributing to Twilio**


karma/makeconf.js
```diff
@@ -72,15 +72,33 @@ function makeConf(defaultFile, browserNoActivityTimeout, requires) {
...
+        break;
+      case 'stable':
+        mochaOptions.grep = '@unstable';
+        mochaOptions.invert = 1;
+        break;
+      case 'all':
+        // when no grep specified, all tests will run.
+        break;
+      default:
+        console.log('invalid value for TEST_STABILITY:', process.env.TEST_STABILITY)
```
syerrapragada: is TEST_STABILITY mandatory? Instead of logging `invalid value`, we could log `Env not set` or something?

test/integration/spec/connect.js
```diff
@@ -47,6 +47,12 @@ const { trackPriority: { PRIORITY_HIGH, PRIORITY_LOW, PRIORITY_STANDARD } } = re
 
 const safariVersion = isSafari && Number(navigator.userAgent.match(/Version\/([0-9.]+)/)[1]);
 
+// placeholder ensures all variations of TEST_STABILITY end up with some tests to run.
+describe('placeholder', () => {
```
syerrapragada: I am not sure if this is needed? What happens if we do not have a placeholder?

makarandp0: when mocha executes a file and did not find any tests to run, it fails the run. so w/o such placeholder a run with TEST_STABILITY=stable would fail if all tests in a file are marked unstable.

karma/makeconf.js
makarandp0: `TEST_STABILITY` is not mandatory, it defaults to `all` - I will just remove this console.log to avoid the noise.

test/integration/spec/connect.js
makarandp0: Thanks for the suggestion of `failOnEmptyTestSuite=false`. Now we can get rid of this and other placeholder tests.

makarandp0:
Sample run with `test_stablitity=stable`
https://app.circleci.com/github/twilio/twilio-video.js/pipelines/a82f58b4-cc6d-42de-8b58-b154f3674b80/workflows/052106e3-57ac-46a7-bdf0-07e74739fdd2

PR #816, opened 2019-11-25
JSDK-2599 setBandwidthProfile should fail on NaN values where number is expected. 

makarandp0:
We were checking for typeof value === `number` but were not special casing for `NaN` values. Now fixed for numeric values for BWP, as well as for `setNetworkQualityConfiguration`
**Contributing to Twilio**

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -477,7 +477,7 @@ class LocalParticipant extends Participant {
       throw E.INVALID_TYPE('networkQualityConfiguration', 'NetworkQualityConfiguration');
     }
     ['local', 'remote'].forEach(prop => {
-      if (prop in networkQualityConfiguration && typeof networkQualityConfiguration[prop] !== 'number') {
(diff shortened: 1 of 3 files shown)


lib/localparticipant.js
```diff
@@ -477,7 +477,7 @@ class LocalParticipant extends Participant {
       throw E.INVALID_TYPE('networkQualityConfiguration', 'NetworkQualityConfiguration');
     }
     ['local', 'remote'].forEach(prop => {
-      if (prop in networkQualityConfiguration && typeof networkQualityConfiguration[prop] !== 'number') {
+      if (prop in networkQualityConfiguration && (typeof networkQualityConfiguration[prop] !== 'number' || isNaN(networkQualityConfiguration[prop]))) {
```
syerrapragada: Should we create a new function in util and use it here?

makarandp0: it does not look like such utility function would have much reuse. Even in this it won't be very useful in other instance where e check for:
```  
    if (type && typeof value !== type) {
      return E.INVALID_TYPE(`${name}.${prop}`, type);
    }
    if (type === 'number' && isNaN(value)) {
      return E.INVALID_TYPE(`${name}.${prop}`, type);
    }
```

PR #830, opened 2019-12-12
Pointed to a new advisory

anna-vasilko:
Readme still pointed to an old Firefox68 advisory at the top. Edited this section to point to a new advisory and upcoming breaking changes in Chrome.


makarandp0:
@anna-vasilko - Change looks good, but please open this PR against `develop` branch. We cut rcs from develop branch, and releases from master (after merging develop into master). That way this change would show up in any RC that we make from now on.

makarandp0:
@anna-vasilko - never mind about `develop`. Thinking a bit about this - I think for this change it does make sense to be in master, as people would be looking at master more than develop.

PR #833, opened 2019-12-14
JSDK-2615 late arrivals do not subscribe to datatracks on firefox.

makarandp0:
This is caused by a firefox issue. I have [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1603887) to track it, where dataChannel gets closed unexpected after a round of negotiation. This issue is demonstrated by [this simple JSFiddle](https://jsfiddle.net/makarandp/wynouqhp/). 

This change workarounds the issue by always ensuring that `m=application` section exists in every offer/answer on firefox. This avoid an extra round of negotiation that results into data channel getting disconnected. 

Good part is we had same workaround in the code for a while for another firefox issue :) 
Added two tests that fail on firefox w/o this fix.

**Update**
The workaround only works for firefox participant, if chrome/mobile participant connect without data channels to firefox, they would see the same issue.

**Contributing to Twilio**

### DIFF
--- a/COMMON_ISSUES.md
+++ b/COMMON_ISSUES.md
@@ -16,7 +16,7 @@ on iOS 13.0.1 fail to send audio.
 
 Firefox Participants sometimes fail to subscribe to DataTracks on Peer-to-Peer Rooms
 ------------------------------------------------------------------------------------
-Because of this Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1603887) participants that join a Peer-to-Peer Room after a DataTrack has been published by firefox participant fail to subscribe to it. You can workaround this issue by publishing a data track while connecting to a peer-to-peer room. (JSDK-2615)
+Because of this Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1603887) Participants that join a Peer-to-Peer Room after a DataTrack has been published by a Firefox Participant fail to subscribe to it. You can work around this issue by publishing a DataTrack while connecting to the Room. (JSDK-2615)
 
 Working around the browsers' autoplay policy
 --------------------------------------------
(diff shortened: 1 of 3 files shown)


makarandp0:
Thank you for looking at this @syerrapragada. I realized there is an issue with the fix. The fix is about the “other” participant always creating offer with m=application section. So the fix as it works for firefox <-> firefox but  won’t not work when other participant is !firefox.

We could fix it for chrome participants by always creating datachannel, but that still leaves mobile clients.

COMMON_ISSUES.md
```diff
@@ -16,7 +16,7 @@ on iOS 13.0.1 fail to send audio.
 
 Firefox Participants sometimes fail to subscribe to DataTracks on Peer-to-Peer Rooms
 ------------------------------------------------------------------------------------
-We have seen some instances where Firefox Participants that join a Peer-to-Peer Room after a DataTrack has been published fail to subscribe to it. We have filed an internal JIRA bug (JSDK-2615) and started working on a fix, which will be available in a future release.
+Because of this Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1603887) participants that join a Peer-to-Peer Room after a DataTrack has been published by firefox participant fail to subscribe to it. You can workaround this issue by publishing a data track while connecting to a peer-to-peer room. (JSDK-2615)
```
manjeshbhargav:
```suggestion
Because of this Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1603887) Participants that join a Peer-to-Peer Room after a DataTrack has been published by a Firefox Participant fail to subscribe to it. You can work around this issue by publishing a DataTrack while connecting to the Room. (JSDK-2615)
```

lib/signaling/v2/peerconnection.js
```diff
@@ -1039,7 +1030,7 @@ class PeerConnectionV2 extends StateMachine {
       dataTrackSender.addDataChannel(dataChannel);
       this._dataChannels.set(dataTrackSender, dataChannel);
     } catch (error) {
-      // Do nothing.
+      this._log.warn('Error creating data channel', error.message);
```
manjeshbhargav:
```suggestion
      this._log.warn(`Error creating an RTCDataChannel for DataTrack "${dataTrackSender.id}": ${error.message}`);
```

PR #851, opened 2020-01-10
Microsoft Edge + Docs Fix + Safari Warning Refactor

manjeshbhargav:
@syerrapragada 

This PR has the following changes:

* JSDK-2625 - Enable integration tests for Chromium-based Edge.
* JSDK-2633 - `isSupported` should return `true` for Chromium-based Edge.
* JSDK-2567 - Add link to MediaTrackConstraints MDN docs.
* JSDK-2638 - JSDoc for LocalTrackPublication and RemoteTrackPublication should `@extend` TrackPublication.
* JSDK-2635 - Limit experimental support warning for Safari 12.0 and below.
* JSDK-2639 - Remove closed PeerConnections from the StatsReport object.
## Tasks

- [x] Unit tests
- [x] Integration tests
- [x] CHANGELOG.md entry
- [x] +1 from one or more reviewers

**Contributing to Twilio**

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -25,23 +25,11 @@ function isGetUserMediaSupported() {
 }
 
 /**
- * @returns {boolean} - true if the browser is explicitly unsupported.
- *  Note: this function returning false does not mean that browser is supported.
- */
-function isExplicitlyUnsupportedBrowser() {
-  if (typeof navigator !== 'undefined' && typeof navigator.userAgent === 'string') {
-    return /Edg(e)?/.test(navigator.userAgent);
-  }
-  return false;
-}
-
-/**
  * Check if the current environment is supported by the SDK.
  * @returns {boolean}
  */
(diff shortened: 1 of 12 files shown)


lib/util/support.js
```diff
@@ -24,24 +24,12 @@ function isGetUserMediaSupported() {
...
-  }
-  return false;
-}
-
 /**
  * Check if the current environment is supported by the SDK.
  * @returns {boolean}
  */
 function isSupported() {
-  return !isExplicitlyUnsupportedBrowser()
```
syerrapragada: We have 2 flavors of Edge (Chromium & Non-Chromium). Would this return `false` for non chromium based browser?

manjeshbhargav: @syerrapragada Isn't the other browser decommissioned? Also, I think the old browser would return true because WebRTC is not supported. We can test this as part of QA.

PR #858, opened 2020-01-16
JSDK-2633 isSupported returns false for non-Chromium Edge.

manjeshbhargav:
@makarandp0 @syerrapragada 

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.1.0 (in progress)
+2.0.1 (in progress)
 ===================
 
 New Features
 ------------
 
-- twilio-video.js now supports [Chromium-based Edge](https://www.microsoftedgeinsider.com/). (JSDK-2633)
+- `Video.isSupported` now returns `true` for [Chromium-based Edge](https://www.microsoftedgeinsider.com/). (JSDK-2633)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,12 +1,12 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.1.0 (in progress)
+2.0.1 (in progress)
```
manjeshbhargav: @syerrapragada @makarandp0 

Since we already announced Edge support with 2.0.0, the changes in the next version of the SDK will not be as significant. So, I'm converting this into a patch release.

test/unit/spec/util/support.js
```diff
@@ -3,7 +3,7 @@
 const assert = require('assert');
 const isSupported = require('../../../../lib/util/support');
 
-describe('isSupported', () => {
+describe.only('isSupported', () => {
```
makarandp0: remove `.only`

lib/util/support.js
```diff
@@ -24,12 +24,25 @@ function isGetUserMediaSupported() {
     || !!(navigator.mozGetUserMedia);
 }
 
+/**
+ * Check whether the current browser is non-Chromium Edge.
+ * @param {string} browser
+ * @returns {boolean}
+ */
+function isNonChromiumEdge(browser) {
+  return browser === 'chrome' && /Edg(e)?/.test(navigator.userAgent) && (
```
syerrapragada: We could just test for `Edge`? Because we know `Edg` in chromium based?

PR #865, opened 2020-01-27
JSDK-2662 Participant reconnecting state.

manjeshbhargav:
@makarandp0 

This PR implements the new Participant state "reconnecting", and new events "reconnecting" and "reconnected". The IDL changes can be referred to [here](https://code.hq.twilio.com/client/video-sdk-api/pull/57/files).

## Checklist

- [x] RemoteParticipant implementation
- [x] RemoteParticipant unit tests
- [x] LocalParticipant implementation
- [x] LocalParticipant unit tests
- [x] Integration tests
- [x] CHANGELOG.md entry

**Contributing to Twilio**

### DIFF
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -36,6 +36,8 @@ let nInstances = 0;
  *    The {@link Participant}'s {@link VideoTrackPublication}s
  * @emits Participant#disconnected
  * @emits Participant#networkQualityLevelChanged
+ * @emits Participant#reconnected
+ * @emits Participant#reconnecting
  * @emits Participant#trackDimensionsChanged
  * @emits Participant#trackStarted
  */
@@ -357,6 +359,15 @@ class Participant extends EventEmitter {
         participantSignaling.removeListener('stateChanged', stateChanged);
         participantSignaling.removeListener('trackAdded', trackSignalingAdded);
         participantSignaling.removeListener('trackRemoved', trackSignalingRemoved);
+      } else if (state === 'reconnecting') {
+        log.info('reconnecting');
+        self.emit('reconnecting');
+      } else if (state === 'connected') {
+        // NOTE(mmalavalli): Any transition to "connected" here is a result of
+        // successful signaling reconnection, and not a first-time establishment
+        // of the signaling connection.
+        log.info('reconnected');
+        self.emit('reconnected');
       }
     });
   }
@@ -452,6 +463,18 @@ class Participant extends EventEmitter {
  * @param {?NetworkQualityStats} networkQualityStats - The {@link NetworkQualityStats}
  *   based on which {@link NetworkQualityLevel} is calculated, if any
  * @event Participant#networkQualityLevelChanged
+ */
+
+/**
+ * The {@link Participant} has reconnected to the {@link Room} after a signaling connection disruption.
+ * @param {Participant} participant - The {@link Participant} that has reconnected.
+ * @event Participant#reconnected
+ */
+
+/**
+ * The {@link Participant} is reconnecting to the {@link Room} after a signaling connection disruption.
+ * @param {Participant} participant - The {@link Participant} that is reconnecting.
+ * @event Participant#reconnecting
  */
 
 /**
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -29,7 +29,7 @@ let nInstances = 0;
  * @property {?NetworkQualityStats} networkQualityStats - The
  *    {@link Participant}'s current {@link NetworkQualityStats}, if any
  * @property {Participant.SID} sid - The {@link Participant}'s SID
- * @property {string} state - "connected", "disconnected" or "failed"
+ * @property {string} state - "connected", "disconnected" or "reconnecting"
  * @property {Map<Track.SID, TrackPublication>} tracks -
  *    The {@link Participant}'s {@link TrackPublication}s
  * @property {Map<Track.SID, VideoTrackPublication>} videoTracks -
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -464,13 +464,11 @@ class Participant extends EventEmitter {
 
 /**
  * The {@link Participant} has reconnected to the {@link Room} after a signaling connection disruption.
- * @param {Participant} participant - The {@link Participant} that has reconnected.
  * @event Participant#reconnected
  */
 
 /**
  * The {@link Participant} is reconnecting to the {@link Room} after a signaling connection disruption.
- * @param {Participant} participant - The {@link Participant} that is reconnecting.
  * @event Participant#reconnecting
  */
 
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -359,9 +359,6 @@ class Participant extends EventEmitter {
         participantSignaling.removeListener('stateChanged', stateChanged);
         participantSignaling.removeListener('trackAdded', trackSignalingAdded);
         participantSignaling.removeListener('trackRemoved', trackSignalingRemoved);
-      } else if (state === 'reconnecting') {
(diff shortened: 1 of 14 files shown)


lib/signaling/participant.js
```diff
@@ -198,14 +208,30 @@ class ParticipantSignaling extends StateMachine {
...
+  /**
+   * Transition to "reconnecting" state.
+   * @returns {boolean}
+   */
+  reconnecting() {
+    if (this.state === 'connecting' || this.state === 'connected') {
+      this.preempt('reconnecting');
+      return true;
+    }
+    return false;
```
makarandp0: curious - why is this check necessary - in what cases would we get here ?

lib/signaling/participant.js
```diff
@@ -198,14 +208,30 @@ class ParticipantSignaling extends StateMachine {
    * @returns {boolean}
    */
   connect(sid, identity) {
-    if (this.state === 'connecting') {
-      this._sid = sid;
-      this._identity = identity;
+    if (this.state === 'connecting' || this.state === 'reconnecting') {
+      if (!this._sid) {
+        this._sid = sid;
+      }
```
makarandp0:
```suggestion
      } else {
         // we must not get different sid on reconnection.
         assert(this._sid === sid);
      }
```

lib/signaling/participant.js
```diff
@@ -198,14 +208,30 @@ class ParticipantSignaling extends StateMachine {
...
+      if (!this._identity) {
+        this._identity = identity;
+      }
       this.preempt('connected');
       return true;
     }
     return false;
```
makarandp0: is this case possible - that we get connected when not 'connecting' or 'reconnecting' ? Maybe it's worth `log.error("unexpected connect message")`

lib/signaling/v2/room.js
```diff
@@ -659,13 +662,17 @@ function handleLocalParticipantEvents(roomV2, localParticipant) {
...
 
-  roomV2.on('stateChanged', function stateChanged(state) {
+  roomV2.on('stateChanged', function stateChanged(state, error) {
     if (state === 'disconnected') {
       localParticipant.removeListener('trackAdded', renegotiate);
       localParticipant.removeListener('trackRemoved', renegotiate);
       localParticipant.removeListener('updated', localParticipantUpdated);
       roomV2.removeListener('stateChanged', stateChanged);
       localParticipant.disconnect();
+    } else if (state === 'reconnecting' && error instanceof SignalingConnectionDisconnectedError) {
```
makarandp0: is it possible to have `(state === 'reconnecting && !error)` - it would be useful to assert or log.error in such case.

lib/signaling/participant.js
```diff
@@ -12,14 +12,25 @@ ParticipantSignaling States
     | connecting |---->| connected |---->| disconnected |
     |            |     |           |     |              |
     +------------+     +-----------+     +--------------+
-
+                           | ^                    ^
```
makarandp0: I think we need an arrow from `connecting` to `reconnecting`.

lib/participant.js
```diff
@@ -454,6 +462,18 @@ class Participant extends EventEmitter {
...
+/**
+ * The {@link Participant} has reconnected to the {@link Room} after a signaling connection disruption.
+ * @param {Participant} participant - The {@link Participant} that has reconnected.
+ * @event Participant#reconnected
+ */
+
+/**
+ * The {@link Participant} is reconnecting to the {@link Room} after a signaling connection disruption.
+ * @param {Participant} participant - The {@link Participant} that is reconnecting.
+ * @event Participant#reconnecting
```
makarandp0: Are these 'reconnecting' 'reconnected' events applicable for LocalParticipant as well? 
Isn't local participant reconnecting is same as room reconnecting/ reconnected ?

lib/participant.js
```diff
@@ -357,6 +359,12 @@ class Participant extends EventEmitter {
         participantSignaling.removeListener('stateChanged', stateChanged);
         participantSignaling.removeListener('trackAdded', trackSignalingAdded);
         participantSignaling.removeListener('trackRemoved', trackSignalingRemoved);
+      } else if (state === 'connected') {
+        // NOTE(mmalavalli): Any transition to "connected" here is a result of
+        // successful signaling reconnection, and not a first-time establishment
+        // of the signaling connection.
+        log.info('reconnected');
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Actually, this function handle TrackSignaling events on the Participant, i.e. "trackSignalingAdded", "trackSignalingRemoved". So this addition is appropriate in here.

lib/participant.js
manjeshbhargav: @makarandp0 Yes, they do apply to LocalParticipant and they are only fired if the reconnection is due to Signaling interruption.

lib/signaling/participant.js
manjeshbhargav: @makarandp0 "connecting" immediately transitions to "connected", as soon as the ParticipantSignaling object is created. So there is no possibility of the transition you are talking about here.

manjeshbhargav: @makarandp0 I need to remove "reconnecting" from the list of states that are allowed to transition from "connecting".

lib/signaling/participant.js
manjeshbhargav: @makarandp0 I don't think this is necessary, since we're sure that the SID won't change.

lib/signaling/participant.js
manjeshbhargav: @makarandp0 No, it is not possible, unless there is a bug in Room Service. So I don't think we need to log the error.

lib/signaling/participant.js
manjeshbhargav: @makarandp0 Transitions that are not permitted will result in an exception. Hence the check.

lib/signaling/v2/room.js
manjeshbhargav: @makarandp0 We need to make sure LocalParticipantV2 transitions to "reconnecting" only if the `error` is an instance of `SignalingConnectionDisconnectedError` (signaling connection interruption).

lib/signaling/participant.js
makarandp0: I know..but I think assertions can help make this assumptions very explicit, and catch issues if they are broken.

lib/signaling/participant.js
makarandp0: yeah it makes sense that it's not possible, I would suggest we remove the if statement that guards for it
`if (this.state === 'connecting' || this.state === 'reconnecting') {`  
But if we are keeping it - which is okay too - an assertion or log.error helps make it clear to reader that this is not expected.

lib/signaling/participant.js
makarandp0: in that case, same point - we should log error for unexpected cases, since we are handling it anyways.

lib/signaling/v2/room.js
makarandp0: ah make sense. maybe then handling the "other" case with a comment would help.

makarandp0:
@manjeshbhargav, I think I have few comments about the same thing - 
when we guard or handle against an unexpected/impossible case - it helps to make the assumption explicit with error.log or an assert and comment - so that reader won't be wondering why are we handling such case.  

But I'll let you decide if you do now want to add such assertions / error logs. LGTM but we should add integration tests before merging.

manjeshbhargav:
@makarandp0 ,

Regarding logging unexpected situations, I agree in general, but right now the log module is not available in ParticipantSignaling class. Once we add logging capability to signaling classes, we can add the extra logs then.

PR #870, opened 2020-01-30
Trigger for CircleCI QE Tests

DenissSazanovs:
CircleCI config to automatically trigger QE tests.

**Contributing to Twilio**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,6 +230,17 @@ jobs:
     environment:
       BROWSER: << parameters.browser >>
       BVER: << parameters.bver >>
+    trigger-qe-tests:
+      docker:
+        - image: circleci/node:latest
+      steps:
+        - run:
+            name: Trigger QE tests
+            command: |
+              curl --silent --output /dev/null -u USER_TOKEN: -X POST \
+              --header "Content-Type: application/json" \
+              -d '{"branch":"v${CIRCLE_TAG:0:1}","parameters":{"sdk_version": CIRCLE_TAG,"is_rc":true}}' \
+              $SDKS_QE_CIRCLECI_VOICE_JS_TESTS_ENDPOINT
 
 workflows:
   version: 2
@@ -422,3 +433,14 @@ workflows:
             - "firefox stable group"
             - "chrome stable p2p"
             - "chrome stable group"
+  Release-candidate:
+    jobs:
+      - trigger-qe-tests:
+          filters:
+            tags:
+              only:
+                - /^\d+\.\d+\.\d+-rc\d+$/
+                - /^\d+\.\d+\.\d+-preview\d+-rc\d+$/
+                - /^\d+\.\d+\.\d+-beta\d+-rc\d+$/
+            branches:
+              ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -443,5 +443,7 @@ workflows:
                 - /^\d+\.\d+\.\d+-preview\d+-rc\d+$/
                 - /^\d+\.\d+\.\d+-beta\d+-rc\d+$/
                 - /^\d+\.\d+\.\d+-rc\.\d+$/
+                - /^\d+\.\d+\.\d+-preview\d+-rc\.\d+$/
+                - /^\d+\.\d+\.\d+-beta\d+-rc\.\d+$/
             branches:
               ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -442,5 +442,6 @@ workflows:
                 - /^\d+\.\d+\.\d+-rc\d+$/
                 - /^\d+\.\d+\.\d+-preview\d+-rc\d+$/
                 - /^\d+\.\d+\.\d+-beta\d+-rc\d+$/
+                - /^\d+\.\d+\.\d+-rc\.\d+$/
             branches:
               ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,13 +231,14 @@ jobs:
       BROWSER: << parameters.browser >>
       BVER: << parameters.bver >>
     trigger-qe-tests:
+      executor: generic-executor
       docker:
         - image: circleci/node:latest
       steps:
         - run:
             name: Trigger QE tests
             command: |
-              curl --silent --output /dev/null -u USER_TOKEN: -X POST \
+              curl --silent --output /dev/null -u SDKSQE_CIRCLECI_API_TOKEN: -X POST \
               --header "Content-Type: application/json" \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -422,3 +433,14 @@ workflows:
             - "firefox stable group"
             - "chrome stable p2p"
             - "chrome stable group"
+  Release-candidate:
+    jobs:
+      - trigger-qe-tests:
+          filters:
+            tags:
+              only:
+                - /^\d+\.\d+\.\d+-rc\d+$/
```
makarandp0: please include `/^\d+\.\d+\.\d+-rc\.\d+$/` in the matching tags as well.
So far we have been using `1.20.1-rc1` format which is not semvar compliant. We are going to move to using `1.20.1-rc.1` which is correct usage.

makarandp0: same goes for beta and preview tags as well.

.circleci/config.yml
```diff
@@ -230,6 +230,17 @@ jobs:
...
+    trigger-qe-tests:
+      docker:
+        - image: circleci/node:latest
+      steps:
+        - run:
+            name: Trigger QE tests
+            command: |
+              curl --silent --output /dev/null -u USER_TOKEN: -X POST \
+              --header "Content-Type: application/json" \
+              -d '{"branch":"v${CIRCLE_TAG:0:1}","parameters":{"sdk_version": CIRCLE_TAG,"is_rc":true}}' \
```
makarandp0: I have not experimented using circleci inbuilt variables in commands - but don't you need `$CIRCLE_TAG` instead of `CIRCLE_TAG` ?

.circleci/config.yml
```diff
@@ -230,6 +230,17 @@ jobs:
     environment:
       BROWSER: << parameters.browser >>
       BVER: << parameters.bver >>
+    trigger-qe-tests:
+      docker:
```
makarandp0: NIT: you just want to execute the curl command right? see if you can use 
`executor: generic-executor`

in this docker file I have tried to limit the number of executors by declaring it upfront at top.

.circleci/config.yml
```diff
@@ -230,6 +230,17 @@ jobs:
...
+        - image: circleci/node:latest
+      steps:
+        - run:
+            name: Trigger QE tests
+            command: |
+              curl --silent --output /dev/null -u USER_TOKEN: -X POST \
```
makarandp0: where does the USE_TOKEN come from? it does not seem to be one of the in-built or custom variables. Also don't you need $USER_TOKEN ?

.circleci/config.yml
makarandp0: how does `"branch":"v${CIRCLE_TAG:0:1}` work - I assume it will resolve to` "branch": "v1"` for the tag `1.20.1-rc1` but we do not have a branched named like that.

.circleci/config.yml
syerrapragada: @DenissSazanovs You need to add a circle ci token to the `sdksqe` context with access to Video QE Slave repo. Please work with @ruescar to setup one.

.circleci/config.yml
```diff
@@ -230,6 +230,17 @@ jobs:
...
+              --header "Content-Type: application/json" \
+              -d '{"branch":"v${CIRCLE_TAG:0:1}","parameters":{"sdk_version": CIRCLE_TAG,"is_rc":true}}' \
+              $SDKS_QE_CIRCLECI_VOICE_JS_TESTS_ENDPOINT
```
syerrapragada: Please change the name from `SDKS_QE_CIRCLECI_VOICE_JS_TESTS_ENDPOINT` -> `SDKS_QE_CIRCLECI_VIDEO_JS_TESTS_ENDPOINT`. You also need to add this env variable to `sdksqe` context. This url should be pointing to the video qe slave repo.
ex: For voice it is https://circleci.com/api/v2/project/gh/twilio/sdks-qe-voice-js-orchestra-slave/pipeline

.circleci/config.yml
syerrapragada: @makarandp0 For every major version QE has a corresponding slave branch version. This command should trigger workflow on that job.

.circleci/config.yml
syerrapragada: Is this change happening @makarandp0?

.circleci/config.yml
DenissSazanovs: This is done already. Both variables are added

ruescar: SDKSQE_CIRCLECI_API_TOKEN has been added to sdks-qe context but won't work until sdksqe account is given Write permission on this project @syerrapragada

ruescar: @DenissSazanovs You should now be able to use the above variable

ruescar: sdksqe account has already been added to this project

.circleci/config.yml
```diff
@@ -231,16 +231,17 @@ jobs:
...
         - run:
             name: Trigger QE tests
             command: |
-              curl --silent --output /dev/null -u USER_TOKEN: -X POST \
+              curl --silent --output /dev/null -u CIRCLECI_PERSONAL_API_TOKEN: -X POST \
               --header "Content-Type: application/json" \
-              -d '{"branch":"v${CIRCLE_TAG:0:1}","parameters":{"sdk_version": CIRCLE_TAG,"is_rc":true}}' \
-              $SDKS_QE_CIRCLECI_VOICE_JS_TESTS_ENDPOINT
+              -d '{"branch":"'v${CIRCLE_TAG:0:1}'","parameters":{"sdk_version":"'$CIRCLE_TAG'","is_rc":true}}' \
+              $SDK_JS_VIDEO_SLACK_WEBHOOK
```
syerrapragada: Is this variable in SDKSQE Context? Could you please move it if not the case?

.circleci/config.yml
```diff
@@ -230,6 +230,18 @@ jobs:
     environment:
       BROWSER: << parameters.browser >>
       BVER: << parameters.bver >>
+    trigger-qe-tests:
```
makarandp0: @DenissSazanovs, Looks like this indentation for this block is off causing it to fail the build (https://circleci.com/gh/twilio/twilio-video.js/10107?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Please fix the indentation, and verify that the build passes.

.circleci/config.yml
DenissSazanovs: Yes, it's in context

.circleci/config.yml
DenissSazanovs: Fixed

Issue #869, opened 2020-01-29
2.0.1 crashing on Chrome iOS


kylealwyn:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

Thanks for the work on this library - it's been a pleasure to use so far. We're experiencing an issue on Chrome iOS with 2.0.1. Looks like it's related to the new browser detection logic added in https://github.com/twilio/twilio-video.js/commit/5c1b33ccb448801fdfe2a0320551cb343dfd3a8d#diff-90df8a4fdd65da9a7dabb2c41927a14dR39. Let me know if I can provide any other details!

<img src="https://user-images.githubusercontent.com/14184138/73384249-bdc05600-427f-11ea-85c0-6e6f17c4922a.png" width="400" />

``

**Expected behavior:**

twilio-video doesn't immediately crash the page on load on Chrome iOS

**Actual behavior:**

twilio-video immediately crashes the page on load on Chrome iOS

**Software versions:**

 - [x] Browser(s): Chrome
 - [x] Operating System: iOS
 - [x] twilio-video.js: 2.0.1
 - [x] Third-party libraries (e.g., Angular, React, etc.): React 16

douglasrcjames:
Bumping this, was about to post the same error effecting my whole site, not just the component. 

A few extra notes:
-  Error also pops up in Firefox browser on iOS, but not on Safari or Edge in iOS
-  Switch the site to show desktop site, it will load the whole site as expected
-  Removing the import statement in my React component for 'twilio-video' will allow app to load as expected

manjeshbhargav:
Hi @douglasrcjames ,

Thanks for writing in with your issue. I'm sorry that you're being blocked by this. To start off, can you please share the `navigator.userAgent` string that you see for Chrome iOS?

Thanks,

[...]

PR #874, opened 2020-02-05
JSDK-2670 Handle Chrome iOS user agent, which does not contain "Version/*".

manjeshbhargav:
@makarandp0 

This PR fixes JSDK-2670, which was created due to #869 . Please confirm that your local version of Simpler Signaling does not crash with this change on Chrome and Firefox on iOS (iPhone and iPad).

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { MediaStreamTrack } = require('@twilio/webrtc');
-const { guessBrowser } = require('@twilio/webrtc/lib/util');
+const { guessBrowser, guessBrowserVersion } = require('@twilio/webrtc/lib/util');
 const createCancelableRoomPromise = require('./cancelableroompromise');
 const createLocalTracks = require('./createlocaltracks');
 const EncodingParametersImpl = require('./encodingparameters');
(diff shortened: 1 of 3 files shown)


lib/connect.js
```diff
@@ -36,8 +36,9 @@ const CancelablePromise = require('./util/cancelablepromise');
 const Log = require('./util/log');
 const { validateBandwidthProfile } = require('./util/validate');
 
-const safariVersion = guessBrowser() === 'safari'
-  && navigator.userAgent.match(/Version\/([^\s]+)/)[1];
+const safariVersion = guessBrowser() === 'safari' && (
```
makarandp0: @manjeshbhargav, 

We should add unit tests for this. I suggest move this into a seperate file (Perhaps into `test/lib/guessbrowser.js`) where it can be unit tested against offending user agent -  and export more structured version from there.   

looks like we have another problem that guessBrowser is returning 'safari' for this Chrome on iOS.
@manjeshbhargav - I think we should also file a task for consolidating our version checks, We have lot of them sprinkled around.

manjeshbhargav: @makarandp0 Pretty soon we will have to refactor our browser detection logic to no longer use `userAgent`, since it will be frozen and deprecated. So we can do this then.

makarandp0: @manjeshbhargav - That's a good point, but  u/a change is chrome only (for now) and it's many months away. Also this work of consolidating the version check will also prepare us for eventual removal of useragent. Once we are done with this consolidation user agent removal will be much simpler localized change.

PR #894, opened 2020-02-25
do not emit public events with statemachine locked

makarandp0:
With this fix, we fire some events (`reconnecting`, `reconnected`) asynchronously - ensuring that our StateMachine lock is not held, so that there would be no side effects by caller synchronously calling into state machine again.

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
 New Features
 ------------
 
+- `reconnecting` and `reconnected` events on Room and LocalParticipant are now fired asynchronously. (JSDK-2696)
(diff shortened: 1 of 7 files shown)


makarandp0:
@manjeshbhargav, Thanks for discussion on this yesterday. As we discussed, the problem is
- we are triggering some public events synchronously while in state transition. Before sending such notification we hold a state Machine lock (a) When user app calls back for say - room.disconnect  or any other function that might change the state. If that function happens to use `StateMachine.preempt` ( like room.disconnect does ), it ends up releasing lock held by (a).    
and when notification is done with when (a) tries to release the lock, StateMachine does not like it as lock was released already (by preempt)

fixing this issue by letting 'room.disconnect` be called while lock is held - seem  dangerous. With such fix, we would have to ensure that the  code initiating the events is preempt safe.  

On the other hand emitting the `reconnecting`, `reconnected` events async is slight change of behavior - but that is something caller must not depend on anyways.

PR #904, opened 2020-03-06
JSDK-2685 Add mediaRegion property to Room

PikaJoyce:
Due to some unforeseen changes, I removed the implementation of ConnectOptions.mediaRegion. Instead, this PR contains the implementation for Room.mediaRegion along with the appropriate integration tests.

WORK DONE : 
- [x] Integration tests
- [x] Bubbling up server response to client on Room.mediaRegion
- [x] Fixing tests

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -354,6 +354,8 @@ function connect(token, options) {
  *   This has no effect in Peer-to-Peer Rooms.
  * @property {boolean} [dominantSpeaker=false] - Whether to enable the Dominant
  *   Speaker API or not. This only takes effect in Group Rooms.
+ * @property {string} [mediaRegion=gll] - Preferred media region; By default, you will be connected to the
+ *   nearest media server determined by latency based routing.
  * @property {boolean} [dscpTagging=false] - <code>(deprecated: use "enableDscp" instead)</code>
  *   DSCP tagging allows you to request enhanced QoS treatment for RTP media packets from any
  *   firewall that the client may be behind. Setting this option to <code>true</code> will
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -354,11 +354,6 @@ function connect(token, options) {
  *   This has no effect in Peer-to-Peer Rooms.
  * @property {boolean} [dominantSpeaker=false] - Whether to enable the Dominant
  *   Speaker API or not. This only takes effect in Group Rooms.
- * @property {string} [mediaRegion=gll] - Allows the Participant to choose the region that is used
- *   for media processing in a Group Room. In order for the property to be honored the Participant
- *   must be the first to connect, and the Room must be either Ad-Hoc with the account's default
- *   media region set to "gll", or created with the "gll" media region via the REST API.
- *   Requests to select a media region are ignored for Peer-to-Peer Rooms, because they do not use a central media server.
  * @property {boolean} [dscpTagging=false] - <code>(deprecated: use "enableDscp" instead)</code>
  *   DSCP tagging allows you to request enhanced QoS treatment for RTP media packets from any
  *   firewall that the client may be behind. Setting this option to <code>true</code> will
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -354,8 +354,11 @@ function connect(token, options) {
  *   This has no effect in Peer-to-Peer Rooms.
  * @property {boolean} [dominantSpeaker=false] - Whether to enable the Dominant
  *   Speaker API or not. This only takes effect in Group Rooms.
- * @property {string} [mediaRegion=gll] - Preferred media region; By default, you will be connected to the
- *   nearest media server determined by latency based routing.
+ * @property {string} [mediaRegion=gll] - Allows the Participant to choose the region that is used
+ *   for media processing in a Group Room. In order for the property to be honored the Participant
+ *   must be the first to connect, and the Room must be either Ad-Hoc with the account's default
+ *   media region set to "gll", or created with the "gll" media region via the REST API.
+ *   Requests to select a media region are ignored for Peer-to-Peer Rooms, because they do not use a central media server.
  * @property {boolean} [dscpTagging=false] - <code>(deprecated: use "enableDscp" instead)</code>
  *   DSCP tagging allows you to request enhanced QoS treatment for RTP media packets from any
  *   firewall that the client may be behind. Setting this option to <code>true</code> will
--- a/lib/room.js
+++ b/lib/room.js
@@ -110,6 +110,10 @@ class Room extends EventEmitter {
         get() {
           return signaling.state;
         }
+      },
+      mediaRegion: {
+        enumerable: true,
+        value: signaling.mediaRegion
       }
     });
 
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -139,7 +139,7 @@ class RoomV2 extends RoomSignaling {
       },
       mediaRegion: {
         enumerable: true,
-        value: initialState.options.media_region
+        value: initialState.options.media_region || null
       }
     });
 
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -136,6 +136,10 @@ class RoomV2 extends RoomSignaling {
       },
       _trackReceiverDeferreds: {
         value: new Map()
+      },
+      mediaRegion: {
+        enumerable: true,
+        value: initialState.options.media_region
       }
     });
 
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -230,6 +230,12 @@ class TwilioConnectionTransport extends StateMachine {
         user_agent: this._userAgent
       };
 
+      if (typeof this._options.mediaRegion === 'string') {
+        message.options = {
+          media_region: this._options.mediaRegion,
+        };
+      }
+
       if (this._bandwidthProfile) {
         message.bandwidth_profile = createBandwidthProfilePayload(
           this._bandwidthProfile);
@@ -255,6 +261,7 @@ class TwilioConnectionTransport extends StateMachine {
     } else if (message.type === 'update') {
       message.session = this._session;
     }
+
     this._twilioConnection.sendMessage(message);
   }
 
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -230,12 +230,6 @@ class TwilioConnectionTransport extends StateMachine {
         user_agent: this._userAgent
       };
 
-      if (typeof this._options.mediaRegion === 'string') {
-        message.options = {
-          media_region: this._options.mediaRegion,
-        };
-      }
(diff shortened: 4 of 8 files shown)


test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+      if (isInvalidRegion) {
+        scenario += `with an invalid mediaRegion : ${mediaRegion}`;
+      } else if (mediaRegion === 'without') {
+        scenario += 'without a mediaRegion ';
+      } else {
+        scenario += `with a valid mediaRegion: ${mediaRegion}`;
+      }
+
+      context(scenario, () => {
+        it(`should return a CancelablePromise that ${isInvalidRegion ? 'rejects with a MediaServerRemoteDescFailedError if the room is not p2p' : 'resolves with a Room'}`, async () => {
```
manjeshbhargav:
```suggestion
        it(`should return a CancelablePromise that ${isInvalidRegion && defaults.topology !== 'peer-to-peer' ? 'rejects with a MediaServerRemoteDescFailedError' : 'resolves with a Room'}`, async () => {
```

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+              if (defaults.topology === 'peer-to-peer') {
+                assert.equal(room.mediaRegion, null);
+              }
+              if (defaults.topology !== 'peer-to-peer') {
+                if (['without'].includes(mediaRegion)) {
+                  assert.equal(typeof room.mediaRegion, 'string');
+                } else {
+                  assert.equal(room.mediaRegion, mediaRegion);
+                }
+              }
```
manjeshbhargav:
```js
if (defaults.topology === 'peer-to-peer') {
  assert.equal(room.mediaRegion, null);
} else if (isInvalidRegion) {
  assert(error instanceof MediaServerRemoteDescFailedError);
} else if (['gll', 'without'].includes(mediaRegion)) {
  assert.equal(typeof room.mediaRegion, 'string');
} else {
  assert.equal(room.mediaRegion, mediaRegion);
}
```

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+              assert(room instanceof Room);
```
manjeshbhargav:
```suggestion
```

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+          let room = null;
+          try {
+            room = await cancelablePromise;
+          } catch (error_) {
+            error = error_;
+          } finally {
+            if (room) {
+              room.disconnect();
+              await completeRoom(sid);
+            }
```
manjeshbhargav: Please disconnect from the Room in an `afterEach()` function.

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
     });
   });
 
+  describe.only('media region', () => {
```
manjeshbhargav:
```suggestion
  describe('media region', () => {
```

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+              }
+              assert(room instanceof Room);
+            }
+          }
+        });
+      });
+    });
+
+    if (defaults.topology !== 'peer-to-peer') {
+      it('should allow the first participant to set the Room\'s media region', async () => {
```
manjeshbhargav: I don't think this test is required, since you're already doing this above.

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+          assert.throws(assert.equal(error, roomRegion));
+          room.disconnect();
+        } finally {
+          if (room) {
+            room.disconnect();
+          }
+        }
+      });
+
+      it('should have another participant match the media region of the first participant regardless of the parameter being passed in', async () => {
```
manjeshbhargav: This test can re-use the first participant in the previous test.

lib/room.js
```diff
@@ -110,6 +110,10 @@ class Room extends EventEmitter {
         get() {
           return signaling.state;
         }
+      },
+      mediaRegion: {
```
manjeshbhargav: Please document this class property.

lib/connect.js
```diff
@@ -354,6 +354,8 @@ function connect(token, options) {
  *   This has no effect in Peer-to-Peer Rooms.
  * @property {boolean} [dominantSpeaker=false] - Whether to enable the Dominant
  *   Speaker API or not. This only takes effect in Group Rooms.
+ * @property {string} [mediaRegion=gll] - Preferred media region; By default, you will be connected to the
```
manjeshbhargav: Please request a review from Chris for the documentation text, as well as the documentation for `Room.mediaRegion`.

lib/room.js
PikaJoyce: Added to CHANGELOG.md !

test/integration/spec/connect.js
PikaJoyce: Reuse the first participant from the tests ran in the context block? Do you think I should just move the entire remote participant it block into the context block?

test/integration/spec/connect.js
PikaJoyce: Will remove this!

lib/room.js
ceaglest: I believe this is where the public docs for Room.mediaRegion should live. If so, the docs should match: https://code.hq.twilio.com/pages/client/video-sdk-api/idl/room.html

lib/connect.js
ceaglest: The docs should explain how the Participant's mediaRegion preference is used. Also, the default value is specified as null in the IDL.

We didn't ship it on mobile yet, but ConnectOptions.mediaRegion in the [IDL](https://code.hq.twilio.com/pages/client/video-sdk-api/idl/connectoptions.html) has a pretty comprehensive write up on how the option is interpreted.

lib/signaling/v2/room.js
```diff
@@ -136,6 +136,10 @@ class RoomV2 extends RoomSignaling {
       },
       _trackReceiverDeferreds: {
         value: new Map()
+      },
+      mediaRegion: {
+        enumerable: true,
+        value: initialState.options.media_region || null
       }
```
ceaglest: Nice! The benefit of constructing RoomSignaling late.

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+            if (room) {
+              room.disconnect();
+              await completeRoom(sid);
+            }
+            if (isInvalidRegion) {
+              if (defaults.topology !== 'peer-to-peer') {
+                assert(error instanceof MediaServerRemoteDescFailedError, `Connected to Room ${room && room.sid} with an invalid mediaRegion "${mediaRegion}"`);
+              } else {
+                assert.equal(room.mediaRegion, null);
+              }
```
ceaglest: Returning MediaServerRemoteDescFailedError when the region is invalid feels like the wrong behavior, but I understand that this is a backend and not an SDK limitation. Ideally, the Server would return [MediaRegionInvalidError](https://code.hq.twilio.com/client/twilio-video-errors/blob/master/twilio-video-errors.yaml#L200) for this error case. Perhaps its worth revisiting this now that it we are adding this new control plane to the SDK?

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+    let roomName = null;
+
+    beforeEach(async () => {
+      const identity = randomName();
+      token = getToken(identity);
+      sid = await createRoom(randomName(), defaults.topology, { MediaRegion: 'gll' });
+    });
+
+    const invalidRegions = ['foo', '34', '$abc', 'abc!'];
+    const validRegions = defaults.regions ? defaults.regions.split(',') : ['au1', 'br1', 'de1', 'ie1', 'in1', 'jp1', 'sg1', 'us1', 'us2'];
```
ceaglest: Are these tests ever intended to be run in other environments than prod? The regions listed are valid in prod, but not in dev or stage which support a subset of this list.

test/integration/spec/connect.js
```diff
@@ -203,6 +204,119 @@ describe('connect', function() {
...
+  describe.only('media region', () => {
+    let sid = null;
+    let token = null;
+    let roomName = null;
+
+    beforeEach(async () => {
+      const identity = randomName();
+      token = getToken(identity);
+      sid = await createRoom(randomName(), defaults.topology, { MediaRegion: 'gll' });
+    });
```
ceaglest: I guess it will not be possible to test Ad-hoc Room creation unless the account under test has "gll" selected by default. Should we be adding tests for it? I did not add coverage of Ad-Hoc in C++ because of the complexity of having tests require credentials for more than 1 account SID.

lib/connect.js
PikaJoyce: Got it, I'll just use the text from the writeup for this then! Thanks :)

test/integration/spec/connect.js
PikaJoyce: I agree. I had originally implemented this test with MediaRegionInvalidError but when I got an actual error back, it was not the case. There is a thread where I asked Seth/Yuri this question and I've filed a ticket to have the documentation adjusted. 

Alternatively, I believe that this error could could also be implemented so that we wouldn't need to make changes to the documentation. I'm just not sure how to best go about this since this test seems biased at this point?

lib/room.js
manjeshbhargav: @PikaJoyce Please update the documentation before resolving the conversation.

test/integration/spec/connect.js
manjeshbhargav: @ceaglest Yes. We would have to manually change the default media region in the console for the Account SID.

test/integration/spec/connect.js
manjeshbhargav: @ceaglest We take in an environment variable `REGIONS=rg1,rg2,...` to override the list of regions in non-prod environments.

test/integration/spec/connect.js
manjeshbhargav: @PikaJoyce Please resolve only after pushing the changes requested.

test/integration/spec/connect.js
manjeshbhargav: @PikaJoyce Please resolve only after pushing the changes.

test/integration/spec/connect.js
manjeshbhargav: @PikaJoyce Please resolve only after pushing the changes.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -255,6 +255,7 @@ class TwilioConnectionTransport extends StateMachine {
     } else if (message.type === 'update') {
       message.session = this._session;
     }
+
```
manjeshbhargav:
```suggestion
```

PR #905, opened 2020-03-06
JSDK-2683  use RSP session timeout for reconnect attempts

makarandp0:
With this change we use session_timeout specified in RSP connected message. 

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ New Features
 
 - `reconnecting` and `reconnected` events on Room and LocalParticipant are now fired asynchronously. (JSDK-2696)
 - twilio-video.js now raises `connect()` errors due to network disruptions quicker by not retrying after the first connection attempt fails. (JSDK-2682)
-- twilio-video.js attempts to reconnect to a Room only while the Participant's session is valid (typically 30 seconds after the `reconnecting` event) instead of using a fixed number of retries.
+- twilio-video.js attempts to reconnect to a Room only while the Participant's session is valid (typically 30 seconds after the `reconnecting` event) instead of using a fixed number of retries. (JSDK-2683)
 - A LocalParticipant will now have an additional `signalingRegion` property which contains the geographical region of the signaling edge LocalParticipant is connected to. (JSDK-2687)
 
 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ New Features
 
 - `reconnecting` and `reconnected` events on Room and LocalParticipant are now fired asynchronously. (JSDK-2696)
 - twilio-video.js now raises `connect()` errors due to network disruptions quicker by not retrying after the first connection attempt fails. (JSDK-2682)
+- twilio-video.js now uses server supplied timeout value to determine reconnect attempts in case of loss of signaling connection.
 - A LocalParticipant will now have an additional `signalingRegion` property which contains the geographical region of the signaling edge LocalParticipant is connected to. (JSDK-2687)
 
 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ New Features
 
 - `reconnecting` and `reconnected` events on Room and LocalParticipant are now fired asynchronously. (JSDK-2696)
 - twilio-video.js now raises `connect()` errors due to network disruptions quicker by not retrying after the first connection attempt fails. (JSDK-2682)
-- twilio-video.js now uses server supplied timeout value to determine reconnect attempts in case of loss of signaling connection.
+- twilio-video.js attempts to reconnect to a Room only while the Participant's session is valid (typically 30 seconds after the `reconnecting` event) instead of using a fixed number of retries.
 - A LocalParticipant will now have an additional `signalingRegion` property which contains the geographical region of the signaling edge LocalParticipant is connected to. (JSDK-2687)
 
 Bug Fixes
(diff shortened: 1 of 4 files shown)


lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -324,6 +330,61 @@ class TwilioConnectionTransport extends StateMachine {
...
+    if (this._sessionTimeoutMS === 0) {
+      // this means either we have never connected.
+      // or we timed out while trying to reconnect
+      // In either case we do not want to reconnect.
+      return null;
+    }
+
+    // start session timer
+    if (!this._sessionTimer) {
+      this._sessionTimer = setTimeout(() => {
```
manjeshbhargav: Please use the Timeout class in lib/util.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -324,6 +330,61 @@ class TwilioConnectionTransport extends StateMachine {
...
+        // called while we were waiting.
+        if (this._sessionTimer) {
+          // do not allow any more reconnect attempts.
+          this._sessionTimeoutMS = 0;
+        }
+      }, this._sessionTimeoutMS);
+    }
+
+    // return promise that waits with exponential backoff.
+    this._reconnectAttempt++;
```
manjeshbhargav: I think the backoff logic should be the same as the ICE restart backoff logic in PeerConnectionV2.

lib/signaling/v2/twilioconnectiontransport.js
makarandp0: I looked into it, but the implementation of Timeout calls `.clear()` before calling callback function. That does not work in this case because we need to check here if the timer has been cleared while we were waiting. 

https://github.com/twilio/twilio-video.js/blob/master/lib/util/timeout.js#L67

we could workaround it, by creating a new timer object everytime, but I think it's much easier w/o the `Timeout` object

lib/signaling/v2/twilioconnectiontransport.js
makarandp0: Thanks, I have updated to use `Backoff` module. Can you take another look ?

makarandp0:
@manjeshbhargav, I have addressed your feeback. can you pls take another look ?

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 Instead of checking whether the timeout object exists, you can use a private boolean property:
```js
this._sessionTimedOut = false;
this._sessionTimeout = new Timeout(() => this._sessionTimedOut = true, SESSION_TIMEOUT_MS);
```

makarandp0: @manjeshbhargav - I agree we could make `Timeout` work by introducing additional variable, but I am not sure that is better solution. I feel basic `setTimeout`  does job effectively here.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -410,27 +464,18 @@ function setupEventListeners(transport) {
       transport.disconnect();
       return;
     }
-    if (transport._reconnectAttemptsLeft <= 0) {
+
+    const reconnectTimer = transport._getReconnectTimer();
+    if (!reconnectTimer) {
```
manjeshbhargav: @makarandp0 I think this ensures that if the timeout happens in the middle of a reconnect attempt, then the attempt is allowed to finish. Is that correct?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ New Features
 
 - `reconnecting` and `reconnected` events on Room and LocalParticipant are now fired asynchronously. (JSDK-2696)
 - twilio-video.js now raises `connect()` errors due to network disruptions quicker by not retrying after the first connection attempt fails. (JSDK-2682)
+- twilio-video.js now uses server supplied timeout value to determine reconnect attempts in case of loss of signaling connection.
```
manjeshbhargav: @makarandp0 Please ask @ceaglest to verify the language here, since he came up with the proposal. Also, please add the ticket number.

lib/signaling/v2/twilioconnectiontransport.js
makarandp0: This check ensures that we do not try to reconnect if timer has expired (or if its 1st time connecting)

Timer expiring does not initiate  disconnect, so that ensures it will not affect ongoing attempt.

CHANGELOG.md
ceaglest: We didn't have language in our CL around the server timeout, because we had hardcoded 30 seconds in previous betas. Here is my suggestion:

```suggestion
- twilio-video.js only attempts to reconnect to a Room while the Participant's session is valid (typically 30 seconds after entering the reconnecting state) instead of using a fixed number of retries.
```

lib/util/constants.js
```diff
@@ -85,6 +85,12 @@ module.exports.iceRestartBackoffConfig = {
   randomisationFactor: 0.5
 };
 
+module.exports.reconnectBackoffConfig = {
+  factor: 1.1,
+  initialDelay: 1,
+  randomisationFactor: 0.5
+};
```
ceaglest: Where were the constants derived from? The [FRD](https://code.hq.twilio.com/client/sdk-frd/blob/master/video/rsp-tcmp-wss-client.md#rsp-transport-state-machine) is not very explicit other than stating that an exponential backoff should be used.

makarandp0: I used the same constants that we use for iceConnectionBackoff. We do use exponential backoff, with a factor of 1.1 - so that each attempt is at `lastAttemptDelay * 1.1`

makarandp0: @ceaglest - what factor do you use in CPP sdk ?

ceaglest: * 80 ms for the initial backoff
* 0.5 for randomization factor
* 1.5 backoff multiplier

PR #918, opened 2020-03-20
JSDK-2660 Chrome tab share should be handled similar to screenshare

PikaJoyce:
Added in extra logic for maxbitrate for chrome tabs as well as application windows to be handled similar to screen share.

**Contributing to Twilio**

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -705,7 +705,7 @@ function isChromeScreenShareTrack(track) {
   // NOTE(mpatwardhan): Chrome creates screen share tracks with label like: "screen:69734272*"
   // we will check for label that starts with "screen:D" where D being a digit.
   const isChrome = util.guessBrowser() === 'chrome';
-  return isChrome && track.kind === 'video' && track.label && /^screen:[0-9]+/.test(track.label) && /^web-contents-media-stream:[0-9/]+/.test(track.label) && /^window:[0-9]+/.test(track.label);
+  return isChrome && track.kind === 'video' && track.label && /^screen:[0-9]+/.test(track.label) || /^web-contents-media-stream:[0-9/]+/.test(track.label) || /^window:[0-9]+/.test(track.label);
 }
 
 exports.constants = constants;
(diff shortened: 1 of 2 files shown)


lib/util/index.js
```diff
@@ -705,7 +705,7 @@ function isChromeScreenShareTrack(track) {
   // NOTE(mpatwardhan): Chrome creates screen share tracks with label like: "screen:69734272*"
   // we will check for label that starts with "screen:D" where D being a digit.
   const isChrome = util.guessBrowser() === 'chrome';
-  return isChrome && track.kind === 'video' && track.label && /^screen:[0-9]+/.test(track.label);
+  return isChrome && track.kind === 'video' && track.label && /^screen:[0-9]+/.test(track.label) && /^web-contents-media-stream:[0-9/]+/.test(track.label) && /^window:[0-9]+/.test(track.label);
```
makarandp0: were you able to verify the change? Should that condition between tests be `||` ?

PikaJoyce: @makarandp0 I was trying around with simpler signaling this morning and couldn't get the breakpoints to show up. Also oops, yeah the conditions should be `||`. Changing it now! 

I'll schedule something for the both of us!

test/unit/spec/util/index.js
```diff
@@ -92,4 +94,40 @@ describe('util', () => {
       return promise;
     });
   });
+
+  describe.only('chromeScreenShare', () => {
+    const labels = ['web-contents-media-stream://1174:3', 'window:1561:0', 'screen:2077749241:0'];
```
makarandp0, line 99: So far you are testing for only +ve cases for the function for chrome. Please add some -ve cases as well.

test/unit/spec/util/index.js
```diff
@@ -92,4 +94,40 @@ describe('util', () => {
       return promise;
     });
   });
+
+  describe.only('chromeScreenShare', () => {
```
makarandp0, line 98: Remove .only

PikaJoyce: Done and done

test/unit/spec/util/index.js
PikaJoyce: Added invalidLabels to the mix.

PR #951, opened 2020-04-14
JSDK-2742 start media monitor earlier

makarandp0:
This change updates our media monitor to start when ice connection state goes to `connected` instead of `disconnected`. it will make this behavior consistent with native sdk, and will kick off the ice restart quicker when connection loses.  

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/iceconnectionmonitor.js
+++ b/lib/signaling/v2/iceconnectionmonitor.js
@@ -36,19 +36,11 @@ class IceConnectionMonitor {
         value: null,
         writable: true,
       },
-      _wasInactive: {
-        value: false,
+      _onIceConnectionStateChanged: {
+        value: null,
         writable: true
       }
     });
-  }
-
-  /**
-   * returns true if monitor reported activity when it was last started.
-   * @returns {boolean}
-   */
-  wasInactive() {
-    return this._wasInactive;
   }
 
   _getActivePairStat(stats) {
@@ -56,7 +48,7 @@ class IceConnectionMonitor {
     const hasInBoundTracks = statsArray.find(stat => stat.type === 'inbound-rtp');
     if (!hasInBoundTracks) {
       // NOTE(mpatwardhan): when there are no tracks shared on a peerConnection
-      // on chrome we see inactivity on bytesReceived. but that is not real inactivity,
+      // we may see inactivity on bytesReceived - but that is not real inactivity,
       // ignore it.
       return null;
     }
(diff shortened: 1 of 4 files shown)


lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -65,18 +87,20 @@ class IceConnectionMonitor {
...
           return;
         }
 
+        // console.log('makarand: iceStats.bytesReceived:', iceStats.bytesReceived);
         if (!this._lastActivity || this._lastActivity.bytesReceived !== iceStats.bytesReceived) {
           this._lastActivity = iceStats;
         }
 
         if (iceStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
+          this._wasInactive = true;
```
manjeshbhargav, line 103: @makarandp0 Instead of having a property and a callback here, why don't we just schedule the callback to be called when `iceConnectionState` is `disconnected`? Then, we likely don't have to change anything in PeerConnectionV2:

```js
if (iceStats.timestamp - this._lastActivity.timestamp >= this._inactivityThresholdMs) {
  if (this._peerConnection.iceConnectionState === 'disconnected') {
    onIceConnectionInactive();
  } else {
    this._peerConnection.addEventListener('iceconnectionstatechange', function onIceConnectionStateChanged() {
      if (this._peerConnection.iceConnectionState === 'disconnected') {
        this._peerConnection.removeEventListener('iceconnectionstatechange', onIceConnectionStateChanged);
        onIceConnectionInactive();
      }
    });
  }
}
```

makarandp0: thanks @manjeshbhargav I thought about it as well. Your suggestion does look very elegant, but the reason I did not go for it was:

1. Right now all the changes in response to peerConnection.iceconnectionstatechange are handled in single place, making it bit easier to follow the code. This handling would make it bit harder to follow.

2. when the  `onIceConnectionInactive` would fire on PeerConnection.js - it's possible that its own handler for iceconnectionstatechange is not fired yet. Thus it may have its state  'thinking" iceConnection is 'connected`.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -35,25 +35,47 @@ class IceConnectionMonitor {
       _timer: {
         value: null,
         writable: true,
+      },
+      _wasInactive: {
```
manjeshbhargav, line 39: @makarandp0 Nitpick: can we rename this as `isInactive`? "Was" implies its no longer inactive.

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -35,25 +35,47 @@ class IceConnectionMonitor {
...
+        writable: true
       }
     });
   }
 
+  /**
+   * returns true if monitor reported activity when it was last started.
+   * @returns {boolean}
+   */
+  wasInactive() {
```
manjeshbhargav, line 50:
```suggestion
   * Whether the {@link ICEConnectionMonitor} has detected inactivity.
   * @property {boolean}
   */
  get isInactive() {
```

lib/signaling/v2/iceconnectionmonitor.js
```diff
@@ -94,6 +118,16 @@ class IceConnectionMonitor {
       this._lastActivity = null;
     }
   }
+
+  /**
+   * Stop monitoring the ICE connection state.
+   * and clears _wasInactive state
+   * @returns {void}
+   */
+  clear() {
```
manjeshbhargav, line 127: @makarandp0 Can we clear the inactive flag in `stop()`. Why do we need a separate method?

lib/signaling/v2/peerconnection.js
```diff
@@ -696,20 +710,25 @@ class PeerConnectionV2 extends StateMachine {
...
-          this._initiateIceRestartBackoff();
-          this.emit('iceConnectionStateChanged');
-          this.emit('connectionStateChanged');
+        if (this._lastIceConnectionState === 'disconnected') {
+          this._handleIceConnectionInactivity();
         }
       });
+    } else if ('disconnected' === iceConnectionState) {
+      // did _iceConnectionMonitor previously indicated no-activity?
+      if (this._iceConnectionMonitor.wasInactive()) {
```
manjeshbhargav, line 726:
```suggestion
    } else if (iceConnectionState === 'disconnected' && this._iceConnectionMonitor.isInactive) {
```

lib/signaling/v2/iceconnectionmonitor.js
makarandp0: Hey @manjeshbhargav - It does look much elegant, so I have updated my PR to use your suggestion.

lib/signaling/v2/iceconnectionmonitor.js
makarandp0: not applicable anymore.

lib/signaling/v2/peerconnection.js
```diff
@@ -708,8 +712,9 @@ class PeerConnectionV2 extends StateMachine {
           this.emit('connectionStateChanged');
         }
       });
-    } else {
+    } else if (!['disconnected', 'complete'].includes(iceConnectionState)) { // don't stop monitoring for disconnected or completed.
```
manjeshbhargav, line 715: @makarandp0 I'm not sure other browsers apart from Chrome reach "completed". Can you confirm?

makarandp0: interestingly I only saw `completed` on safari. but your comment found an issue: fixed as:  92f0f57

PR #955, opened 2020-04-16
JSDK-2756 isSupported returns false for Chrome based browsers that are not Electron or Edge.

manjeshbhargav:
@makarandp0 

With this change, `isSupported` should return `false` for all Chrome-based browsers that are not Electron or Edge (ex: Brave, Samsung, Opera).

**Contributing to Twilio**

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,10 +10,8 @@ const {
  * @module twilio-video
  * @property {boolean} isSupported - true if the current browser is officially
  *   supported by twilio-video.js; In this context, "supported" means that
- *   twilio-video.js has been extensively tested with this browser; You are free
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -8,7 +8,14 @@ const {
 
 /**
  * @module twilio-video
- * @property {boolean} isSupported - true if the current browser is officially supported by twilio-video.js.
+ * @property {boolean} isSupported - true if the current browser is officially
+ *   supported by twilio-video.js; In this context, "supported" means that
+ *   twilio-video.js has been extensively tested with this browser; You are free
```
makarandp0, line 13: I would remove the `You are free to use...` line. It's better left implicit.

PR #975, opened 2020-04-25
JSDK-2684 better heartbeat

makarandp0:
update the signaling websocket connection heartbeat logic to:

- Each outgoing message is now treated as an outgoing heartbeat.
- Each incoming message is treated as an incoming heartbeat.
- The incoming heartbeat timeout is now 15 seconds, .
- Removed the offset added to the incoming heartbeat timeout.
- added offset of -200ms  200 to outgoing timeout to account for network latency.
- Add a timeout for getting the onopen event on the WebSocket (15 sec). After that, attempt to reconnect only if this is not the first attempt. 

- [x] TODO: verify that server is okay with not sending heartbeat.
**Contributing to Twilio**


lib/twilioconnection.js
```diff
@@ -30,14 +30,16 @@ const states = {
...
 const DEFAULT_WELCOME_TIMEOUT = 5000;
-const HEARTBEAT_TIMEOUT_OFFSET = 100;
+const OUTGOING_HEARTBEAT_OFFSET = 200;
 const WS_CLOSE_NORMAL = 1000;
 const WS_CLOSE_WELCOME_TIMEOUT = 3000;
 const WS_CLOSE_HEARTBEATS_MISSED = 3001;
 const WS_CLOSE_HELLO_FAILED = 3002;
 const WS_CLOSE_SEND_FAILED = 3003;
 const WS_CLOSE_NETWORK_CHANGED = 3004;
+const WS_CLOSE_FAILED_TO_OPEN = 3005;
```
manjeshbhargav, line 42: @makarandp0 Nitpick:
```suggestion
const WS_CLOSE_OPEN_TIMEOUT = 3005;
```

lib/twilioconnection.js
```diff
@@ -205,7 +206,7 @@ class TwilioConnection extends StateMachine {
           // Do nothing.
           break;
         case 'heartbeat':
-          this._handleHeartbeat();
+          // Do nothing.
```
manjeshbhargav, line 209: @makarandp0 You can remove this case altogether I think.

lib/twilioconnection.js
```diff
@@ -303,11 +305,18 @@ class TwilioConnection extends StateMachine {
     if (this.state !== 'connecting') {
       return;
     }
-    const heartbeatTimeout = negotiatedTimeout + HEARTBEAT_TIMEOUT_OFFSET;
+
+    const { maxConsecutiveMissedHeartbeats } = this._options;
+    // we will essentially wait for (maxConsecutiveMissedHeartbeats) missed heartbeats before closing the connection.
+    const heartbeatTimeout = (negotiatedTimeout * maxConsecutiveMissedHeartbeats);
```
manjeshbhargav, line 311:
```suggestion

    // NOTE(mpatwardhan): Wait for 3 missed heartbeats before closing the TwilioConnection.
    const heartbeatTimeout = negotiatedTimeout * maxConsecutiveMissedHeartbeats;
```

lib/twilioconnection.js
```diff
@@ -303,11 +305,18 @@ class TwilioConnection extends StateMachine {
...
+
+    // Outgoing heartbeat should be sent 200 ms earlier than the specified interval to account for network latency.
+    const outgoingHeartbeatTimeout = negotiatedTimeout - OUTGOING_HEARTBEAT_OFFSET;
```
manjeshbhargav, line 314:
```suggestion
    // NOTE(mpatwardhan): Outgoing heartbeat is sent 200 ms earlier to account for network latency.
    const outgoingHeartbeatTimeout = negotiatedTimeout - OUTGOING_HEARTBEAT_OFFSET;
```

lib/twilioconnection.js
makarandp0: I think keeping this case with `// do nothing` comment helps with understaning possible `message.type`  values.

lib/twilioconnection.js
```diff
@@ -435,7 +445,7 @@ class TwilioConnection extends StateMachine {
  * {@link TwilioConnection} options
  * @typedef {object} TwilioConnectionOptions
  * @property {LogLevel} [logLevel=warn] - Log level of the {@link TwilioConnection}
- * @property {number} [maxConsecutiveMissedHeartbeats=5] - Max. number of consecutive "heartbeat" messages that can be missed
+ * @property {number} [maxConsecutiveMissedHeartbeats=3] - Max. number of (effective) consecutive "heartbeat" messages that can be missed
```
ceaglest, line 448: Is the 5 just a documentation a typo, while DEFAULT_MAX_CONSECUTIVE_MISSED_HEARTBEATS represents the correct value of 3?

lib/twilioconnection.js
```diff
@@ -30,14 +30,16 @@ const states = {
 
 const DEFAULT_MAX_CONSECUTIVE_MISSED_HEARTBEATS = 3;
 const DEFAULT_MAX_REQUESTED_HEARTBEAT_TIMEOUT = 5000;
+const DEFAULT_OPEN_TIMEOUT = 15000;
 const DEFAULT_WELCOME_TIMEOUT = 5000;
```
ceaglest, line 34: Welcomes should be allowed for up to 15 seconds, and the hello should be sent up to 3 times at [0, 5, 10] seconds. It seems like this change should be part of this PR rather than https://github.com/twilio/twilio-video.js/pull/985 (JSDK-2777).

lib/twilioconnection.js
```diff
@@ -197,6 +195,9 @@ class TwilioConnection extends StateMachine {
         this.emit('error', error);
         return;
       }
+
+      // Each incoming message should be treated as an incoming heartbeat
+      this._handleHeartbeat();
```
ceaglest, line 200: From our discussion on slack I don't think you should treat each sent or received message as a heartbeat. I was planning on rolling back this change in C++ since it is out of sync with the Server and we don't have agreement between teams. The concern I have is that treating every message as a heartbeat allows the connection to become further and further behind.

ceaglest:
FWIW, it looks like treating every message as a heartbeat was attempted in the Java implementation but the [PR](https://code.hq.twilio.com/twilio/twilio-connections/pull/63) was not merged.

lib/twilioconnection.js
makarandp0: @ceaglest I understand that we shouldn't treat outgoing message as heartbeat - because server expects heartbeats. But don't see why not for incoming messages. If we miss heartbeat but received incoming message why should we disconnect ?
Can you elaborate more on 
>  treating every message as a heartbeat allows the connection to become further and further behind.

makarandp0: resolved after the offline discussion - we decided to treat every incoming message as incoming heartbeat, and outgoing message as outgoing heartbeat. since server already [does the same](https://code.hq.twilio.com/twilio/twilio-connections/blob/master/connections-core/src/main/java/com/twilio/connections/actor/ServerConnectionActor.java#L559) -

lib/twilioconnection.js
makarandp0: yes, seems like just documentation typo.

lib/twilioconnection.js
makarandp0: Looks like #985 already covered that.

lib/twilioconnection.js
makarandp0: resolved after offline discussion - we decided to treat incoming message as heartbeat.

lib/twilioconnection.js
makarandp0: looks like #985 already handled this requirement.

makarandp0:
@ceaglest and @manjeshbhargav - Thank you for the discussion on this earlier.  I have re-introduce "treat outgoing message as heartbeat" change. Please take another look. I am planning to verify that server indeed is okay with us treating outgoing message as heartbeat before merging this.

lib/twilioconnection.js
```diff
@@ -267,6 +277,9 @@ class TwilioConnection extends StateMachine {
...
```
ceaglest, line 282: Calling `_handleHeartbeat` for every `message.type` might result in unexpected logs from the SDK. Move this call into `heartbeat` and `msg` instead?

makarandp0: Good point - I have updated now to call this only for msg and heartbeat cases.

makarandp0:
hey @ceaglest, I have addressed your feedback, can you please take another look.

makarandp0:
I verified that
-  if client does not send any heartbeats but send messages (track enable/disable) at regular interval. server does not disconnect.
- if client does not send any heartbeats and not messages as well. server does disconnect the connection after few missed heartbeats.

This verifies our assumption that treating messages as heartbeat is okay with server. I will merge this now.

makarandp0:
update the signaling websocket connection heartbeat logic to:

- Each outgoing message is now treated as an outgoing heartbeat.
- Each incoming message is treated as an incoming heartbeat.
- The incoming heartbeat timeout is now 15 seconds, .
- Removed the offset added to the incoming heartbeat timeout.
- added offset of -200ms  200 to outgoing timeout to account for network latency.
- Add a timeout for getting the onopen event on the WebSocket (15 sec). After that, attempt to reconnect only if this is not the first attempt.

PR #985, opened 2020-04-30
JSDK-2777 TCMP busy-wait retry mechanism.

manjeshbhargav:
@makarandp0 @ceaglest 

This PR covers the following changes:

* `TwilioConnection` transitions to a new "wait" state when VSG responds with a "busy" message when "hello" is sent.
* `TwilioConnection` implements the initial "early" state which is part of the TCMP spec.
* `TwilioConnection` transitions from "wait" to "early" after a delay of `retryAfter` ms if `keepAlive` is `false`.
* `TwilioConnection` transitions from "wait" to "connecting" after a delay of `retryAfter` ms if `keepAlive` is `true`.
* `TwilioConnection` saves the `cookie` from the "busy" message and passes it in subsequent "hello" messages, so that different VSG hosts behind an ALB can track retry attempts.
* `connect` fails if `retryAfter` is `-1`.
* Send "connect" RSP message in the "hello" body.
* Emit `EventListener` events for `TwilioConnection` state transitions.

### TODO

- [x] New TwilioError for `connect` failure due to `retryAfter = -1`.
- [x] Document the different `EventListener` events.

### Resources

* [EventListener IDL](https://code.hq.twilio.com/client/video-sdk-api/pull/64/files)
* [EventListener events](https://code.hq.twilio.com/client/rtc-insights-spec/pull/54/files)

**Contributing to Twilio**


lib/twilioconnection.js
```diff
@@ -201,6 +252,9 @@ class TwilioConnection extends StateMachine {
         case 'bad':
           this._handleBad(message);
           break;
+        case 'busy':
```
makarandp0, line 255: with the introduction of new states, we should add a check like this before handling messages. 

```
const legalMessages = {
  closed: [],
  connecting: ['bad', 'busy', 'welcome'], 
  early:  [], 
  open:  ['bad', 'bye', 'heartbeat', 'msg'],
  wait: [],
}

if (legalMessages[this.state].indexOf(message.type) < 0) {
  this._log.error(`Illegal message: ${message.type} when in state:${this.state}`);
  return;
}
```

lib/twilioconnection.js
```diff
@@ -222,9 +276,7 @@ class TwilioConnection extends StateMachine {
 
     ws.addEventListener('open', () => {
```
makarandp0, line 277: nit: we should move this before the message handling code

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -485,7 +497,6 @@ function setupEventListeners(transport) {
     const twilioConnection = createOrResetTwilioConnection();
     twilioConnection.once('close', disconnect);
     twilioConnection.on('message', handleMessage);
-    twilioConnection.once('open', connect);
```
makarandp0, line 488: since we do not use open event anymore, how about stopping twilio connection from emitting it? - I am not very sure about this though - given that we have this logger approroach to events now.

makarandp0, line 488: but I suggest we move  following lines to `createOrResetTwilioConnection` so that a) we can also unhook `close` from old connection. and b) It would look more obvious if hooking and unhooking was in the same function.  
```
    twilioConnection.once('close', disconnect);
    twilioConnection.on('message', handleMessage);
```

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -442,17 +454,17 @@ function reduceUpdates(updates) {
...
     if (transport._twilioConnection) {
       transport._twilioConnection.removeListener('message', handleMessage);
     }
     const { _options, _wsServer } = transport;
-    const { TwilioConnection } = transport._options;
-    transport._twilioConnection = new TwilioConnection(_wsServer, _options);
+    const { TwilioConnection } = _options;
+
+    transport._twilioConnection = new TwilioConnection(_wsServer, Object.assign({
+      helloBody: transport._createConnectOrSyncOrDisconnectMessage()
```
makarandp0, line 465: With the introduction of wait state, it could take long time for transport to actually send helloBody. I am wondering that helloBody should be obtained from the callback when transport is ready to send it. That way we can pass most recent offer/peer connection information.

lib/util/eventobserver.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const { EventEmitter } = require('events');
```
makarandp0, line 3: think we should expand this into a logger, it looks so much like it.

ceaglest:
I have posted a PR to cover the new error code:

https://code.hq.twilio.com/client/twilio-video-errors/pull/46

lib/twilioconnection.js
```diff
@@ -12,22 +12,39 @@ let nInstances = 0;
...
 
-  +--------------+       +----------+
-  |  connecting  | ----> |  closed  |
-  +--------------+       +----------+
-         |                    ^
-         v                    |
-     +--------+               |
-     |  open  | ---------------
-     +--------+
+       ------------------------------------------
```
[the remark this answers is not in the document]
ceaglest: @manjeshbhargav @makarandp0 how do you feel about the test coverage added?

The states and allowed transitions are what I expected. 👍

lib/twilioconnection.js
```diff
@@ -165,27 +212,31 @@ class TwilioConnection extends StateMachine {
...
-    this._ws = new this._options.WebSocket(serverUrl);
-    const log = this._log;
-    const ws = this._ws;
-
+  _connect() {
+    if (this.state === 'wait') {
+      this.transition('early');
+    } else if (this.state !== 'early') {
+      return;
+    }
```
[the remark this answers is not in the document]
ceaglest: Should `else if (this.state !== 'early')` be treated as an exceptional case? It seems like programmer error is the only reason why the early return should occur.

Edit: I agree with @makarandp0 assuming assert() fits into your coding style.

lib/twilioconnection.js
ceaglest: Thanks for making this comment, because I think it is worthwhile to consider which messages are allowed. Based upon what I have in C++, here is what I would use:

```
const legalMessages = {
  // All messages are allowed but ignored...
  closed: [],
  // msg is deferred in case it arrives out of order. Mostly a concern for transports that do not enforce ordering. Since WS does you could omit it
  connecting: ['bad', 'busy', 'welcome', 'msg'],
  // C++ defers welcome and msg but this seems unnecessary to me 
  early:  [], 
  open:  ['bad', 'bye', 'heartbeat', 'msg'],
  // I'd like to remove heartbeat but there was a server bug.
  // busy is possible due to the client sending up to 3x hellos and server responding to each hello
  wait: ['busy', 'heartbeat'],
}
```

lib/twilioconnection.js
```diff
@@ -247,6 +299,41 @@ class TwilioConnection extends StateMachine {
     this.emit('error', new Error(reason));
   }
 
+  /**
+   * Handle an incoming "busy" message.
+   * @param {{cookie: string, keepAlive: boolean, retryAfter: number}} message
+   * @private
+   */
+  _handleBusy({ cookie, keepAlive, retryAfter }) {
+    if (this.state !== 'connecting') {
```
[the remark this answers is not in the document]
ceaglest: busy can arrive in the wait state due to hello / busy crossing so you need to allow busy messages in "wait" as well.

lib/twilioconnection.js
```diff
@@ -247,6 +299,41 @@ class TwilioConnection extends StateMachine {
...
+      return;
+    }
+    const reason = retryAfter < 0
```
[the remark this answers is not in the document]
ceaglest: We should make sure to clarify this with @ykhodosh. My understanding is retryAfter is required and is always a number. I like treating all negative values as the terminal case, though I do wonder how 0 should be handled (is that retry instantly, or also a failure?)

lib/twilioconnection.js
```diff
@@ -247,6 +299,41 @@ class TwilioConnection extends StateMachine {
...
+      ? 'Received terminal "busy" message'
+      : `Received "busy" message, retrying after ${retryAfter} ms`;
+
+    if (retryAfter < 0) {
+      this._handleBad({ reason });
+      return;
+    }
+    // TODO(mmalavalli): Remove check for the existence of "cookie" once VSG supports it.
```
ceaglest, line 319: I added a comment to the doc, and @ykhodosh agreed that the cookie should remain optional in busy.

lib/twilioconnection.js
```diff
@@ -386,6 +482,17 @@ class TwilioConnection extends StateMachine {
...
 
+  /**
+   * Start the TCMP handshake.
+   * @private
+   */
+  _startHandshake() {
+    this.transition('connecting');
+    this._sendHello();
+    const { welcomeTimeout } = this._options;
+    this._welcomeTimeout = new Timeout(() => this._handleWelcomeTimeout(), welcomeTimeout);
```
ceaglest, line 493: This may be covered in @makarandp0's other PR, but the welcome timeout should be 15 seconds not 5 seconds, and the welcome should be sent at 0, 5, and 10 seconds during this time.

test/unit/spec/signaling/v2/twilioconnectiontransport.js
```diff
@@ -1023,19 +1018,6 @@ describe('TwilioConnectionTransport', () => {
...
-          assert.equal(message.name, test.name);
-          assert.equal(message.participant, test.localParticipantState);
-          assert.deepEqual(message.peer_connections, test.peerConnectionManager.getStates());
-          assert.equal(message.token, test.accessToken);
-          assert.equal(message.type, 'connect');
-          assert.equal(message.version, 2);
-          assert.equal(message.publisher.name, `${name}.js`);
-          assert.equal(message.publisher.sdk_version, version);
-          assert.equal(typeof message.publisher.user_agent, 'string');
-        });
```
ceaglest, line 1038: It may still be worth it to test sending an update sent while connecting after the hello body.

test/unit/spec/twilioconnection.js
```diff
@@ -173,185 +234,294 @@ describe('TwilioConnection', function() {
...
+                          assert.equal(hello.timeout, twilioConnection._options.requestedHeartbeatTimeout);
+                          assert.equal(hello.type, 'hello');
+                          assert.equal(hello.cookie, 'foo');
+                          if (options.helloBody) {
+                            assert.equal(hello.body, options.helloBody);
+                          } else {
+                            assert.equal('body' in hello, false);
+                          }
+                        });
+                      });
```
ceaglest, line 494: Is it possible to test that retryAfter is respected rather than a hello being eventually sent?

Note: The cookie should be optional, if the server busy does not include one then the client hello should not either.

test/unit/spec/twilioconnection.js
```diff
@@ -173,185 +234,294 @@ describe('TwilioConnection', function() {
...
-
-        it('should call .close on the underlying WebSocket', () => {
-          sinon.assert.callCount(twilioConnection._ws.close, 1);
-        });
-
-        it('should emit "close" on the TwilioConnection with the appropriate Error', () => {
-          assert(error instanceof Error);
-          assert(/^WebSocket Error 3000/.test(error.message));
-        });
       });
```
ceaglest, line 525: It looks like many of the individual cases are being tested, but it may be worthwhile to try the full busy call flow including more than one loop through early -> connecting -> wait -> ...

lib/twilioconnection.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 According to the new proposal, `retryAfter` will always be a number.

lib/twilioconnection.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Please check out `test/unit/spec/twilioconnection.js` to look at the new tests added.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 I don't think that is necessary, because there is no way to change which tracks are published until the Room object is available to the developer. So, there is no way for the tracks in offer/answer to change until the Room is connected.

lib/twilioconnection.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I don't think it should be considered terminal to receive a "busy" after the "hello" <--> "welcome" handshake is complete.

lib/twilioconnection.js
```diff
@@ -12,23 +12,40 @@ let nInstances = 0;
...
+const events = {
+  closed: 'close',
+  open: 'open',
+  wait: 'wait'
+};
+
+const TCMP_VERSION = 2;
+
 const DEFAULT_MAX_CONSECUTIVE_MISSED_HEARTBEATS = 3;
+const DEFAULT_MAX_CONSECUTIVE_FAILED_HELLOS = 3;
```
ceaglest, line 48: Nitpick: I'm not sure that this value needs to be configurable. The idea is that hellos are just heartbeats used to establish the initial connection so you could reuse `DEFAULT_MAX_CONSECUTIVE_MISSED_HEARTBEATS`.

lib/twilioconnection.js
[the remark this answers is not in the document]
ceaglest: @manjeshbhargav WDYT?

lib/twilioconnection.js
ceaglest: @manjeshbhargav WDYT?

lib/twilioconnection.js
```diff
@@ -201,6 +268,9 @@ class TwilioConnection extends StateMachine {
         case 'bad':
           this._handleBad(message);
           break;
+        case 'busy':
+          this._handleBusy(message);
+          break;
         case 'bye':
           // Do nothing.
           break;
```
ceaglest, line 276: I am surprised that nothing is done in response to a **bye**. The appropriate response would be to close the WebSocket transport. The server should be closing their side but it would be nice to do the same on the client just to be safe.

lib/twilioconnection.js
manjeshbhargav: @ceaglest You're right. We don't override the default behavior when using TwilioConnection in the SDK. However, we use this for unit tests.

lib/twilioconnection.js
manjeshbhargav: @ceaglest BYE is usually followed by VSG closing the WebSocket with a `1006` error code. I'm not sure if we would want to change the behavior here. What do you do in the C++ SDK?

lib/twilioconnection.js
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest You're right. `_connect()` should not be called unless the state is either `wait` or `early`. This is just defensive programming. We don't use asserts typically, but I will add a warning or error log here. Does that sound good?

lib/twilioconnection.js
manjeshbhargav: @ceaglest Sure, I'll add the log messages.

lib/twilioconnection.js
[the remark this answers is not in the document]
ceaglest: Yes, I think a warning or error log would be fine if you don't use asserts.

lib/twilioconnection.js
ceaglest: Receiving a bye from the server triggers the client sending a bye and then closing the WebSocket (we only use a 1000 error for now). Since the server's WebSocket close frame usually arrives in the same TCP / TLS packet as the bye the close is handled automatically for us most of the time.

This shouldn't be special case logic though. Is it possible to just have the bye cause a transition to the closed state? This feedback is probably out of scope for the PR at hand.

lib/twilioconnection.js
manjeshbhargav: @ceaglest @makarandp0 I'm adding these warning messages within their respective `_handle` methods, since I already check for invalid states there.

lib/twilioconnection.js
manjeshbhargav: @ceaglest Right now, we depend on the server-side error code to trigger a reconnect. So let's work towards normalizing this behavior in a subsequent PR.

test/unit/spec/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @ceaglest We [enqueue](https://github.com/twilio/twilio-video.js/blob/master/lib/signaling/v2/twilioconnectiontransport.js#L291) all `update` messages until we receive "connected" message from Room Service.

test/unit/spec/twilioconnection.js
manjeshbhargav: @ceaglest We do have a test for `early -> connecting -> busy -> connecting -> open` as well as `early -> connecting -> busy -> early -> connecting -> open`.

lib/twilioconnection.js
```diff
@@ -392,6 +404,8 @@ class TwilioConnection extends StateMachine {
    */
   _handleMessage({ body }) {
     if (this.state !== 'open') {
+      this._log.warn(`Unexpected state "${this.state}" for handling a "msg" message`
+        + ' from the TCMP server.');
```
ceaglest, line 408: You could defer messages while in the connecting state just in case they somehow arrive before the welcome. It shouldn't happen with a WebSocket.

lib/twilioconnection.js
```diff
@@ -309,11 +319,14 @@ class TwilioConnection extends StateMachine {
 
   /**
    * Handle an incoming "busy" message.
-   * @param {{cookie: string, keepAlive: boolean, retryAfter: number}} message
+   * @param {{cookie: ?string, keepAlive: boolean, retryAfter: number}} message
```
ceaglest, line 322: Optional is good. 👍

test/unit/spec/twilioconnection.js
```diff
@@ -40,7 +41,7 @@ class FakeWebSocket extends EventTarget {
   FakeWebSocket[readyState] = i;
 });
 
-describe('TwilioConnection', function() {
+describe.only('TwilioConnection', function() {
```
ceaglest, line 44: Revert.

CHANGELOG.md
```diff
@@ -1,5 +1,39 @@
...
+  });
+
+  connect('token', { eventListener: sdkEvents }).then(room => {
+    console.log('Successfully connected to Room:', room.name);
+  }, error => {
+    if (error.code === 53006) {
+      console.error('Twilio\'s signaling server cannot accept connection requests at this time.');
+    }
+  });
+  ```
```
ceaglest, line 35: This works for me. It might be worth mentioning the hello.body() enhancement, but it could be added later.

manjeshbhargav: @ceaglest I'm not sure how perceptive the perceived reduction in connect times will be. Can you link me to the CHANGELOG entry you wrote for C++?

lib/twilioconnection.js
manjeshbhargav: @ceaglest This has never happened so far, so I'll defer on this for now.

PR #987, opened 2020-05-01
JSDK-2577 setNetworkQualityConfiguration fixes

makarandp0:
`setNetworkQualityConfiguration` did update the configuration, but did not fire the `networkQualityLevelChanged` if there were no network level changes.

This change makes some small changes to setNetworkQualityConfiguration handling.
- now it fires an event if the stats data has added/removed fields (this ensures that `networkQualityLevelChanged` would fire when user updates the configuration, and server sends new stats for the configuration.

- fixed instability in existing tests for the failures I observed during local run.
-- they were listening for  `networkQualityLevelChanged`, but were checking the network level async. network level could change between the event getting fired and the check.  
-- the tests were expecting the remote network stats even when remote network level was 0. I noticed that we do not get any stats when remote network level is 0.

- added new tests to 
-- validate setNetworkQualityConfiguration
-- validate `networkQuality=false`
 

**Contributing to Twilio**


makarandp0:
@manjeshbhargav @ceaglest Please take a look when you get a chance. It changes the behavior of the `networkQualityLevelChanged` slightly.

ceaglest:
> I think we would be changing the behavior of the networkQualityLevelChanged event here. According to the Participant IDL, the event is raised only when the score changes, irrespective of the verbosity level.

If the event is only fired when the level changes then how will the developer know when to retrieve updated NetworkQualityStats? It seems like either a new event is needed or the behavior of the existing event needs to be modified. I think the existing event should be fired if:

1. The NetworkQualityLevel changes.
2. An updated NetworkQualityStats object is available.

Or we need to introduce a new event..

You might want to ask @paynerc about the behavior as he implemented support for mobile platforms.

makarandp0:
Thanks @ceaglest! 

@paynerc, can you please comment on this behavior wrt mobile.

makarandp0:
From offline response from @paynerc - Looks like this issue isnt applicatble to mobie sdk since they do not have additional starts reporting.
> We did not add the ability to change NetworkQualityLevelConfigurations from the LocalParticipant. Since we don’t support the detailed level of stats reporting, we didn’t add the complexity to change it after the initial configuration was provided as the ConnectOption.

makarandp0:
Thanks you @manjeshbhargav and @ceaglest for your feedback and discussion on this. I have removed the behavior changes now, and just updated the test stability. 

I also filed a low priority feature request to add support for `networkQualityStatsChange` event (https://issues.corp.twilio.com/browse/JSDK-2827)

Please take another look.

PR #995, opened 2020-05-05
Check whether environment is a browser before other functions reference navigator

sedenardi:
**Contributing to Twilio**

The PR https://github.com/twilio/twilio-video.js/pull/955/files changed the order in which `isSupported()` evaluates different browser features. By moving `!!browser` from the beginning, it attempts to access `navigator` and other browser globals before even checking whether it's in a browser environment. This breaks using the library with NodeJS server-side rendering. This fix first checks to see whether it's running in a browser before attempting to use any of the browser globals.

Let me know if you have any questions/comments/concerns. Thanks.

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -87,9 +87,9 @@ function rebrandedChromeBrowser(browser) {
 function isSupported() {
   const browser = guessBrowser();
   const rebrandedChrome = rebrandedChromeBrowser(browser);
-  return isGetUserMediaSupported()
+  return !!browser
+    && isGetUserMediaSupported()
     && isRTCPeerConnectionSupported()
-    && !!browser
     && (!rebrandedChrome || SUPPORTED_CHROME_BASED_BROWSERS.includes(rebrandedChrome))
     && !isNonChromiumEdge(browser);
 }


manjeshbhargav:
Hi @sedenardi ,

Thanks for the PR. We already have a bug filed for this and will work on it in the coming sprint. I'm curious however, what is the reason for using `isSupported` server side? Even after this issue is fixed, it will still return `false` on Node.js. This flag is more useful if it is run on the browser to know whether twilio-video.js explicitly supports the browser.

Thanks,

Manjesh Malavalli
JSDK Team

sedenardi:
No problem @manjeshbhargav.

I'm not explicitly using `isSupported` or even `twilio-video` on the server, but rather I'm using it in a React app that uses server-side rendering. So even though there's no call to `Twilio.createLocalTracks` or `Twilio.connect` or anything that requires a browser, simply loading components on the server which reference this library experience an issue.

I believe this is because `isSupported` is [immediately evaluated](https://github.com/twilio/twilio-video.js/blob/master/lib/index.js#L41), rather than when you begin to use the library.

sdenardi:
Looks like this was fixed by https://github.com/twilio/twilio-video.js/pull/1013/files and released in `2.5.0`. Feel free to close this PR, thanks @manjeshbhargav.

PR #999, opened 2020-05-06
JSDK-2780 missing localtrack stats

makarandp0:
This change consumes: https://github.com/twilio/twilio-webrtc.js/pull/119 to get correct values for packetsLost,   roundTripTime for local tracks

**Contributing to Twilio**


lib/stats/localtrackstats.js
```diff
@@ -21,19 +21,19 @@ class LocalTrackStats extends TrackStats {
       bytesSent: {
         value: typeof statsReport.bytesSent === 'number'
           ? statsReport.bytesSent
-          : null,
+          : 0,
```
manjeshbhargav, line 24: @makarandp0 I think we should retain `null` here (since this is public-facing) and convert them to 0 when we are actually publishing it.

lib/stats/localtrackstats.js
```diff
@@ -21,19 +21,19 @@ class LocalTrackStats extends TrackStats {
...
         enumerable: true
       },
       packetsSent: {
         value: typeof statsReport.packetsSent === 'number'
           ? statsReport.packetsSent
-          : null,
+          : 0,
```
manjeshbhargav, line 30: @makarandp0 Same as above.

lib/stats/localtrackstats.js
```diff
@@ -21,19 +21,19 @@ class LocalTrackStats extends TrackStats {
...
         enumerable: true
       },
       roundTripTime: {
         value: typeof statsReport.roundTripTime === 'number'
           ? statsReport.roundTripTime
-          : null,
+          : 0,
```
manjeshbhargav, line 36: @makarandp0 Same as above.

makarandp0:
Hey, @manjeshbhargav, I have addressed your feedabck, can you take another look?

CHANGELOG.md
```diff
@@ -37,6 +40,11 @@ For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/suppor
 - Reduced connection times by removing a round trip during the initial handshake with Twilio's
   signaling server. (JSDK-2777)
 
+Bug Fixes
+---------
+
+- Fixed a bug in `Room.getStats()` where it did not return correct values for packetsLost, roundTripTime for local tracks. (JSDK-2780, JSDK-2755)
```
manjeshbhargav, line 46:
```suggestion
- Fixed a bug in `Room.getStats()` where it did not return correct values for `packetsLost`, `roundTripTime` for LocalTracks. (JSDK-2780, JSDK-2755)
```

lib/signaling/v2/room.js
```diff
@@ -765,8 +765,8 @@ function periodicallyPublishStats(roomV2, transport, intervalMs) {
       stats.forEach((response, id) => {
         // NOTE(mmalavalli): A StatsReport is used to publish a "stats-report"
         // event instead of using StandardizedStatsResponse directly because
-        // StatsReport will add nulls to properties that do not exist.
-        const report = new StatsReport(id, response);
+        // StatsReport will add zeros to properties that do not exist.
+        const report = new StatsReport(id, response, true /* prepareForInsights */);
```
manjeshbhargav, line 769: @makarandp0 I was thinking we could leverage [replaceNullWithDefaults](https://github.com/twilio/twilio-video.js/blob/master/lib/signaling/v2/room.js#L825) in RoomV2? Then we won't need to change `StatsReport` which is a public facing class.

lib/stats/localaudiotrackstats.js
```diff
@@ -12,9 +12,10 @@ class LocalAudioTrackStats extends LocalTrackStats {
   /**
    * @param {string} trackId - {@link LocalAudioTrack} ID
    * @param {StandardizedTrackStatsReport} statsReport
+   * @param {boolean} prepareForInsights
    */
-  constructor(trackId, statsReport) {
-    super(trackId, statsReport);
+  constructor(trackId, statsReport, prepareForInsights) {
+    super(trackId, statsReport, prepareForInsights);
```
manjeshbhargav, line 18: @makarandp0 Please see my comment above.

lib/stats/localtrackstats.js
```diff
@@ -13,27 +13,28 @@ class LocalTrackStats extends TrackStats {
   /**
    * @param {string} trackId - {@link LocalTrack} ID
    * @param {StandardizedTrackStatsReport} statsReport
+   * @param {boolean} prepareForInsights
    */
-  constructor(trackId, statsReport) {
+  constructor(trackId, statsReport, prepareForInsights) {
```
manjeshbhargav, line 18: @makarandp0 Please see my comment above.

makarandp0:
@manjeshbhargav - I looked into reusing something like `replaceNullWithDefaults` but things get more complicated with it.

This is mainly because `replaceNullsWithDefaults` operates on response from getStats, and does not do any property renaming. 

StatsReport does renaming and folding of properties. For example: 
`response.frameRateInput` becomes `report.captureFrameRate`

I think it's better to let  statsReport manage defaults instead of overwriting them outside it.

lib/signaling/v2/room.js
makarandp0: @manjeshbhargav - I looked into reusing something like `replaceNullWithDefaults` but things get more complicated with it. 

I think  `replaceNullsWithDefaults` was necessary for `activeIceCandidatePair` because it operates on response from getStats, and does not do any property renaming. Where as for stats-report we have a good implementation of StatsReport does that renaming and folding of properties. For example: 
`response.frameRateInput` becomes `report.captureFrameRate`

I think it's better to let  statsReport manage defaults instead of overwriting them outside it.

lib/stats/localaudiotrackstats.js
makarandp0: same as above.

PR #1000, opened 2020-05-06
JSDK-2769 Renegotiate when a Track is added during a current negotiation.

manjeshbhargav:
@makarandp0 

The reason sometimes calling `unpublishTrack` / `publishTrack` multiple times in quick succession resulted in publication not being complete is because a new negotiation was not being triggered after adding the track to the PeerConnection. This was happening because of the following sequence:
```
PeerConnectionV2.offer() -- remove track --> _needsAnswer = true
createOffer1 = PeerConnection.createOffer()
PeerConnectionV2.offer() -- add track --> _shouldOffer = true, before createOffer1 is resolved
await createOffer1 --> _shouldOffer = false
```
Because of this sequence of events, there is no negotiation initiated for the second `offer()` when a new track is added, and therefore the publication is never completed.

This fix makes sure that after the current negotiation is complete, we renegotiate if we notice that the number of send* m= lines in the most recent local offer is less than the number of non-null RTCRtpSenders in the PeerConnection. This now triggers renegotiation for the second `offer()` in the above sequence, thereby completing the publication.

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -858,20 +858,11 @@ class PeerConnectionV2 extends StateMachine {
     let shouldReoffer = this._shouldOffer;
 
     if (localDescription && localDescription.sdp) {
-      // NOTE(mmalavalli): For "unified-plan" sdps, if the remote RTCPeerConnection sends
-      // an offer with fewer audio m= lines than the number of audio RTCRTPSenders
-      // in the local RTCPeerConnection, then the local RTCPeerConnection creates
-      // an answer with the same number of audio m= lines as in the offer. This
-      // behavior was triggered by the removal of 'offerToReceiveAudio' from the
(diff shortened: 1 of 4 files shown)


makarandp0:
Great analysis and fix @manjeshbhargav! The fix looks good to me. 
However I would like to see an update to test - The tests even after your changes does not reproduce the issue consistently, and I suspect that's because, both participants are joining simultaneously, and that makes it sometime work depending on timing. 

What do you think about writing a new tests where alice joins the room first, and Bob joins after alice is fully in the room and publishes two tracks rapidly, it would reproduce the issue consistently? WDYT?

manjeshbhargav:
@makarandp0 The original issue is also not consistently reproducible out in the field. This test mirrors the steps that the customers are going through when they are seeing the issue. So this test should be sufficient.

PR #1013, opened 2020-05-11
JSDK-2676 In-band TURN implementation

manjeshbhargav:
@makarandp0 

This PR pulls in your in-band TURN changes, with a few modifications to get it to work with the recent BUSY/WAIT changes to TwilioConnection and including "iced" and "connect" RSP changes in the body of HELLO.

This PR also fixes JSDK-2818. ([Commit](https://github.com/twilio/twilio-video.js/pull/1013/commits/970b65a53aa73240cd066bfc4de8ac6995dfa363))

**Contributing to Twilio**


lib/connect.js
```diff
@@ -336,12 +317,6 @@ function connect(token, options) {
  * You may pass these options to {@link connect} in order to override the
  * default behavior.
  * @typedef {object} ConnectOptions
- * @property {boolean} [abortOnIceServersTimeout=false] - If fetching ICE
```
manjeshbhargav, line 339: TODO: Add CHANGELOG entry for the deprecation of this property.

lib/connect.js
```diff
@@ -376,8 +351,6 @@ function connect(token, options) {
  *   might be useful for your own reporting and diagnostics.
  * @property {Array<RTCIceServer>} iceServers - Override the STUN and TURN
  *   servers used when connecting to {@link Room}s
- * @property {number} [iceServersTimeout=3000] - Override the amount of time, in
```
manjeshbhargav, line 379: TODO: Add CHANGELOG entry for the deprecation of this property.

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,133 @@ const {
...
-      // NOTE(mmalavalli): PeerConnectionManager#createAndOffer() queues the
-      // initial offer in the event queue for the 'description' event. So,
-      // we are dequeueing to prevent the spurious 'update' message sent by
-      // the client after connecting to a room.
-      peerConnectionManager.dequeue('description');
+  let transport;
 
-      return new Promise((resolve, reject) => {
+  return new CancelablePromise((resolve, reject, isCanceled) => {
+    return new Promise((resolve, reject) => {
```
makarandp0, line 32: I remember you making a suggestion of resolving icePromise locally, instead of having transport always send onIced on my PR - which I think made this code much simpler. Why not take the same route? 
```
const ensureIceServers = iceServers ? Promise.resolve(iceServers) : waitForEvent(transport, 'iced');
```

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,133 @@ const {
...
+        wsServerInsights
+      } : {}, InsightsPublisher ? {
+        InsightsPublisher
+      } : {}, NullInsightsPublisher ? {
+        NullInsightsPublisher
+      } : {}, bandwidthProfile ? {
+        bandwidthProfile
+      } : {});
+
+      transport = new Transport(
```
makarandp0, line 95: I think we should move creation of transport earlier (beofre onIce is declared) - that would make this flow look better

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -463,11 +480,13 @@ function setupEventListeners(transport) {
...
       transport._twilioConnection.removeListener('message', handleMessage);
     }
-    const { _options, _wsServer } = transport;
+    const { _iceServersStatus, _options, _wsServer, state } = transport;
     const { TwilioConnection } = _options;
 
     const twilioConnection = new TwilioConnection(_wsServer, Object.assign({
-      helloBody: transport._createConnectOrSyncOrDisconnectMessage()
+      helloBody: state === 'connecting' && _iceServersStatus === 'acquire'
+        ? transport._createIceMessage()
```
makarandp0, line 488: we want to get ice servers just once, and not for every reconnection right? Wouldnt this cause ice message to be sent for every reconnection ?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -594,7 +618,14 @@ function setupEventListeners(transport) {
     }
   });
 
-  createOrResetTwilioConnection();
+  const { _options, _iceServersStatus } = transport;
+  const { iceServers, onIced } = _options;
+
+  if (_iceServersStatus === 'overrode') {
```
makarandp0, line 624: this should not part be part of `setupEventListeners` - I suggest we move this call to constructor and also make `createOrResetTwilioConnection` class member. That way, you can clearly make onIced processing just for 1st connection.

lib/util/constants.js
```diff
@@ -10,40 +10,10 @@ module.exports.WS_SERVER = (environment, region) => {
     ? `wss://${region}.vss.twilio.com/signaling`
     : `wss://${region}.vss.${environment}.twilio.com/signaling`;
 };
-module.exports.ECS_SERVER = (environment, realm) => {
```
makarandp0, line 13: while at it, see if you can fix the WS_SERVER as well (by using new insights endpoint)

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 Nope. During reconnects, the `state` is in `reconnecting`, so the hello body will now contain a "sync" message.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 `setupEventListeners` is called only once anyway. So, I think this is fine where it is.

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,133 @@ const {
...
-          automaticSubscription: options.automaticSubscription,
-          dominantSpeaker: options.dominantSpeaker,
-          environment: options.environment,
-          eventObserver: options.eventObserver,
-          logLevel: options.logLevel,
-          networkMonitor: options.networkMonitor,
-          networkQuality: options.networkQuality,
-          iceServerSourceStatus: iceServerSource.status,
-          insights: options.insights,
-          realm: options.realm,
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I checked with Yuri, and they still have not deployed the new insights gateway everywhere. So, let's hold off on this for now.

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,133 @@ const {
...
-      // we are dequeueing to prevent the spurious 'update' message sent by
-      // the client after connecting to a room.
-      peerConnectionManager.dequeue('description');
+  let transport;
 
-      return new Promise((resolve, reject) => {
+  return new CancelablePromise((resolve, reject, isCanceled) => {
+    return new Promise((resolve, reject) => {
+      const onIced = iceServers => {
         if (isCanceled()) {
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 This is true when you are raising an exception in the `then()` success callback. But here, we already have access to the `reject` callback of the parent promise.

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,133 @@ const {
...
-          token,
-          localParticipant,
-          peerConnectionManager,
-          wsServer,
-          transportOptions);
-
-        transport.once('connected', function connected(initialState) {
+        return peerConnectionManager.createAndOffer().then(() => {
           if (isCanceled()) {
             reject(cancelationError);
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 While we are rejecting the parent promise here, we still have to reject the promise returned by `createAndOffer().then()`. So we throw.

lib/util/constants.js
manjeshbhargav: @makarandp0 New insights endpoint is not available in all regions yet.

lib/signaling/v2/cancelableroomsignalingpromise.js
manjeshbhargav: @makarandp0 You were emitting an "iced" event on the Transport and then calling a separate function called `resumeConnect()`. In this case, there is no need for `resumeConnect()` since `onIced` returns a Promise which can be used by the Transport to resume connecting.

lib/signaling/v2/twilioconnectiontransport.js
makarandp0: yes, you are right - its called just once, and also as you mentioned "connecting" state happens just once, so we are good. But I would still move it to constructor - since it feels outof place in `setupEventListeners`

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -527,6 +546,11 @@ function setupEventListeners(transport) {
         }
       case 'connecting':
         switch (message.type) {
+          case 'iced':
+            transport._options.onIced(message.ice_servers).then(() => {
+              transport._sendConnectOrSyncOrDisconnectMessage();
+            });
```
makarandp0, line 552: we should .catch` case here and disconnect. This will happen when user cancels the promise.

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,120 +12,137 @@ const {
...
-            return;
+            reject(cancellationError);
+            throw cancellationError;
           }
+          // NOTE(mmalavalli): PeerConnectionManager#createAndOffer() queues the
+          // initial offer in the event queue for the 'description' event. So,
+          // we are dequeueing to prevent the spurious 'update' message sent by
+          // the client after connecting to a room.
+          peerConnectionManager.dequeue('description');
+        }).catch(reject);
```
makarandp0, line 53:
```suggestion
        }).catch((err) => {
            log.error('error creating offer`, error);
            throw err;
        });
```

test/unit/spec/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -45,14 +39,30 @@ describe('createCancelableRoomSignalingPromise', () => {
...
+      const promise = test.onIced(iceServers);
+      test.createAndOfferDeferred.resolve();
+      return promise;
+    });
+
+    it('calls .setConfiguration on the newly-constructed PeerConnectionManager', () => {
+      assert.deepEqual(test.peerConnectionManager.setConfiguration.args[0][0].iceServers, iceServers);
+    });
+
+    it('calls .createAndOffer on the newly-constructed PeerConnectionManager', () => {
```
makarandp0, line 57: we should add test for createAndOffer rejecting.

lib/signaling/v2/cancelableroomsignalingpromise.js
makarandp0: ah, never mind that's not possible since it takes onIce as a parameter.

lib/signaling/v2/cancelableroomsignalingpromise.js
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav, I think you are shadowing outer `reject` with inner variable - We should really enable warning to not allow variable shadowing - but that means outer promise wont get rejected immediately. It will only get rejected in the `catch`  of the inner promise.

isn't 
```    reject(cancellationError);
          return Promise.reject(cancellationError);
``` 
same as simpler
``` 
throw cancellationError;
```
Thinking a bit about this, do we really need this inner `Promise` ?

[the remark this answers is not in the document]
makarandp0: Another suggestion is to move onIced outside the CancellablePromise.
``` 
const peerConnectionManager = new PeerConnectionManager(encodingParameters, preferredCodecs, options);
const trackSenders = flatMap(localParticipant.tracks, trackV2 => [trackV2.trackTransceiver]);
peerConnectionManager.setTrackSenders(trackSenders);
let cancellationError = null; // this will get set in onCancel handler if caller cancels.
const onIced = iceServers => {
  log.debug('Got ICE servers:', iceServers);
  if (cancellationError) {
    throw cancellationError;
  }
  options.iceServers = iceServers;
  peerConnectionManager.setConfiguration(options);

  return peerConnectionManager.createAndOffer().then(() => {
    log.debug('createAndOffer() succeeded.');
    if (cancellationError) {
      throw cancellationError;
    }
    // NOTE(mmalavalli): PeerConnectionManager#createAndOffer() queues the
    // initial offer in the event queue for the 'description' event. So,
    // we are dequeueing to prevent the spurious 'update' message sent by
    // the client after connecting to a room.
    peerConnectionManager.dequeue('description');
  }).catch(error => {
    log.error('createAndOffer rejected', error);
    throw error;
  });
};

return new CancelablePromise((resolve, reject, isCanceled) => {
  ...
}, () => {
  if (transport) {
    transport.disconnect();
    cancellationError = new Error('Canceled');
  }
});
```

lib/signaling/v2/cancelableroomsignalingpromise.js
makarandp0: ah makes sense. I think I like this approach too. But I do have some suggestion below to make the handling of cancellable promise and OnIced more streamlined.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 Disconnect is handled by the rejection of the CancelablePromise.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 What you're suggesting will require quite a bit of refactoring, so what I'll do is rename the function to `setupTransport()`, so that this function is not just about event listeners.

lib/signaling/v2/cancelableroomsignalingpromise.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 `onIced` has to be declared inside the CancelablePromise because, we need the `isCanceled` callback to reject the CancelablePromise. Also, we can't throw the first `cancellationError` because it is not being done within a Promise executor. Therefore we need `Promise.reject(cancellationError)`.

test/unit/spec/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -197,131 +222,25 @@ describe('createCancelableRoomSignalingPromise', () => {
...
-              test.cancelableRoomSignalingPromise.cancel();
-            });
-            return test.cancelableRoomSignalingPromise.then(() => {
-              throw new Error('Unexpected resolution');
-            }, () => {
-              assert(test.room.disconnect.calledOnce);
-            });
-          });
-
-          it('does not call .disconnect on the Transport', () => {
```
makarandp0, line 277: wondering why was this case there ? and why was it deleted? is that change in behavior?

lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -12,125 +12,140 @@ const {
...
-        if (options.InsightsPublisher) {
-          transportOptions.InsightsPublisher = options.InsightsPublisher;
-        }
+  let transport;
 
-        if (options.NullInsightsPublisher) {
-          transportOptions.NullInsightsPublisher = options.NullInsightsPublisher;
-        }
+  const cancelablePromise = new CancelablePromise((resolve, reject, isCanceled) => {
+    const onIced = iceServers => {
```
makarandp0, line 32: was moving onIced out of the promise not looking good? I thought it would make this code bit simpler, as onIced was pretty independent (with only complication was its dependency on `cancellationError `)

test/unit/spec/signaling/v2/cancelableroomsignalingpromise.js
manjeshbhargav: @makarandp0 These tests were making sure that if the CancelablePromise is canceled just after RoomV2 is created, then we should close the Transport and PeerConnectionManager. That is why we had a Promise within the CancelablePromise before. However, I noticed that in `lib/cancelableroompromise.js`, we do the same thing after `cancelableRoomSignalingPromise` is resolved. So, I just removed the redundant code so that we no longer need the Promise within the CancelablePromise.

lib/signaling/v2/cancelableroomsignalingpromise.js
manjeshbhargav: @makarandp0 If I move `onIced` out of the CancelablePromise, then I can't interrupt the events in the callback when the CancelablePromise is cancelled. So, it needs to be defined inside the CancelablePromise.

PR #1020, opened 2020-05-14
JSDK-2828: safari audio issues workaround: Monitor audio track and restart if ended

makarandp0:
This change implements a workaround to avoid problems with audio track when Safari is background. When safari (13.4.1)  is backgrounded and a Siri or some other app that uses audio is used, the safari's audio track goes into ended state. 

This change workarounds the issue by watching for track going in ended state, and also for document visibility changes. When document becomes visible and if track is in ended or muted state, the sdk will now silently try to re-aquire audio track and replace existing audio track.  

This still needs some work, but putting up the PR to get some early feedabck.

**Contributing to Twilio**


lib/media/track/localaudiotrack.js
```diff
@@ -80,6 +182,9 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
    * @fires LocalAudioTrack#stopped
    */
   stop() {
+    // track was stopped, stop watching, and don't start watching again.
+    this._startWatching(null);
+    this._wasStopped = true;
```
manjeshbhargav, line 187: @makarandp0 Nitpick: Can we call this `_didCallStop`? We have another member named `didCallEnd` and it would be nice to have the naming consistent.

lib/media/track/localaudiotrack.js
```diff
@@ -20,14 +22,114 @@ const LocalMediaAudioTrack = mixinLocalMediaTrack(AudioTrack);
...
+            }
+          }).catch(err => {
+            this._log.debug('Failed to acquire track', err);
+          });
+        }
+      }, 10);
+    }
+  }
+
+  _startWatching(track) {
```
manjeshbhargav, line 100: @makarandp0 I think this can be simplified to:
```js
class LocalAudioTrack {
  _didCallStop: { value: false, writable: true }

  _restart(constraints) {
    return super._restart(constraints).then(() => {
      this._didCallStop = true;
      this._restartWhenInadvertentlyStopped();
    });
  }

  _restartWhenInadvertentlyStopped() {
    waitForEvent(this, 'stopped').then(() => {
      if (this._didCallStop) {
        return;
      }
      if (document.visibilityState === 'visible') {
        return this._restart();
      }
      return waitForEvent(document, 'visibilitychange').then(() => {
        // Is this necessary?
        return waitForSomeTime(10);
      }).then(() => {
        return this._restart();
      });
    });
  }
}
```

manjeshbhargav:
@makarandp0 A couple more things:
* Every time a track is restarted, a "started" event should be fired again. This should be documented and tested.
* When a track is restarted, we should make sure `isEnabled`, `isStopped` are updated but `id`and `name` still point to the initial values. Also, we should make sure `LocalMediaTrack.id === LocalMediaTrack.mediaStreamTrack.id` by overwriting the new MediaStreamTrack's `id` property.

lib/createlocaltracks.js
```diff
@@ -102,6 +102,11 @@ function createLocalTracks(options) {
       : {}
   };
 
+  // eslint-disable-next-line no-undefined
+  if (options.audio && options.audio.workaroundWebKitBug1208516 !== undefined) {
+    localTrackNameOptions.audio.workaroundWebKitBug1208516 = options.audio.workaroundWebKitBug1208516;
+  }
+
```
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav, I think this is needed because we although we do pass localTrackOptions down https://github.com/twilio/twilio-video.js/blob/master/lib/createlocaltracks.js#L130 - They are in the wrong format. LocalAudioTrack constructor expects options to be `{ workaroundWebKitBug1208516: true }`
But here they are passed as: `{  audio: { workaroundWebKitBug1208516: true  } }` 
I think the way we pass it down is a bug.

lib/media/track/transceiver.js
```diff
@@ -25,7 +25,8 @@ class MediaTrackTransceiver extends TrackTransceiver {
       },
       track: {
         enumerable: true,
-        value: mediaStreamTrack
+        value: mediaStreamTrack,
+        writable: true,
```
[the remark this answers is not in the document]
makarandp0: Thanks for catching this. I will update.

lib/media/track/localaudiotrack.js
```diff
@@ -20,14 +22,114 @@ const LocalMediaAudioTrack = mixinLocalMediaTrack(AudioTrack);
...
+    const isDocumentVisible = document && document.visibilityState === 'visible';
+    return isDocumentVisible && (this.mediaStreamTrack.muted || this.mediaStreamTrack.readyState === 'ended');
+  }
+
+  _acquireTrack() {
+    const mediaStreamPromise = this._getUserMedia({ audio: true, video: false });
+    return mediaStreamPromise.then(mediaStream => {
+      return mediaStream.getAudioTracks()[0];
+    });
+  }
```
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav - I wanted to keep the workaround very localized, and did not want to make more changes than absolutely needed. Can we setup a quick meeting to talk about this?

lib/media/track/localaudiotrack.js
```diff
@@ -20,14 +22,114 @@ const LocalMediaAudioTrack = mixinLocalMediaTrack(AudioTrack);
...
+    // although the bug is seen  mainly on iOS devices, we do not have a reliable way to tell iOS from MacOs
+    // userAgent on iOS pretends its macOs if Safari is set to request desktop pages.
+    const workaroundWebKitBug1208516 = guessBrowser() === 'safari';
+
+    options = Object.assign({
+      workaroundWebKitBug1208516,
+      getUserMedia,
+    }, options);
+
     super(mediaStreamTrack, options);
```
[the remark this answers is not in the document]
makarandp0: is it okay to update the id for a LocalMediaTrack ? I am bit concerned about changing the id.  updating isEnabled/isStoppped makes sense though. .

lib/media/track/sender.js
```diff
@@ -26,7 +29,19 @@ class MediaTrackSender extends MediaTrackTransceiver {
...
+  }
+
+  replaceTrack(newTrack) {
+    this.track = newTrack;
+    this._senders.forEach(sender => {
+      if (sender.track) {
+        sender.replaceTrack(this.track).catch(() => {});
+      }
+    });
+    this._clones.forEach(clone => clone.replaceTrack(newTrack.clone()));
```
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav, I will update to return promise, however I am not very sure about not stopping the track if replaceTrack fails. Because I have seen replaceTrack failing often (because peerConnection being in closed state). I think its more like we need to remove sender when they are no more in use, but looks like its not happening.

lib/media/track/localaudiotrack.js
makarandp0: @manjeshbhargav - I looked into this alternative, and found that track does not reliably get ended event. For example in this code 
```
var stream = await navigator.mediaDevices.getUserMedia({ audio: true, video: false});
var track = stream.getAudioTracks()[0];
track.addEventListener('ended', () => console.log('ended'));
track.stop();
assert(track.readyState === 'ended');
```
we do not get ended event. I think I will keep my version.

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 We should make sure we honor the existing rule that `LocalMediaTrack.id === LocalMediaTrack.mediaStreamTrack.id`. Otherwise it can be considered as a breaking change.

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Let's talk more about this today.

lib/media/track/localaudiotrack.js
```diff
@@ -20,14 +23,115 @@ const LocalMediaAudioTrack = mixinLocalMediaTrack(AudioTrack);
...
+    }
+  }
+
+  _shouldAttemptRestart() {
+    const isDocumentVisible = document && document.visibilityState === 'visible';
+    return isDocumentVisible && (this.mediaStreamTrack.muted || this.mediaStreamTrack.readyState === 'ended');
+  }
+
+  _acquireTrack() {
+    const mediaStreamPromise = this._getUserMedia({ audio: true, video: false });
```
manjeshbhargav, line 72: @makarandp0 We can't just use `{ audio: true }` here. We should make sure existing constraints are used:
```js
_getUserMedia({ audio: this.mediaStreamTrack.getConstraints() });
```

lib/media/track/localaudiotrack.js
manjeshbhargav: @makarandp0 Let's discuss this today.

lib/media/track/localmediatrack.js
```diff
@@ -33,23 +33,40 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
         }
       });
     }
 
+    /**
+     * @private
+     * replaces underlying track and
+     * returns a promise.
+     */
+    _replaceTrack(track) {
```
manjeshbhargav, line 58: @makarandp0 Nitpick:
```suggestion
    _setMediaStreamTrack(mediaStreamTrack) {
```

lib/media/track/sender.js
```diff
@@ -14,9 +15,12 @@ class MediaTrackSender extends MediaTrackTransceiver {
   constructor(mediaStreamTrack) {
     super(mediaStreamTrack.id, mediaStreamTrack);
     Object.defineProperties(this, {
+      _clones: {
+        value: new Set()
+      },
       _senders: {
         value: new Set()
-      }
+      },
```
manjeshbhargav, line 23: @makarandp0 
```suggestion
      }
```

lib/media/track/sender.js
```diff
@@ -26,7 +30,27 @@ class MediaTrackSender extends MediaTrackTransceiver {
...
+   * @param {MediaStreamTrack} mediaStreamTrack
+   * @returns {Promise<void>}
+   */
+  setMediaStreamTrack(mediaStreamTrack) {
+    const clones = Array.from(this._clones);
+    const senders = Array.from(this._senders);
+    return Promise.all(clones.map(clone => {
+      return clone.setMediaStreamTrack(mediaStreamTrack.clone());
+    }).concat(senders.map(sender => {
+      return sender.replaceTrack(mediaStreamTrack).catch(() => {});
```
manjeshbhargav, line 49: @makarandp0 
```suggestion
      return sender.replaceTrack(mediaStreamTrack).catch(() => {
        log.warn(...);
      });
```

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
makarandp0: @manjeshbhargav  I think we need to honor the contract that the id of the LocalTrack does not change once created. id is a only unique field ( since sid is not available yet) that developers would  to associate the LocalTrack with dom element - if we changed the id they won't be able to associate it anymore and that would  break the applications. 
I suspect it's pretty common pattern.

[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I want to do the other way around.
```js
Object.defineProperties(mediaStreamTrack, {
  id: {
    enumerable: true,
    get() {
      return localMediaTrack.id;
    }
  }
});
```

lib/media/track/sender.js
makarandp0: adding log here would need some plumbing, and this error is seen often when peer connection is in closed state.

makarandp0: but let me add that warn anyways. It would be useful.

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
makarandp0: I am worried about this lets talk more today.

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
makarandp0: we will hold off on this and revisit when we consider @manjeshbhargav s proposal

lib/media/track/localaudiotrack.js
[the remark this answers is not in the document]
makarandp0: As we talked offline, decided to update the trackId that we sent in SDPs instead of changing id of the MediaStreamTrack

lib/media/track/localmediatrack.js
```diff
@@ -24,32 +24,52 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
       super(mediaTrackSender, options);
 
       Object.defineProperties(this, {
-        _didCallEnd: {
-          value: false,
-          writable: true
-        },
```
manjeshbhargav, line 30: Can we retain this here and then overload `_initialize`:
```js
_initialize() {
  this._didCallEnd = false;
  super._initialize.call(this);
}
```

lib/media/track/localaudiotrack.js
```diff
@@ -80,10 +110,71 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+        log.debug('replacing track with constrains ', newTrack.id, constraints);
+        return localAudioTrack._setMediaStreamTrack(newTrack);
+      }).catch(err => {
+        log.debug('Failed to acquire track', err);
+      });
+    });
+  }
+
+  // we want to re-acquire track if document becomes visible
+  // but the track has ended while page was background.
```
manjeshbhargav, line 168:
```suggestion
```

lib/media/track/localaudiotrack.js
```diff
@@ -80,10 +110,71 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+  document.addEventListener('visibilitychange', handleTrackStateChange);
+
+  // however on iOS when you quickly activate siri - while safari on foreground
+  // the page does not receive visibilitychange, but the audio track goes into
+  // ended state anyways. To catch such cases lets listen on ended event as well.
+  localAudioTrack.addListener('stopped', handleTrackStateChange);
```
manjeshbhargav, line 174:
```suggestion
  localAudioTrack.on('stopped', handleTrackStateChange);
```

lib/media/track/localmediatrack.js
```diff
@@ -24,32 +24,52 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
 
+    /**
+     * @private
+     * replaces underlying track and
+     * returns a promise.
+     */
+    _setMediaStreamTrack(mediaStreamTrack) {
+      // Note: ideally we would have liked to change the id
+      // of the new mediaStreamTrack to match exsiting id, but
+      // the property is readOnly and non configurable.
```
manjeshbhargav, line 57: We should document in CHANGELOG and LocalAudioTrack class that:
* `mediaStreamTrack` can change.
* `started` and `stopped` events can be fired more than once.
* What to do about any event listeners attached to the old MediaStreamTrack by the developer?
```suggestion
```

lib/media/track/localmediatrack.js
```diff
@@ -24,32 +24,52 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+      return this._trackSender.setMediaStreamTrack(mediaStreamTrack).catch(err => {
+        // log the error
+        this._log.warn('failed to replace track', err);
+      }).then(() => {
+        // Initialize the new MediaStreamTrack (defined in MediaTrack)
+        this._initialize();
+
+        // Attach the newly created MediaStreamTrack.
+        this._getAllAttachedElements().forEach(el => this._attach(el));
+      });
```
manjeshbhargav, line 70:
```suggestion
      // NOTE(mpatwardhan): Preserve the value of the "enabled" flag.
      mediaStreamTrack.enabled = this.mediaStreamTrack.enabled;

      return this._trackSender.setMediaStreamTrack(mediaStreamTrack).catch(err => {
        this._log.warn('failed to replace track', err);
      }).then(() => {
        this._initialize();
        this._getAllAttachedElements().forEach(el => this._attach(el));
      });
```

lib/media/track/mediatrack.js
```diff
@@ -36,6 +36,10 @@ class MediaTrack extends Track {
 
     /* istanbul ignore next */
     Object.defineProperties(this, {
+      _didCallEnd: {
+        value: false,
+        writable: true
+      },
```
manjeshbhargav, line 42:
```suggestion
```

lib/media/track/mediatrack.js
```diff
@@ -87,7 +93,7 @@ class MediaTrack extends Track {
    */
   _initialize() {
     const self = this;
-
+    this._didCallEnd = false;
```
manjeshbhargav, line 96:
```suggestion
```

lib/signaling/v2/peerconnection.js
```diff
@@ -940,6 +940,18 @@ class PeerConnectionV2 extends StateMachine {
...
+   * we sometimes replace media stream after publishing the track. In such cases published tracks id won't
+   * match the id of the mediaStream track, Fix it here to match original track id.
+   * @private
+   * @param {RTCSessionDescription} description
+   * @return {RTCSessionDescription}
+   */
+  _mapTrackIdToOriginalTrackId(id) {
+    const mediaTrackSender = [...this._rtpSenders.keys()].find(ms => ms.track.id === id);
+    return mediaTrackSender ? mediaTrackSender.id : id;
+  }
```
manjeshbhargav, line 953:
```suggestion
  /**
   * Get the MediaTrackSender ID of the given MediaStreamTrack ID.
   * Since a MediaTrackSender's underlying MediaStreamTrack can be
   * replaced, the corresponding IDs can mismatch.
   * @private
   * @param {Track.ID} id
   * @return {Track.ID}
   */
  _getMediaTrackSenderId(trackId) {
    const mediaTrackSender = Array.from(this._rtpSenders.keys()).find(({ track: { id } }) => id === trackId);
    return mediaTrackSender ? mediaTrackSender.id : trackId;
  }
```

lib/signaling/v2/peerconnection.js
```diff
@@ -955,15 +967,15 @@ class PeerConnectionV2 extends StateMachine {
     // actual MediaStreamTrack IDs. So, we add or re-write the actual MediaStreamTrack IDs
     // to the assigned m= sections here.
     const assignedTransceivers = activeTransceivers.filter(({ mid }) => mid);
-    const midsToTrackIds = new Map(assignedTransceivers.map(({ mid, sender }) => [mid, sender.track.id]));
+    const midsToTrackIds = new Map(assignedTransceivers.map(({ mid, sender }) => [mid, this._mapTrackIdToOriginalTrackId(sender.track.id)]));
```
manjeshbhargav, line 970:
```suggestion
    const midsToTrackIds = new Map(assignedTransceivers.map(({ mid, sender }) => [mid, this._getMediaTrackSenderId(sender.track.id)]));
```

lib/signaling/v2/peerconnection.js
```diff
@@ -955,15 +967,15 @@ class PeerConnectionV2 extends StateMachine {
...
     const sdp1 = unifiedPlanAddOrRewriteTrackIds(description.sdp, midsToTrackIds);
 
     // NOTE(mmalavalli): Chrome and Safari do not apply the offer until they get an answer.
     // So, we add or re-write the actual MediaStreamTrack IDs to the unassigned m= sections here.
     const unassignedTransceivers = activeTransceivers.filter(({ mid }) => !mid);
     const newTrackIdsByKind = new Map(['audio', 'video'].map(kind => [
       kind,
-      unassignedTransceivers.filter(({ sender }) => sender.track.kind === kind).map(({ sender }) => sender.track.id)
+      unassignedTransceivers.filter(({ sender }) => sender.track.kind === kind).map(({ sender }) => this._mapTrackIdToOriginalTrackId(sender.track.id))
```
manjeshbhargav, line 978:
```suggestion
      unassignedTransceivers.filter(({ sender }) => sender.track.kind === kind).map(({ sender }) => this._getMediaTrackSenderId(sender.track.id))
```

lib/signaling/v2/peerconnection.js
```diff
@@ -1369,6 +1381,11 @@ class PeerConnectionV2 extends StateMachine {
   }
 }
 
+function rewriteLocalTrackId(pcv2, stats) {
+  const trackId = pcv2._mapTrackIdToOriginalTrackId(stats.trackId);
```
manjeshbhargav, line 1385:
```suggestion
  const trackId = pcv2._getMediaTrackSenderId(stats.trackId);
```

test/integration/spec/localtracks.js
```diff
@@ -74,4 +105,92 @@ const createLocalTracks = require('../../../lib/createlocaltrack');
       });
     });
   });
+
+  describe('replaceTrack', function() {
```
manjeshbhargav, line 109: Please make sure to check that, after re-acquiring a Track:
* `started` is fired
* `enabled` is preserved
* Re-acquire logic also works for re-acquired Tracks.

lib/media/track/localaudiotrack.js
makarandp0: I have addressed this by logging in the caller side.

lib/media/track/localaudiotrack.js
```diff
@@ -20,14 +24,42 @@ const LocalMediaAudioTrack = mixinLocalMediaTrack(AudioTrack);
...
    * @param {MediaStreamTrack} mediaStreamTrack - An audio MediaStreamTrack
    * @param {LocalTrackOptions} [options] - {@link LocalTrack} options
    */
   constructor(mediaStreamTrack, options) {
+    // NOTE(mpatwardhan): by default workaround for WebKitBug1208516 will be enabled on Safari browsers
+    // although the bug is seen  mainly on iOS devices, we do not have a reliable way to tell iOS from MacOs
+    // userAgent on iOS pretends its macOs if Safari is set to request desktop pages.
+    const workaroundWebKitBug1208516 = guessBrowser() === 'safari' &&
+    // eslint-disable-next-line no-undefined
+    document && document.addEventListener && document.visibilityState !== undefined;
```
manjeshbhargav, line 40: Accessing `document` directly will result in a `TypeError` if its not present. In order to prevent this, we should use `typeof`. Also, it is not good practice to compare with `undefined` directly.
```suggestion
    const workaroundWebKitBug1208516 = guessBrowser() === 'safari'
      && typeof document === 'object'
      && typeof document.addEventListener === 'function'
      && typeof document.visibilityState === 'string';
```

lib/media/track/localaudiotrack.js
```diff
@@ -80,10 +112,51 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+function restartWhenInadvertentlyStopped(localAudioTrack) {
+  const { _getUserMedia: getUserMedia, _log: log } = localAudioTrack;
+
+  function shouldReacquireTrack() {
+    return document.visibilityState === 'visible'
+      && (localAudioTrack.mediaStreamTrack.muted || localAudioTrack.isStopped)
+      && !localAudioTrack._didCallStop;
+  }
+
+  function acquireTrack(constraints) {
```
manjeshbhargav, line 129: Nitpick:
```suggestion
  function reacquireTrack(constraints) {
```

lib/media/track/localaudiotrack.js
```diff
@@ -80,10 +112,51 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+  function handleTrackStateChange() {
+    Promise.race([
+      waitForEvent(localAudioTrack.mediaStreamTrack, 'unmute'),
+      waitForSometime(50)
+    ]).then(() => {
+      log.debug('checking if audio track needs to be replaced');
+      if (!shouldReacquireTrack()) {
+        return;
+      }
+      // replace tracks.
```
manjeshbhargav, line 144: Nitpick
```suggestion
```

lib/media/track/localmediatrack.js
```diff
@@ -33,23 +33,50 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+     * @private
+     * replaces underlying track and
+     * returns a promise.
+     */
+    _setMediaStreamTrack(mediaStreamTrack) {
+      // NOTE(mpatwardhan): Preserve the value of the "enabled" flag.
+      mediaStreamTrack.enabled = this.mediaStreamTrack.enabled;
+
+      return this._trackSender.setMediaStreamTrack(mediaStreamTrack).catch(err => {
+        // log the error
```
manjeshbhargav, line 63: Nitpick:
```suggestion
```

lib/media/track/mediatrack.js
```diff
@@ -87,7 +89,6 @@ class MediaTrack extends Track {
    */
   _initialize() {
     const self = this;
-
```
manjeshbhargav, line 90: Nitpick: Please add the line back.

lib/signaling/v2/peerconnection.js
```diff
@@ -940,6 +940,19 @@ class PeerConnectionV2 extends StateMachine {
     });
   }
 
+  /**
+   * Get the MediaTrackSender ID of the given MediaStreamTrack ID.
+   * Since a MediaTrackSender's underlying MediaStreamTrack can be
+   * replaced, the corresponding IDs can mismatch.
+   * @private
+   * @param {Track.ID} id
+   * @return {Track.ID}
```
manjeshbhargav, line 949:
```suggestion
   * @returns {Track.ID}
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.6.0 (in progress)
+===================
+
+- Worked around this iOS Safari [bug](https://bugs.webkit.org/show_bug.cgi?id=208516) which causes your application to lose the microphone when another application (Siri, YouTube, FaceTime, etc.) reserves the microphone. Now your application will regain the microphone after foregrounding. For this SDK reacquires mediaStream track on gaining foreground, causing `LocalAudioTrack.mediaStreamTrack` to change. SDK will emits 'started' event on LocalAudioTrack - you can use this event to setup listeners on new mediaStreamTrack. (JSDK-2828)
```
manjeshbhargav, line 6:
```suggestion
- Worked around this iOS Safari [bug](https://bugs.webkit.org/show_bug.cgi?id=208516) which causes your application to lose the microphone when another application (Siri, YouTube, FaceTime, etc.) reserves the microphone. Now your application will regain the microphone after foregrounding. As a result of this, the LocalAudioTrack's `mediaStreamTrack` property will now point to the newly acquired MediaStreamTrack, and the `started` event is fired again on the LocalAudioTrack. The `id` of the LocalAudioTrack is now no longer guaranteed to be equal to the `id` of the MediaStreamTrack. Also, if you want to listen to events on the MediaStreamTrack, we recommend that you do so in the `started` event handler, since it guarantees that you are always listening to events on the most recently acquired MediaStreamTrack. (JSDK-2828)
```js
const { createLocalAudioTrack } = require('twilio-video');

const localAudioTrack = await createLocalAudioTrack();

function onMute() {
  console.log('MediaStreamTrack muted!');
}

function onUnmute() {
  console.log('MediaStreamTrack unmuted!');
}

localAudioTrack.on('started', () => {
  const { mediaStreamTrack } = localAudioTrack;
  mediaStreamTrack.addEventListener('mute', onMute);
  mediaStreamTrack.addEventListener('unmute', onUnmute);
});

localAudioTrack.on('stopped', () => {
  const { mediaStreamTrack } = localAudioTrack;
  mediaStreamTrack.removeEventListener('mute', onMute);
  mediaStreamTrack.removeEventListener('unmute', onUnmute);
});
\`\`\`
```

lib/media/track/localaudiotrack.js
```diff
@@ -80,10 +113,50 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+        return;
+      }
+      const constraints = localAudioTrack.mediaStreamTrack.getConstraints();
+      reacquireTrack(constraints).then(newTrack => {
+        log.debug('replacing track with constrains ', newTrack.id, constraints);
+        return localAudioTrack._setMediaStreamTrack(newTrack);
+      }).catch(err => {
+        log.debug('Failed to acquire track', err);
+      });
+    });
```
manjeshbhargav, line 152: Please modify the log statements:
```suggestion
    ]).then(() => {
      if (!shouldReacquireTrack()) {
        return;
      }
      const constraints = localAudioTrack.mediaStreamTrack.getConstraints();
      log.info('Re-acquiring the MediaStreamTrack.');
      log.debug('Constraints:', constraints);
      return reacquireTrack(constraints).then(newTrack => {
        log.info('Re-acquired the MediaStreamTrack.');
        log.debug('MediaStreamTrack:', newTrack);
        return localAudioTrack._setMediaStreamTrack(newTrack);
      }).catch(err => {
        log.warn('Failed to re-acquire the MediaStreamTrack:', err);
      });
    });
```

lib/media/track/localmediatrack.js
```diff
@@ -33,23 +33,49 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+     * @private
+     * replaces underlying track and
+     * returns a promise.
+     */
+    _setMediaStreamTrack(mediaStreamTrack) {
+      // NOTE(mpatwardhan): Preserve the value of the "enabled" flag.
+      mediaStreamTrack.enabled = this.mediaStreamTrack.enabled;
+
+      return this._trackSender.setMediaStreamTrack(mediaStreamTrack).catch(err => {
+        this._log.warn('failed to replace track', err);
```
manjeshbhargav, line 63:
```suggestion
        this._log.warn('setMediaStreamTrack failed:', err, mediaStreamTrack);
```

PR #1032, opened 2020-05-27
Prep for 2.5.0-rc3.

manjeshbhargav:
@makarandp0 @ceaglest 

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/cancelableroomsignalingpromise.js
+++ b/lib/signaling/v2/cancelableroomsignalingpromise.js
@@ -123,10 +123,11 @@ function createCancelableRoomSignalingPromise(token, wsServer, localParticipant,
     });
 
     transport.once('stateChanged', (state, error) => {
-      log.debug('Transport state changed:', state, error);
       if (state === 'disconnected') {
         transport = null;
         reject(error || new SignalingConnectionDisconnectedError());
+      } else {
+        log.debug('Transport state changed:', state);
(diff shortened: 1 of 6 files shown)


lib/signaling/v2/cancelableroomsignalingpromise.js
```diff
@@ -136,13 +137,12 @@ function createCancelableRoomSignalingPromise(token, wsServer, localParticipant,
...
   });
 
-  cancelablePromise.catch(error => {
+  cancelablePromise.catch(() => {
     if (transport) {
       transport.disconnect();
       transport = null;
     }
     peerConnectionManager.close();
-    throw error;
```
makarandp0, line 145: curious - why was this throw removed? Which case was it causing an issue?

makarandp0, line 145: if not throw at least we should log the error here.

manjeshbhargav: @makarandp0 This will never be caught because we are returning the CancelablePromise and not the Promise returned by `catch()`. So, anything thrown here will be displayed as an uncaught error in the JS console.

manjeshbhargav: @makarandp0 We already log this in several other places.

PR #1039, opened 2020-05-29
JSDK-2832 Stop supporting prefixed WebRTC APIs.

manjeshbhargav:
@makarandp0

This PR makes sure we stop supporting Chrome (23 - 55) and Firefox (22 - 43) versions that support only the prefixed webkit and moz WebRTC APIs that are now deprecated.

consumes: https://github.com/twilio/twilio-webrtc.js/pull/123

**Contributing to Twilio**

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -1,4 +1,4 @@
-/* globals RTCPeerConnection, webkitRTCPeerConnection, mozRTCPeerConnection, chrome, navigator */
+/* globals chrome, navigator, RTCPeerConnection */
 'use strict';
 
 const { guessBrowser } = require('@twilio/webrtc/lib/util');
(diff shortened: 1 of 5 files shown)


lib/util/support.js
```diff
@@ -11,24 +11,14 @@ const SUPPORTED_CHROME_BASED_BROWSERS = [
...
-/**
- * Check whether GetUserMedia API is supported.
- * @returns {boolean}
- */
-function isGetUserMediaSupported() {
-  return !!(navigator.mediaDevices && navigator.mediaDevices.getUserMedia)
-    || !!(navigator.getUserMedia)
-    || !!(navigator.webkitGetUserMedia)
-    || !!(navigator.mozGetUserMedia);
+function isWebRTCSupported() {
```
makarandp0, line 17: this function should go into twilio-webrtc, and used from here.

lib/util/support.js
```diff
@@ -1,7 +1,7 @@
-/* globals chrome, navigator, RTCPeerConnection */
+/* globals chrome, navigator */
 'use strict';
 
-const { guessBrowser } = require('@twilio/webrtc/lib/util');
+const { guessBrowser, support: isWebRTCSupported } = require('@twilio/webrtc/lib/util');
```
makarandp0, line 4: isnt `support` a very generic name? I suggest rename it to `isWebRTCSupported`?

manjeshbhargav: @makarandp0 Since it is a util method within the WebRTC lib, I think its fine. That is why I'm renaming it when I'm consuming the method in twilio-video.js. I'm following the same naming convension as the `support()` function in twilio-video.js.

makarandp0: @mmalavalli - Do you mean twilio-video.isSupported property? I think that would be fine too or even better isWebRTCSupported, but export name `support()` sounds very confusing to me.

manjeshbhargav: @makarandp0 I think `TwilioWebRTC.isWebRTCSupported` is redundant.

makarandp0: @mmalavalli - I think the difference is twilio-webrtc is a public module, and its exports are not internal (unlike `support` function that is completely internal to twilio-video.js.
But Generally I think you are better at naming than me and if you really think it needs to be twilio-webrtc.support,   Lets go for it.

PR #1042, opened 2020-05-30
JSDK-2841 TCMP client transitions from "waiting" to "open" upon receiving "welcome".

manjeshbhargav:
@makarandp0 

**Contributing to Twilio**

### DIFF
--- a/lib/twilioconnection.js
+++ b/lib/twilioconnection.js
@@ -444,7 +444,7 @@ class TwilioConnection extends StateMachine {
       return;
     }
 
-    if (this.state === 'waiting' && this._busyWaitTimeout) {
+    if (this.state === 'waiting') {
       log.debug('Received "welcome" message, no need to retry connection.');
       this._busyWaitTimeout.clear();
     }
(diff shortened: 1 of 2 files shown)


lib/twilioconnection.js
```diff
@@ -436,18 +436,21 @@ class TwilioConnection extends StateMachine {
...
+
+    if (!['connecting', 'waiting'].includes(this.state)) {
+      log.warn(`Unexpected state "${this.state}" for handling a "welcome"`
         + ' message from the TCMP server.');
       return;
     }
 
-    const { maxConsecutiveMissedHeartbeats } = this._options;
+    if (this.state === 'waiting' && this._busyWaitTimeout) {
+      log.debug('Received "welcome" message, no need to retry connection.');
```
makarandp0, line 448: The check for `this._busyWaitTimeout` is unnecessary because it will be always true when state is waiting.
We never set this._busyWaitTimeout to null. Maybe the check needs to be changed to   `this._busyWaitTimeout.isSet()` or removed altogether.

PR #1046, opened 2020-06-04
Unpause media when foregrounded

makarandp0:
@manjeshbhargav 

Because of webkit bug: https://bugs.webkit.org/show_bug.cgi?id=212780
Mobile Safari pauses media elements (audio/video) when Safari is backgrounded (and some other application start using audio), but sometimes it fails to play them when safari gains foreground. This PR workarounds this issue by remembering the playing elements when losing foreground, and restoring them when safari is back on foreground. 

**Contributing to Twilio**


lib/media/track/remotemediatrack.js
```diff
@@ -137,6 +151,35 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
   };
 }
 
+function playIfPausedWhileInBackground(audioTrack) {
+  const { _log: log } = audioTrack;
+  let wasVisible = document.visibilityState === 'visible';
+  let elementsPlaying = [];
+
+  document.addEventListener('visibilitychange', () => {
```
manjeshbhargav, line 159: @makarandp0 I think this will result in event listener leaks when this RemoteMediaTrack is unsubscribed. I would rather listen to the `pause` event on the elements:
```js
function playIfPausedWhileInBackground(remoteMediaTrack) {
  remoteMediaTrack._attachments.forEach(el => {
    el.addEventListener('pause', () => {
      if (document.visibilityState === 'hidden') {
        // This ensures no event listener leaks as it is removed as soon as it is fired once.
        waitForEvent(document, 'visibilitychange').then(() => el.play());
      }
    });
  });
}
```

manjeshbhargav: @makarandp0 Also, we need to make sure we detach the pause event listener when the element is `detach()`ed.

makarandp0: Thanks @manjeshbhargav for catching the listener leak. One problem I see with listening on individual elements is they are added/removed dynamically, requiring that we hook/unhook on them individually as they are attach/detach'ed. It also means maintaining a map of el => listener so that we can remove those listeners when the elements are detached.

What do you think about keep listening on document visibility, but unhook listener when there are no elements attached. I have updated code to unhook the listener now.

makarandp0:
@manjeshbhargav - I have addressed your feedback after our chat. Please take another look.

makarandp0:
@manjeshbhargav - I pushed another change that moves the "re-acquire media" workaround so that it applies for video tracks as well.

lib/media/track/localmediatrack.js
```diff
@@ -104,11 +131,60 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
     }
   };
 }
 
+function restartWhenInadvertentlyStopped(localMediaTrack) {
+  const { _getUserMedia: getUserMedia, _log: log } = localMediaTrack;
+
+  function shouldReacquireTrack() {
+    return document.visibilityState === 'visible'
+      && (localMediaTrack.mediaStreamTrack.muted || localMediaTrack.isStopped);
```
manjeshbhargav, line 150:
```suggestion
    const { _workaroundWebKitBug1208516Cleanup, isStopped, mediaStreamTrack:  { muted } } = localMediaTrack;
    const isInadvertentlyStopped = isStopped && !!_workaroundWebKitBug1208516Cleanup;
    return document.visibilityState === 'visible' && (muted || isInadvertentlyStopped);
```

lib/media/track/localmediatrack.js
```diff
@@ -104,11 +131,60 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+        log.debug('MediaStreamTrack:', newTrack);
+        localMediaTrack._setMediaStreamTrack(newTrack);
+      }).catch(err => {
+        log.warn('Failed to re-acquire the MediaStreamTrack:', err);
+      });
+    });
+  }
+
+  document.addEventListener('visibilitychange', handleTrackStateChange);
+  localMediaTrack.on('stopped', handleTrackStateChange);
```
manjeshbhargav, line 183: Wouldn't this risk re-acquiring the track twice?

lib/media/track/remotemediatrack.js
```diff
@@ -131,12 +151,47 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+          this._workaroundWebKitBug212780Cleanup();
+          this._workaroundWebKitBug212780Cleanup = null;
+        }
       }
       return result;
     }
   };
 }
 
+function playIfPausedWhileInBackground(audioTrack) {
```
manjeshbhargav, line 166:
```suggestion
function playIfPausedWhileInBackground(remoteMediaTrack) {
```

lib/media/track/remotemediatrack.js
```diff
@@ -131,12 +151,47 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+  const { _log: log } = audioTrack;
+  let elementsPlaying = [];
```
manjeshbhargav, line 168:
```suggestion
  let elementsPlaying = [];
  
```

lib/media/track/remotemediatrack.js
```diff
@@ -131,12 +151,47 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+  function onVisibilityChanged() {
+    const isVisible = document.visibilityState === 'visible';
+    if (!isVisible) {
+      // when document goes to background, remember currently playing elements.
```
manjeshbhargav, line 172:
```suggestion
```

lib/media/track/remotemediatrack.js
```diff
@@ -131,12 +151,47 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+      elementsPlaying = [...audioTrack._attachments.values()].filter(el => !el.paused);
+    } else {
+      // when document regains foreground restore playing elements.
```
manjeshbhargav, line 175:
```suggestion
```

lib/media/track/localmediatrack.js
```diff
@@ -104,11 +131,60 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+      waitForEvent(localMediaTrack.mediaStreamTrack, 'unmute'),
+      waitForSometime(50)
+    ]).then(() => {
+      if (!shouldReacquireTrack()) {
+        return;
+      }
+      reacquireTrack().then(newTrack => {
+        log.info('Re-acquired the MediaStreamTrack.');
+        log.debug('MediaStreamTrack:', newTrack);
+        localMediaTrack._setMediaStreamTrack(newTrack);
```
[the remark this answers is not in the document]
makarandp0: return here is not necessary since this is event  handler. But I see your point of being consisten with return, I will fold this into `shouldReacquireTrack` check.

lib/media/track/localmediatrack.js
makarandp0: not really, because even if both handlers fire closer to each other `shouldReacquireTrack` would return false for one of them.

lib/media/track/remotemediatrack.js
makarandp0: I think this comment is useful

lib/media/track/remotemediatrack.js
makarandp0: I think this comment is useful

lib/media/track/localmediatrack.js
makarandp0: Thank you for noticing this. - maybe there is a slight chance that they may both endup obtaining track. I will fix this part.

lib/media/track/localmediatrack.js
```diff
@@ -104,11 +131,60 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+    log.info('Re-acquiring the MediaStreamTrack.');
+    log.debug('Constraints:', constraints);
+    return getUserMedia(constraints).then(mediaStream => {
+      return mediaStream.getTracks()[0];
+    });
+  }
+
+  function handleTrackStateChange() {
+    Promise.race([
+      waitForEvent(localMediaTrack.mediaStreamTrack, 'unmute'),
```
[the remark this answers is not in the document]
makarandp0: this is only place its referenced in the function,  destructure'd variable is unnecessary.

makarandp0:
@manjeshbhargav, I have addressed your feedback and updated this PR to point to new `develop` branch. Please take another look.

lib/media/track/remotemediatrack.js
manjeshbhargav: Please prefix with `NOTE(mpatwardhan):`.

lib/media/track/remotemediatrack.js
manjeshbhargav: Please prefix with `NOTE(mpatwardhan):`.

CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
...
+localAudioTrack.on('stopped', () => {
+  const { mediaStreamTrack } = localAudioTrack;
+  mediaStreamTrack.removeEventListener('mute', onMute);
+  mediaStreamTrack.removeEventListener('unmute', onUnmute);
+});
+```
+
+- Worked around iOS Safari [bug](https://bugs.webkit.org/show_bug.cgi?id=212780) which caused remote
+  tracks to stay in paused state, even after safari gains foreground. To workaround this twilio-video
+  will now play the remote tracks that were paused when safari gains foreground.
```
manjeshbhargav, line 43:
```suggestion
- Worked around iOS Safari [bug](https://bugs.webkit.org/show_bug.cgi?id=212780) where, when the application is foregrounded, it sometimes does not resume playback of the HTMLMediaElements attached to RemoteTracks that are paused when the application is backgrounded.
```

CHANGELOG.md
```diff
@@ -1,5 +1,47 @@
 For 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
+2.6.0 (in progress)
+===================
```
manjeshbhargav, line 4:
```suggestion
===================

Changes
-------
```

PR #1056, opened 2020-06-09
JSDK-2880 - do not use EventEmitter.off

makarandp0:
This addresses JSDK-2880 found by QA - which is caused by usage of `off` on EventEmitter which was missing in EventEmitter prototype generated by Simpler Signaling. The same function does exist in code generated by VBT or AhoyApp.

For unblocking testing on the safari workaround - I am fixing this usage. But we do need to investigate further to see which part of the Simpler-Singaling Build is causing this issue, and if there are any other functio usages that need to be fixed.

  

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/localmediatrack.js
+++ b/lib/media/track/localmediatrack.js
@@ -187,7 +187,7 @@ function restartWhenInadvertentlyStopped(localMediaTrack) {
   }
 
   document.addEventListener('visibilitychange', handleTrackStateChange);
-  localMediaTrack.addListener('stopped', handleTrackStateChange);
+  localMediaTrack.on('stopped', handleTrackStateChange);
   return () => {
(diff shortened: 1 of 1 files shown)


lib/media/track/localmediatrack.js
```diff
@@ -187,10 +187,10 @@ function restartWhenInadvertentlyStopped(localMediaTrack) {
   }
 
   document.addEventListener('visibilitychange', handleTrackStateChange);
-  localMediaTrack.on('stopped', handleTrackStateChange);
+  localMediaTrack.addListener('stopped', handleTrackStateChange);
```
manjeshbhargav, line 190: Nitpick: We use `on` to attach listeners.
```suggestion
  localMediaTrack.on('stopped', handleTrackStateChange);
```

PR #1062, opened 2020-06-12
JSDK-2870 LocalMediaTrack.restart()

manjeshbhargav:
@makarandp0 

This PR implements `LocalMediaTrack.restart()`.

### Resources

- [Proposal](https://docs.google.com/document/d/10Rh8Sz0PjiKmN7XZc9cgrPS02yOtv6APQNsjiIWvxog/edit?usp=sharing)
- [IDL changes](https://code.hq.twilio.com/client/video-sdk-api/pull/68)

### TODO

- [x] Implementation
- [x] Unit tests
- [x] Integration tests
- [x] Documentation
- [x] Changelog

**Contributing to Twilio**


lib/media/track/localmediatrack.js
```diff
@@ -103,16 +106,72 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+    _restart(constraints) {
+      const { _log: log, mediaStreamTrack } = this;
+      constraints = constraints || mediaStreamTrack.getConstraints();
+      return this._reacquireTrack(constraints).catch(error => {
+        log.error('Failed to re-acquire the MediaStreamTrack:', error, constraints);
+        throw error;
+      }).then(newMediaStreamTrack => {
+        log.info('Re-acquired the MediaStreamTrack');
+        log.debug('MediaStreamTrack:', newMediaStreamTrack);
+        return this._setMediaStreamTrack(newMediaStreamTrack);
```
makarandp0, line 145: we should to catch and log errors from `_setMediaStreamTrack` and not propagate them.

lib/media/track/localmediatrack.js
```diff
@@ -129,21 +188,28 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
       return this.enable(false);
     }
 
+    restart(constraints) {
```
makarandp0, line 191: do you think its valid usage to call `localTrack.restart()` I think it would be useful when say a usb device is attached/detached. To support such case we should add line:
``` 
constraints = constraints === `undefined` ? true;
```
if you think that usage is something that we shouldn't allow then we should validate that constrains are not `undefined` - 

Also what happens if constraints given were `{}` ?

lib/media/track/localmediatrack.js
manjeshbhargav: We already do that [here](https://github.com/twilio/twilio-video.js/pull/1062/files#diff-65a9cef12e9eb87a881dd86ac35ad263R160).

lib/media/track/localmediatrack.js
manjeshbhargav: @makarandp0 You bring up a good point. We can use the default device by setting the constraints as `{}`. I will document this.

I think it makes sense to have `LocalMediaTrack.restart()` re-use existing constraints unless a constraints object is specified.

lib/createlocaltracks.js
```diff
@@ -102,14 +102,15 @@ function createLocalTracks(options) {
       : {}
   };
 
-  // eslint-disable-next-line no-undefined
-  if (options.audio && options.audio.workaroundWebKitBug1208516 !== undefined) {
-    localTrackNameOptions.audio.workaroundWebKitBug1208516 = options.audio.workaroundWebKitBug1208516;
+  extraLocalTrackOptions.audio.isCreatedByCreateLocalTracks = true;
```
makarandp0, line 105: since this is always assigned I would move it inside the declaration of `extraLocalTrackOptions` like
```
const extraLocalTrackOptions = {
   audio: { isCreatedByCreateLocalTracks: true },
  video: { isCreatedByCreateLocalTracks: true}
}
```

and then add `name` and `workaroundWebKitBug1208516` conditionally here.

test/unit/spec/media/track/localmediatrack.js
```diff
@@ -222,6 +209,103 @@ const { defer } = require('../../../../../lib/util');
...
     });
 
+    describe('#restart', () => {
+      combinationContext([
+        [
+          [true, false],
+          x => `when called on a ${description} that was ${x ? '' : 'not '}created using either createLocalTracks or create${description}`
+        ],
+        [
+          [undefined, { foo: 'bar' }],
```
makarandp0, line 219: we should add a case for `{}` since its bit special.

test/unit/spec/media/track/localmediatrack.js
```diff
@@ -222,6 +209,103 @@ const { defer } = require('../../../../../lib/util');
...
+          x => `when called with${x ? '' : 'out'} constraints`
+        ],
+        [
+          ['rejects with an Error', 'resolves with a MediaStream'],
+          x => `when getUserMedia ${x}`
+        ]
+      ], ([isCreatedByCreateLocalTracks, constraints, gUMBehavior]) => {
```
makarandp0, line 226:
```suggestion
          ['rejects with an Error', 'resolves with a MediaStream'],
          x => `when getUserMedia ${x ? ''resolves with a MediaStream' : 'rejects with an Error' }`
        ]
      ], ([isCreatedByCreateLocalTracks, constraints, gUMSucceeds]) => {
```

test/unit/spec/media/track/localmediatrack.js
```diff
@@ -222,6 +209,103 @@ const { defer } = require('../../../../../lib/util');
...
+        let mediaStream;
+        let track;
+
+        before(async () => {
+          getUserMedia = /rejects/.test(gUMBehavior) ? sinon.spy(() => {
+            return Promise.reject(new Error('foo'));
+          }) : sinon.spy((...args) => fakeGetUserMedia(...args).then(stream => {
+            mediaStream = stream;
+            return stream;
+          }));
```
makarandp0, line 239:
```suggestion
          getUserMedia = sinon.spy((...args) => {
              return gUMSucceeds ? fakeGetUserMedia(...args).then(stream => {
                mediaStream = stream;
                return stream;
              }) : return Promise.reject(new Error('foo')) ;
          })
```

package.json
```diff
@@ -31,7 +31,7 @@
     "babel-preset-es2015": "^6.24.1",
     "browserify": "^14.3.0",
     "cheerio": "^0.22.0",
-    "chromedriver": "^2.28.0",
+    "chromedriver": "^83.0.0",
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 This fixes framework tests which do not like the old version of ChromeDriver.

test/integration/spec/localtracks.js
```diff
@@ -107,174 +108,188 @@ async function assertMediaFlow(room, mediaFlowExpected,  errorMessage) {
...
+      const stoppedPromise = waitForEvent(bobLocalTrackA, 'stopped');
 
-    // stop track and its clones.
-    bobVideoTrackA._trackSender.track.stop();
-    bobVideoTrackA._trackSender._clones.forEach(clone => clone.track.stop());
+      // Bob restarts track.
+      await bobLocalTrackA.restart();
+
+      // "stopped" and "started" events should fire in order.
+      await stoppedPromise;
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I want to make sure that the Promises are resolved in order, but good idea about using `waitFor`.

test/unit/spec/media/track/localmediatrack.js
manjeshbhargav: @makarandp0 I'm not sure about doing it in a unit test, since we cannot verify the behavior. We can write an integration test, but I don't think it will work if we don't run it in an environment where there is more than one device available.

lib/createlocaltracks.js
manjeshbhargav: @makarandp0 I think this is OK for now.

PR #1068, opened 2020-06-22
synchronize documentVisibility handling

makarandp0:
This change 
- adds a  `DocumentVisibilityMonitor` module to ensure correct sequence between re-acquiring tracks and playing remote media. No more `delay(2000)`
- uses detectSilenceTrack workaround, as sometimes we saw track being silence (all packets with value of 128) 
- waits for `replaceTrack` operations to finish - to fix an issue that caused LocalMediaTrack not listening on 'ended' event after re-acquiring tracks. 

**Contributing to Twilio**


lib/createlocaltracks.js
```diff
@@ -1,8 +1,21 @@
+/* eslint-disable no-console */
 'use strict';
 
 const asLocalTrack = require('./util').asLocalTrack;
 const buildLogLevels = require('./util').buildLogLevels;
-const getUserMedia = require('@twilio/webrtc').getUserMedia;
+const getUserMediaReal = require('@twilio/webrtc').getUserMedia;
```
makarandp0, line 6: this is temporary code. need to be removed before merge.

lib/media/track/localmediatrack.js
```diff
@@ -167,36 +170,52 @@ function restartWhenInadvertentlyStopped(localMediaTrack) {
...
+          mediaStreamTrack.addEventListener('ended', onStreamEnded);
           trackChangeInProgress = false;
         });
       }
     });
   }
 
-  document.addEventListener('visibilitychange', handleTrackStateChange);
-  mediaStreamTrack.addEventListener('ended', handleTrackStateChange);
+  function onStreamEnded() {
```
makarandp0, line 205: this is temporary change to add logging, need to be removed before merge.

lib/util/documentvisibilitymonitor.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+const supportedPhases = ['phase1', 'phase2'];
+/**
+ * exposes a mechanism for listening for document visibility in
+ * ordered phases.
+ * caller can subscriber for document visibility on phase1 and phase2
+ * phase1 callbacks will be executed first. Those callbacks may return async.
+ * phase2 callbacks will be executed only after phase1 callbacks are settled.
+ */
+class DocumentVisibilityMonitor {
```
manjeshbhargav, line 16: @makarandp0 I think this class can be simplified a great deal like this:
```js
'use strict';

/**
 * The {@link DocumentVisibilityMonitor} monitors the visibility state of the DOM
 * and executes the attached listeners in phase order when the DOM is visible.
 */
class DocumentVisibilityMonitor {
  /**
   * Constructor.
   * @param {number} [nPhases=1] - the number of phases
   */
  constructor(nPhases = 1) {
      _listeners: {
        value: []
      },
      _nPhases: {
        value: nPhases
      },
      _onVisibilityChange: {
        value: () => {
          if (document.visibilityState === 'visible') {
            this._emitVisible();
          }
        }
      }
    });

    for (let i = 0; i < nPhases; i++) {
      this._listeners.push(new Set());
    }
  }

  get _listenerCount() {
    return this._listeners.reduce((count, phaseListeners) => count + phaseListeners.size, 0);
  }

  /**
   * Call all the listeners. Makes sure that all listeners for a given phase
   * are executed before calling the listeners of the next phase.
   * @private
   */
  _emitVisible() {
    let promise = Promise.resolve();
    for (let phase = 1; phase <= this._nPhases; phase++) {
      promise = promise.then(() => this._emitVisiblePhase(phase));
    }
    return promise;
  }

  /**
   * Call all the listeners for a given phase.
   * @private
   */
  _emitVisiblePhase(phase) {
    const phaseListeners = Array.from(this._listeners[phase - 1]);
    return Promise.all(phaseListeners.map(listener => {
      const ret = listener();
      return ret instanceof Promise ? ret : Promise.resolve(ret);
    }));
  }

  /**
   * Start listening to the DOM visibility state change.
   * @private
   */
  _start() {
    document.addEventListener('visibilitychange', this._onVisibilityChange);
  }

  /**
   * Stop listening to the DOM visibility state change.
   * @private
   */
  _stop() {
    document.removeEventListener('visibilitychange', this._onVisibilityChange);
  }

  /**
   * Listen for the DOM to be visible at the given phase.
   * @param {number} phase
   * @param {function} listener
   * @returns {this}
   */
  onVisible(phase, listener) {
    this._listeners[phase - 1].add(listener);
    if (this._listenerCount === 1) {
      this._start();
    }
    return this;
  }

  /**
   * Stop listening for the DOM to be visible at the given phase.
   * @param {number} phase
   * @param {function} listener
   * @returns {this}
   */
  offVisible(phase, listener) {
    this._listeners[phase - 1].delete(listener);
    if (this._listenerCount === 0) {
      this._stop();
    }
    return this;
  }
}

module.exports = new DocumentVisibilityMonitor(2);
```

lib/util/index.js
```diff
@@ -403,7 +403,7 @@ function buildLogLevels(logLevel) {
   if (typeof logLevel === 'string') {
     return {
       default: logLevel,
-      media: logLevel,
+      media: 'debug', // TODO(mpatwardhan): restore before release.
```
manjeshbhargav, line 406: Didn't you already do this in [createLocalTracks](https://github.com/twilio/twilio-video.js/pull/1068/files#diff-10abc4ab7f10b4544b8f050f56e9e9c3R16)?

lib/webaudio/detectsilence.js
```diff
@@ -40,7 +40,8 @@ function detectSilence(audioContext, stream, timeout) {
       return Promise.resolve(true);
     }
     analyser.getByteTimeDomainData(samples);
-    return samples.some(sample => sample)
+    // Note(mpatwardhan): we find that silence is sometimes all 128s and sometimes all 0s
```
manjeshbhargav, line 43:
```suggestion
    // NOTE(mpatwardhan): An audio MediaStreamTrack can be silent either due to all samples being equal to 128 or all samples being equal to 0.
```

lib/createlocaltracks.js
```diff
@@ -12,7 +12,8 @@ const {
 
 const MediaStreamTrack = require('@twilio/webrtc').MediaStreamTrack;
 const Log = require('./util/log');
-const DEFAULT_LOG_LEVEL = require('./util/constants').DEFAULT_LOG_LEVEL;
+// const DEFAULT_LOG_LEVEL = require('./util/constants').DEFAULT_LOG_LEVEL;
+const DEFAULT_LOG_LEVEL = 'debug'; // TODO(mpatwardhan): restore before release.
```
manjeshbhargav, line 16:
```suggestion
// const DEFAULT_LOG_LEVEL = require('./util/constants').DEFAULT_LOG_LEVEL;
// TODO(mpatwardhan): restore before release.
const DEFAULT_LOG_LEVEL = 'debug';
```

lib/util/index.js
makarandp0: I found that change in createLocalTracks wasn't sufficient - it wasn't propagated to `Track`. This ensures that all tracks get it.

lib/util/documentvisibilitymonitor.js
makarandp0: @manjeshbhargav, I like your suggestion in general, but it does look like we are implementing some of eventEmitter here. also while running my unit tests against your suggestion, one immediate issue I found is the use of `Set` to store listeners - that breaks the case where same callback is registered multiple times. But I think overall this looks good, and since you have already taken effort to write this - I will fix the unit tests, and get it working.

lib/util/index.js
manjeshbhargav: @makarandp0 In that case you don't need to do it in `createLocalTracks`.

lib/util/documentvisibilitymonitor.js
manjeshbhargav: @makarandp0 We're only recreating a little bit of EventEmitter. Also, in your implementation, EventEmitter is only used to store listeners and nothing is emitted on the EventEmitter. So, we can just maintain the queue ourselves in a simpler fashion.

lib/media/track/remotemediatrack.js
```diff
@@ -213,33 +212,31 @@ function shimMediaElement(mediaElement) {
...
         });
-      });
-    }
+      }
+    });
   }
 
-  document.addEventListener('visibilitychange', onVisibilityChanged);
+  // NOTE(mpatwardhan): listen for document visibility callback on phase 2.
+  // this ensures that any LocalMediaTrack (which listen on phase 1) get executed
```
manjeshbhargav, line 234:
```suggestion
  // this ensures that any LocalMediaTrack's restart (which listen on phase 1) gets executed
```

lib/util/documentvisibilitymonitor.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+      }
+    });
+
+    for (let i = 0; i < nPhases; i++) {
+      this._listeners.push([]);
+    }
+  }
+
+  _listenerCount() {
```
manjeshbhargav, line 31: I think this can be made into a getter property
```suggestion
  get _listenerCount() {
```

makarandp0: maybe we should talk more about this - but I think functions are in general better choice than properties,
 because 
1) they make it obvious that some processing in involved 
2) it avoids bugs like
``` 
if (foo._llistenerCount) {
  // do something useful
}
```
where the property does not really exist...but code wont fail immediately, thus causing bugs. 

3) they are bit easier to work with when testing/spying/stubbing on them

PR #1103, opened 2020-07-16
JSDK-2914 Work around Chromium bug 1106157 by disabling RTX for Firefox 79+.

manjeshbhargav:
@makarandp0 

This PR fixes Firefox 79+ <-> Chrome interop issue in audio only Peer-to-Peer Participants. I have filed a [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1106157) for this issue.

### Changes

- [x] Implementation
- [x] Unit test
- [x] Verify interop fix locally
- [x] CHANGELOG.md entry

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -538,9 +538,7 @@ class PeerConnectionV2 extends StateMachine {
     }).then(() => {
       return this._peerConnection.createAnswer();
     }).then(answer => {
-      if (!isFirefox) {
-        answer = workaroundIssue8329(answer);
-      } else {
+      if (isFirefox) {
         // NOTE(mmalavalli): We work around Chromium bug 1106157 by disabling
         // RTX in Firefox 79+. For more details about the bug, please go here:
         // https://bugs.chromium.org/p/chromium/issues/detail?id=1106157
@@ -548,6 +546,8 @@ class PeerConnectionV2 extends StateMachine {
           sdp: disableRtx(answer.sdp),
           type: answer.type
         });
+      } else {
+        answer = workaroundIssue8329(answer);
       }
 
       // NOTE(mpatwardhan): Upcoming chrome versions are going to remove ssrc attributes
@@ -911,9 +911,7 @@ class PeerConnectionV2 extends StateMachine {
     }).catch(() => {
       throw new MediaClientLocalDescFailedError();
     }).then(offer => {
-      if (!isFirefox) {
-        offer = workaroundIssue8329(offer);
-      } else {
+      if (isFirefox) {
         // NOTE(mmalavalli): We work around Chromium bug 1106157 by disabling
         // RTX in Firefox 79+. For more details about the bug, please go here:
         // https://bugs.chromium.org/p/chromium/issues/detail?id=1106157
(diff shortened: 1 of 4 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -904,6 +913,14 @@ class PeerConnectionV2 extends StateMachine {
     }).then(offer => {
       if (!isFirefox) {
         offer = workaroundIssue8329(offer);
+      } else {
+        // NOTE(mmalavalli): We work around Chromium bug 1106157 by disabling
+        // RTX in Firefox 79+. For more details about the bug, please go here:
+        // https://bugs.chromium.org/p/chromium/issues/detail?id=1106157
+        offer = new this._RTCSessionDescription({
```
makarandp0, line 920: you want to apply this only for chrome right? so why not add isChrome check?

lib/signaling/v2/peerconnection.js
```diff
@@ -904,6 +913,14 @@ class PeerConnectionV2 extends StateMachine {
...
+          sdp: disableRtx(offer.sdp),
```
makarandp0, line 921: I would also wrap these changes around in a function like:

``` 
offer = workaround 1106157(offer);
```

I am hoping that this bug gets fixed and we can remove this workaround asap.

lib/util/sdp/index.js
```diff
@@ -510,15 +510,70 @@ function removeSSRCAttributes(sdp, ssrcAttributesToRemove) {
...
+    const rtxSSRCMatches = mediaSection.match(/a=ssrc-group:FID [0-9]+ ([0-9]+)/);
+    const rtxSSRC = rtxSSRCMatches && rtxSSRCMatches[1];
+
+    // Remove the following lines associated with the RTX payload types:
+    // 1. "a=fmtp:<rtxPt> apt=<pt>"
+    // 2. "a=rtpmap:<rtxPt> rtx/..."
+    // 3. "a=ssrc:<rtxSSRC> cname:..."
+    // 4. "a=ssrc-group:FID <SSRC> <rtxSSRC>"
+    const filterRegexes = [
+      /^a=fmtp:.+ apt=.+$/,
```
makarandp0, line 551: I am not super familiar with sdps, and wondering if these regex's could ever match unintended lines?
some of the expressions has "rtx" in them but some not.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @makarandp0 No, the fix is to disable RTX in Firefox 79+ SDPs. So, the fix is on the Firefox side, if you notice carefully.

lib/signaling/v2/peerconnection.js
manjeshbhargav: @makarandp0 I would like to make the workaround as transparent as possible so that a casual reader knows what the workaround does, in this case disabling RTX for Firefox SDPs.

lib/util/sdp/index.js
manjeshbhargav: @makarandp0 I'll explain the SDP lines not having "rtx":

1. `/^a=fmtp:.+ apt=.+$/` - Only RTX fmtp lines have the associated payload type `apt`.
2. `/^a=ssrc-group:.+$/` - This line maps the primary SSRC to its RTX SSRC.

So, the lines matching the above regexes can be removed.

lib/signaling/v2/peerconnection.js
makarandp0: ofcourse! its only for firefox :) I misread `!isFirefox` I think it would make it easier to switch this to
```
if (isFirefox) {
   // disable RTX for Firefox SDPs.
   offer = workaround 1106157(offer);
} else {
  offer = workaroundIssue8329(offer);
}
```

PR #1105, opened 2020-07-16
JSDK-2878 Alternate way of filtering parenthesized substrings in the user agent.

manjeshbhargav:
@makarandp0 

This fixes a bug in the logic for filtering out parenthesized substrings in the user agent, which resulted in `isSupported` returning `false` for Android Chrome in Motorola phones.

Ex. user agent: `Mozilla/5.0 (Linux; Android 6.0.1; Moto G (4)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Mobile Safari/537.36`

Now, substrings like `(Linux; Android 6.0.1; Moto G (4))` that contain nested parentheses are handled correctly.
 

**Contributing to Twilio**

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -11,6 +11,28 @@ const SUPPORTED_CHROME_BASED_BROWSERS = [
 ];
 
 /**
+ * Get the top level parenthesized substrings within a given string.
+ * Ex: "abc (def) gh(ij) (kl (mn)o)" => [""(def)", "(ij)", "(kl (mn)o)"]
+ * @param {string} string
+ * @returns {string[]}
+ */
+function getParenthesizedSubstrings(string) {
+  const openParenthesisPositions = [];
+  const substrings = [];
+  for (let i = 0; i < string.length; i++) {
+    if (string[i] === '(') {
+      openParenthesisPositions.push(i);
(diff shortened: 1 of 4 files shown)


lib/util/support.js
```diff
@@ -10,6 +10,28 @@ const SUPPORTED_CHROME_BASED_BROWSERS = [
...
+  const substrings = [];
+  for (let i = 0; i < string.length; i++) {
+    if (string[i] === '(') {
+      openParenthesisPositions.push(i);
+    } else if (string[i] === ')' && openParenthesisPositions.length > 0) {
+      const openParenthesisPosition = openParenthesisPositions.pop();
+      if (openParenthesisPositions.length === 0) {
+        substrings.push(string.substring(openParenthesisPosition, i + 1));
+      }
+    }
```
makarandp0, line 30: Looks like this code is forgiving if an extra close parenthesis is encountered. like  "(foo))(hello)"
Looks like it will return "(foo)" and "(hello)" which is not bad, but a commenting it in else case here would make it clear.

manjeshbhargav: @makarandp0 I will add a line in the documentation for the function. 👍

PR #1109, opened 2020-07-21
JSDK-2915 Detect silent video upon unmute and restart the LocalVideoTrack.

manjeshbhargav:
@makarandp0 

Reopening #1106 here because the JIRA ticket used in the previous PR was the wrong one (JSDK-2899). This PR updates the commit messages, branch name and CHANGELOG.md entry to use the correct JIRA ticket (JSDK-2915).

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/localvideotrack.js
+++ b/lib/media/track/localvideotrack.js
@@ -30,13 +30,31 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
    * @param {LocalTrackOptions} [options] - {@link LocalTrack} options
    */
   constructor(mediaStreamTrack, options) {
+    options = Object.assign({
+      workAroundSilentLocalVideo: guessBrowser() === 'safari'
+        && typeof document !== 'undefined'
+        && typeof document.createElement === 'function'
+    }, options);
+
     super(mediaStreamTrack, options);
+
+    Object.defineProperties(this, {
+      _workAroundSilentLocalVideo: {
+        value: options.workAroundSilentLocalVideo
+          ? workAroundSilentLocalVideo
+          : null
+      },
+      _workAroundSilentLocalVideoCleanup: {
+        value: null,
+        writable: true
+      }
+    });
 
     // NOTE(mmalavalli): In iOS Safari, we work around a bug where local video
     // MediaStreamTracks are silent (even though they are enabled, live and unmuted)
     // after accepting/rejecting a phone call.
-    if (guessBrowser() === 'safari') {
-      workAroundSilentLocalVideo(this);
+    if (this._workAroundSilentLocalVideo) {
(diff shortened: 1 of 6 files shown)


lib/media/track/localvideotrack.js
```diff
@@ -27,7 +30,32 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
+        && typeof document !== 'undefined'
+        && typeof document.createElement === 'function'
+    }, options);
+
     super(mediaStreamTrack, options);
+
+    Object.defineProperties(this, {
+      _workaroundSilentLocalVideo: {
+        value: options.workaroundSilentLocalVideo
+          ? workaroundSilentLocalVideo
```
makarandp0, line 44: I would keep it boolean than keeping function value. It's not writable, so don't gain much by keeping function value in this property.  By  keeping function value gives access to it to the api callers - which is not needed.

makarandp0: if you want to keep it function then maybe assign it a default value that does nothing...that way you can get rid of `if (this._workaroundSilentLocalVideo)` checks.

```
      _workaroundSilentLocalVideo: {
        value: options.workaroundSilentLocalVideo
          ? workaroundSilentLocalVideo
          : () => null
      },
      _workaroundSilentLocalVideoCleanup: {
        value: null,
        writable: true
      }
    });

this._workaroundSilentLocalVideoCleanup = this._workaroundSilentLocalVideo(this, document);

```

lib/media/track/localvideotrack.js
```diff
@@ -102,7 +130,18 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
   restart() {
-    return super.restart.apply(this, arguments);
+    if (this._workaroundSilentLocalVideoCleanup) {
+      this._workaroundSilentLocalVideoCleanup();
+      this._workaroundSilentLocalVideoCleanup = null;
+    }
+    const promise = super.restart.apply(this, arguments);
+
+    if (this._workaroundSilentLocalVideo) {
+      promise.finally(() => {
```
makarandp0, line 140: I think it's better to make caller wait for cleanup
```
      promise = promise.finally(() => {
        this._workaroundSilentLocalVideoCleanup = this._workaroundSilentLocalVideo(this, document);
      });
```

lib/media/track/localvideotrack.js
manjeshbhargav: @makarandp0 I had initially written it that way, but it would not be clear to the reader if that function is called without conditions. So, I think I will keep it as it is for now.

lib/media/track/localvideotrack.js
manjeshbhargav: @makarandp0 The problem with this is that any error thrown in earlier promise chains may not be passed to the Promise returned by `finally()`. Also, I don't think any application code needs to depend on the workaround function being called.

PR #1110, opened 2020-07-22
JSDK-2889 Playing unintentionally paused media elements if not backgrounded.

manjeshbhargav:
@makarandp0 This PR makes the following changes:

1. Move the element shimming logic from RemoteMediaTrack to MediaTrack, since local audio/video elements can also be inadvertently paused.
2. Playing unintentionally paused media elements only if not backgrounded. (We have already implemented playing unintentionally paused media elements after backgrounding and foregrounding).

Please take a look and let me know what you think.

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -1,6 +1,9 @@
 'use strict';
 
-const MediaStream = require('@twilio/webrtc').MediaStream;
+const { guessBrowser } = require('@twilio/webrtc/lib/util');
+const { MediaStream } = require('@twilio/webrtc');
+
+const { waitForEvent, waitForSometime } = require('../../util');
 const Track = require('./');
 
 /**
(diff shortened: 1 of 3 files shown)


lib/media/track/mediatrack.js
```diff
@@ -237,4 +271,82 @@ class MediaTrack extends Track {
...
+    },
+    unShim() {
+      el.pause = origPause;
+      el.play = origPlay;
+      if (onPause) {
+        el.removeEventListener('pause', onPause);
+      }
+    }
+  };
+}
```
makarandp0, line 350: This would look simpler if you change the default value for `onUnintentionallyPaused`

```suggestion
/**
 * Shim the pause() and play() methods of the given HTMLMediaElement so that
 * we can detect if it was paused unintentionally.
 * @param {HTMLMediaElement} el
 * @param {?function} [onUnintentionallyPaused=() => {}]
 * @returns {{pausedIntentionally: function, unShim: function}}
 */
function shimMediaElement(el, onUnintentionallyPaused = () => {}) {
  const origPause = el.pause;
  const origPlay = el.play;

  let pausedIntentionally = false;

  el.pause = () => {
    pausedIntentionally = true;
    return origPause.call(el);
  };

  el.play = () => {
    pausedIntentionally = false;
    return origPlay.call(el);
  };

  const onPause = () => {
    if (!pausedIntentionally) {
      onUnintentionallyPaused();
    }
  };

  el.addEventListener('pause', onPause);

  return {
    pausedIntentionally() {
      return pausedIntentionally;
    },
    unShim() {
      el.pause = origPause;
      el.play = origPlay;
        el.removeEventListener('pause', onPause);
    }
  };
}
```

manjeshbhargav: @makarandp0 Although this reads better, I don't want to add unnecessary pause event handlers.

PR #1113, opened 2020-07-23
JSDK-2928 Restart silent LocalAudioTrack on unmute.

manjeshbhargav:
@makarandp0 

This PR fixes the silent local audio issue (after playing youtube, etc) for iOS 13.6+, where the track does not go to ended state like before. Rather, it gets unmuted. Now, we detect silence after unmuted and reacquire the track.

This PR also merges the workaround for the local video silent issue (after a phone call) with the workaround for the local audio silent issue into a common workaround function.

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/localmediatrack.js
+++ b/lib/media/track/localmediatrack.js
@@ -6,9 +6,13 @@ const { guessBrowser } = require('@twilio/webrtc/lib/util');
 
 const { capitalize, defer, waitForSometime, waitForEvent } = require('../../util');
 const { typeErrors: { ILLEGAL_INVOKE } } = require('../../util/constants');
+const detectSilentAudio = require('../../util/detectsilentaudio');
+const detectSilentVideo = require('../../util/detectsilentvideo');
 const documentVisibilityMonitor = require('../../util/documentvisibilitymonitor.js');
 const gUMSilentTrackWorkaround = require('../../webaudio/workaround180748');
 const MediaTrackSender = require('./sender');
+
+const isSafari = guessBrowser() === 'safari';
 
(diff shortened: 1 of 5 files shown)


lib/media/track/localmediatrack.js
```diff
@@ -228,24 +244,62 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
...
+  }
+
   function shouldReacquireTrack() {
     const { _workaroundWebKitBug1208516Cleanup, isStopped, mediaStreamTrack: { muted } } = localMediaTrack;
     const isInadvertentlyStopped = isStopped && !!_workaroundWebKitBug1208516Cleanup;
-    return document.visibilityState === 'visible' && (muted || isInadvertentlyStopped) && !trackChangeInProgress;
+    return Promise.resolve().then(() => {
+      return document.visibilityState === 'visible'
+        && !trackChangeInProgress
+        && (muted || isInadvertentlyStopped || checkSilence());
```
makarandp0, line 283: `checkSilence` returns promise, but looks like here we are converting it to `boolean`. Shouldn't we wait for the checkSilence Promise?

manjeshbhargav: @makarandp0 As per our discussion, I will add a comment explaining the behavior.

makarandp0: @manjeshbhargav as you explained offline, I see how it works now. Suggest add some comments or update this to 
```
    return Promise.resolve().then(() => {
      if (document.visibilityState === 'visible'
        && !trackChangeInProgress
        && (muted || isInadvertentlyStopped)) {
          return checkSilence();
        } else {
          return false;
        }
```
to make it obvious

PR #1115, opened 2020-07-25
JSDK-2932 Work around WebKit Bug 213853

manjeshbhargav:
@makarandp0 
 
This PR works around this WebKit [bug](https://bugs.webkit.org/show_bug.cgi?id=213853) by playing inadvertently paused elements only after the local media is restarted after it has ended or unmuted.

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/localmediatrack.js
+++ b/lib/media/track/localmediatrack.js
@@ -9,6 +9,7 @@ const { typeErrors: { ILLEGAL_INVOKE } } = require('../../util/constants');
 const detectSilentAudio = require('../../util/detectsilentaudio');
 const detectSilentVideo = require('../../util/detectsilentvideo');
 const documentVisibilityMonitor = require('../../util/documentvisibilitymonitor.js');
+const localMediaRestartDeferreds = require('../../util/localmediarestartdeferreds');
 const gUMSilentTrackWorkaround = require('../../webaudio/workaround180748');
 const MediaTrackSender = require('./sender');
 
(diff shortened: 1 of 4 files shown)


lib/media/track/localmediatrack.js
```diff
@@ -321,22 +322,49 @@ function restartWhenInadvertentlyStopped(localMediaTrack) {
...
   }
 
+  function onMute() {
+    const { _log: log, kind } = localMediaTrack;
+    log.info('Muted');
+    log.debug('LocalMediaTrack:', localMediaTrack);
+
+    // NOTE(mmalavalli): When a LocalMediaTrack is muted without the app being
+    // backgrounded, and the inadvertently paused elements are played before it
+    // is restarted, it never gets unmuted due to the WebKit Bug 213853. Hence,
```
makarandp0, line 337: @manjeshbhargav - when the element gets muted because of phone call, does it ever get unmuted? what ensures that we will call `maybeRestart` ?

manjeshbhargav: @makarandp0 If you play the paused elements before the audio track is ended/unmuted, then the audio track does not get ended/unmuted, so `maybeRestart` is never called. So, we wait for unmuted/ended to fire, and then play the paused elements.

makarandp0: and I assume we do get `unmuted` in those cases? LGTM

manjeshbhargav: @makarandp0 We get the `ended` event.

PR #1131, opened 2020-08-07
JSDK-2893: Fixes incorrect answer revision in sync messages

makarandp0:
In a sequence where:

Step 1. Peer A => sends offer1, revision 1
Step 2. Peer B => returns answer1, revision 1

Step 3. Peer A => sends offer2, revision 2
Step 4. Peer B => sends sync with answer1 but marks it as revision 2

This is what I see in https://issues.corp.twilio.com/browse/JSDK-2893 looking at participant logs. And I see the possibility in code of this if a sync message was sent after step 3 before PeerB set the new answer ( answer2, revision 2) as the local description.

This change ensures that sync message will always set the revision of the local description that is set, not the one that is in process of setting. 

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 2.7.2 (In Progress)
 ===================
 
+Bug Fixes
+---------
+- Fixed a race condition that causes the sdk to return stale answer SDP when it loses connection momentarily, which leads to an error: 'Error setting remote description` (JSDK-2893)
+
 Changes
 -------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 Bug Fixes
 ---------
-- Fixed a race condition that causes the sdk to return stale answer SDP when it loses connection momentarily, which leads to an error: 'Error setting remote description` (JSDK-2893)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 2.7.2 (In Progress)
 ===================
 
+Bug Fixes
+---------
+- Fixed a race condition that causes the sdk to return stale answer SDP when it loses connection momentarily, which leads to an error: 'Error setting remote description` (JSDK-2893)
```
manjeshbhargav, line 10:
```suggestion

- Fixed a bug where a Participant in a large Group Room sometimes gets inadvertently disconnected with a [MediaServerRemoteDescFailedError](https://media.twiliocdn.com/sdk/js/video/releases/2.7.2/docs/MediaServerRemoteDescFailedError.html). (JSDK-2893)
```

lib/signaling/v2/peerconnection.js
```diff
@@ -1268,9 +1268,14 @@ class PeerConnectionV2 extends StateMachine {
     if (!this._localDescription) {
       return null;
     }
+
+    // return the last revision that is set in the local description.
+    // not the one that we are in processing. they would be different if this function gets called for
+    // sending 'sync' message, while we were processing a new offer (but not yet sent the answer for it.)
```
manjeshbhargav, line 1274:
```suggestion
    // NOTE(mpatwardhan): If the most recent local description is an answer, and this method is called for
    // sending a "sync" message while the next remote offer is being processed, we need to send the most
    // recent stable description revision instead of the current description revision, which is supposed
    // to be for the next local answer.
```

PR #1157, opened 2020-08-28
JSDK-2959 : add testPreflight api

makarandp0:
this adds an api to test the twilio connectivity, and generate report on success. 
api definition here: https://code.hq.twilio.com/client/video-sdk-api/pull/71

```js
var Video = require('twilio-video');
var publisherToken = getAccessToken('alice');
var subscriberToken = getAccessToken('bob');
var preflightTest = Video.testPreflight(publisherToken, subscriberToken);
preflightTest.on('completed', function(report) {
  console.log("Test completed in " + report.testTiming.duration + " milliseconds.");
  console.log(" It took " + report.networkTiming.connect.duration + " milliseconds to connect");
  console.log(" It took " + report.networkTiming.media.duration + " milliseconds to receive media");
  console.log(" Your network score was: " + report.stats.networkQuality); // only available for group rooms.
});
preflightTest.on('failed', function(error) {
  console.log("Test failed:" + error);
});
preflightTest.on('progress', function(progressState) {
  console.log(progressState);
});
```

TODO
- [x] gather local ice candidates
- [x] check about incoming/outgoing bitrate stats

**Contributing to Twilio**


lib/preflight.js
```diff
@@ -0,0 +1,86 @@
...
+ * @example
+ * var Video = require('twilio-video');
+ * var publisherToken = getAccessToken('alice');
+ * var subscriberToken = getAccessToken('bob');
+ * var preflightTest = Video.testPreflight(publisherToken, subscriberToken);
+ * preflightTest.on('completed', function(report) {
+ *   console.log("Test completed in " + report.testTiming.duration + " milliseconds.");
+ *   console.log(" It took " + report.networkTiming.connect.duration + " milliseconds to connect");
+ *   console.log(" It took " + report.networkTiming.media.duration + " milliseconds to receive media");
+ *   console.log(" Your network score was: " + report.stats.networkQuality); // only available for group rooms.
```
timmydoza, line 27: How would a developer choose what room type to use? Does the preflight test use the room name that is contained in the token? Or does this test create a new room using the default room type from the Twilio console. I'm wondering if it would be useful if the docs went into a little more detail about this.

timmydoza:
Looks good! I'm able to run the tests. Just had one question about some of the values from the test report.

`stats.incomingBitrate.min` is always 0. Also the `average`, `min`, and `max` values for `outgoingBitrate` are always 300,000. It seems odd to me that these values never change no matter how many times I try the test. What do you think?

lib/preflight.js
makarandp0: Good point, Yes I will add more details in the js docs about this. 
The test uses the token supplied, so it will  use  the room mentioned in the token, if it does not exist it will be created using the default room type from console.

I think developers typically would not want to specify room-type, because there app either uses group room or p2p room - whichever room type the app uses should be set as default room type in console.

But if they do want to call it with different room-type they should create room before calling the api - as our integration test does.

makarandp0:
@timmydoza, Thank you for trying this out. I have fixed the issue with stats.incomingBitrate.min being 0 ( it was caused by a bug where if the incoming bit rate was not calculated then it was considered as 0)

I have also fixed an issue with outgoingBitrate - now it should reflect the bitrate taken from publisher (instead of subscriber). 

I have  also update the doc entry to specify mechanics of which room is used.

lib/preflight/preflighttest.js
```diff
@@ -0,0 +1,483 @@
...
+              collectedStats.rtt.push(activeIceCandidatePair.totalRoundTripTime * 1000);
+            }
+            if (typeof activeIceCandidatePair.availableIncomingBitrate === 'number') {
+              collectedStats.incomingBitrate.push(activeIceCandidatePair.availableIncomingBitrate);
+            }
+
+            if (!collectedStats.selectedIceCandidatePairStats) {
+              collectedStats.selectedIceCandidatePairStats = {
+                localCandidate: activeIceCandidatePair.localCandidate,
+                remoteCandidate: activeIceCandidatePair.remoteCandidate
```
charliesantos, line 179: `activeIceCandidatePair.remoteCandidate` and `activeIceCandidatePair.localCandidate` is of type [RTCIceCandidateStats](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateStats) and not [RTCIceCandidate](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate) correct?

lib/preflight/preflighttest.js
```diff
@@ -0,0 +1,483 @@
...
+        jitter: [],
+        rtt: [],
+        outgoingBitrate: [],
+        incomingBitrate: [],
+        packetLoss: [],
+        selectedIceCandidatePairStats: null,
+        iceCandidateStats: [],
+      };
+    }
+    return waitForSometime(STAT_INTERVAL).then(() => {
```
charliesantos, line 243: This is a pause correct? Why do we need this? Will this affect time measurements?

lib/preflight/preflighttest.js
```diff
@@ -0,0 +1,483 @@
...
+  // @returns {Array<number>}
+  function collectNetworkQualityForDuration(duration) {
+    const networkQuality = [];
+    const localParticipant = subscriberRoom.localParticipant;
+    if (localParticipant.networkQualityLevel) {
+      networkQuality.push(localParticipant.networkQualityLevel);
+    }
+    const networkQualityCallback = () => networkQuality.push(localParticipant.networkQualityLevel);
+    localParticipant.addListener('networkQualityLevelChanged', networkQualityCallback);
+    return waitForSometime(duration).then(() => {
```
charliesantos, line 265: Same here. Why do we need this? Will this affect time measurements?

charliesantos (changes requested):
Looks good Mak. Just left a few comments/questions.

lib/stats/makestat.js
```diff
@@ -0,0 +1,23 @@
...
+  if (values.length >= 1) {
+    let sum = 0;
+    let min = values[0];
+    let max = values[0];
+    values.forEach(val => {
+      min = Math.min(min, val);
+      max = Math.max(max, val);
+      sum += val;
+    });
+    const average = sum / values.length;
```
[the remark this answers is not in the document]
makarandp0: re: precision - I think we should let developer's decide the precision they want

lib/stats/makestat.js
```diff
@@ -0,0 +1,23 @@
...
+ * @returns {{min: number, max: number: average: number}|null}
+ */
+function makeStat(values) {
+  if (values.length >= 1) {
+    let sum = 0;
+    let min = values[0];
+    let max = values[0];
+    values.forEach(val => {
+      min = Math.min(min, val);
+      max = Math.max(max, val);
```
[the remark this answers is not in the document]
makarandp0: Thanks @charliesantos, it does read better!

lib/stats/makestat.js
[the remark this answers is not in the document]
charliesantos: Maybe a default precision of let's say `5` for now? We just don't want to see something like `323.45892634827463` in the report :)

lib/preflight/preflighttest.js
makarandp0: yes they are of type `RTCIceCandidateStats`

lib/preflight/preflighttest.js
makarandp0: this ensures that we sample stats at 1000 ms. This is similar to doing something like:
```
setInterval(() => collectStats(), 1000)
```

lib/preflight/preflighttest.js
makarandp0: this ensures that function returns after collecting network quality score for `duration`. we watch for the  `networkQualityLevelChanged` during that time.

lib/stats/makestat.js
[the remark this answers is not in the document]
makarandp0: I think developers would want precision of 2 (than 5), so I would rather let them make the decision. Also [toPrecision](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision)() converts a number to string :)

lib/preflight/preflighttest.js
```diff
@@ -0,0 +1,483 @@
...
+        return participantsConnected(publisherRoom);
+      });
+    }).then(() => {
+      return executePreflightStep('subscriber sees publisher', () => {
+        return participantsConnected(publisherRoom);
+      });
+    }).then(() => {
+      return executePreflightStep('validate rooms', () => {
+        if (subscriberRoom.sid !== publisherRoom.sid) {
+          throw new Error('wrong room tokens. Ensure that both token join to the same room');
```
timmydoza, line 349: How about `'Incorrect room tokens. Ensure that both tokens connect to the same room.'`?

lib/preflight/preflighttest.js
```diff
@@ -0,0 +1,483 @@
...
+        }
+        if (subscriberRoom.participants.size !== 1) {
+          throw new Error('unexpected participant found in the room');
```
timmydoza, line 352: How about `'Unexpected participant found in the room. Ensure that the room has no participants in it.'`?

CHANGELOG.md
```diff
@@ -2,6 +2,43 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
 
+2.7.3 (In Progress)
+===================
+
+New Features
+------------
+
+- Added a new api `testPreflight`. This api helps test the connectivity with twilio-servers for the video rooms. It connects two participants to the video room using supplied tokens. It publishes synthetic audio and video tracks from one participant and ensures that other participant receives media on those tracks. After successfully verifying the room operations, it generates a report containing statistics about the connection.
+
+Note: the tokens used for the connection must specify the room to be used for test. This must be a test room and must not to be used fo regular operations.
```
timmydoza, line 13: Typo:  `to be used for regular operations.`

lib/preflight/preflight.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * You may pass these options to {@link module:twilio-video.testPreflight} in order to override the
+ * default behavior.
+ * @typedef {object} PreflightOptions
+ * @property {string} [region='gll'] - Preferred signaling region; By default, you will be connected to the
+ *   nearest signaling server determined by latency based routing. Setting a value other
+ *   than <code style="padding:0 0">gll</code> bypasses routing and guarantees that signaling traffic will be
+ *   terminated in the region that you prefer. Please refer to this <a href="https://www.twilio.com/docs/video/ip-address-whitelisting#signaling-communication" target="_blank">table</a>
+ *   for the list of supported signaling regions.
+ * @property {number} [duration=5000] - number of milliseconds to run test for.
```
timmydoza, line 73: Looks like the default duration is 10000 ms.

lib/stats/makestat.js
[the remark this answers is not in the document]
charliesantos: >so I would rather let them make the decision
Ok. In voice, we use 5 as default and we let devs to use to if they like. This way, @timmydoza doesn't have to adjust it when he consumes it. But this is ok for now for signal and we can talk more about this later on.

>Also toPrecision() converts a number to string :)
I was thinking we it'll be using `Number(average.toPrecision(5))` :)

timmydoza:
Looks good! I was curious about the synthetic media streams, so I joined a room that had a preflight test in progress. I could hear the oscillator, but I could not view the video. It looks like a video track is successfully being published, but nothing is being displayed. Should I be able to see the random colors from the synthetic video track?

makarandp0:
@timmydoza, @charliesantos - Thank you for taking a look. I have incorporated your feedback now. Please take another look when you get a chance.

@timmydoza I am not sure why you did not see video track - I tested the code individually and I do see video track when rendered on video element both locally and from the remote.

charliesantos (approved):

PR #1069, opened 2020-06-25
pre-commit hook to avoid common mistakes

makarandp0:
Yesterday I commited a change which marked a test suite  `.only` - I ended up cutting a rc with it.
This change reverts  the `.only` and adds a pre-commit hook to avoid such errors in future. The pre-commit hook is not on by default - instructions to enable it are added to readme.

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #1184, opened 2020-09-22
JSDK-2985 - MaxEventListeners Warning Fix

PikaJoyce:
JIRA Ticket: [JSDK-2985](https://issues.corp.twilio.com/browse/JSDK-2985)

This is a WIP PR for JSDK-2985. This fixes the max event listeners by unpublishing LocalParticipant's publications.

IN PROGRESS:
- [x] Write Test

**Contributing to Twilio**

### DIFF
--- a/lib/room.js
+++ b/lib/room.js
@@ -528,7 +528,9 @@ function handleSignalingEvents(room, signaling) {
         room.participants.forEach(participant => {
           participant._unsubscribeTracks();
         });
-        room.emit(state, room, error);
+        room.localParticipant.tracks.forEach(publication => {
+          publication.unpublish();
+        });
         signaling.removeListener('stateChanged', stateChanged);
         break;
       case 'reconnecting':
--- a/lib/room.js
+++ b/lib/room.js
@@ -528,6 +528,7 @@ function handleSignalingEvents(room, signaling) {
         room.participants.forEach(participant => {
           participant._unsubscribeTracks();
         });
+        room.emit(state, room, error);
         room.localParticipant.tracks.forEach(publication => {
           publication.unpublish();
         });
--- a/lib/room.js
+++ b/lib/room.js
@@ -12,7 +12,7 @@ let nInstances = 0;
  * {@link RemoteParticipant}s sharing {@link AudioTrack}s and
  * {@link VideoTrack}s.
  * <br><br>
- * You can connect to a {@link Room} by calling {@link connect}.
+ * You can connect to a {@link Room} by calling {@link module:twilio-video.connect}.
  * @extends EventEmitter
  * @property {?RemoteParticipant} dominantSpeaker - The Dominant Speaker in the
  *   {@link Room}, if any
--- a/test/integration/spec/localparticipant.js
+++ b/test/integration/spec/localparticipant.js
@@ -1394,4 +1394,48 @@ describe('LocalParticipant', function() {
       assert(!error);
     });
   });
+
+  describe.only('JSDK-2985 - Event Listeners cleaned up after leaving room', () => {
+    let room;
+    let error;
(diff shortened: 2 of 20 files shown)


lib/room.js
```diff
@@ -529,6 +529,9 @@ function handleSignalingEvents(room, signaling) {
           participant._unsubscribeTracks();
         });
         room.emit(state, room, error);
+        room.localParticipant.tracks.forEach(publication => {
+          publication.unpublish();
```
makarandp0, line 533: @PikaJoyce - I assume you are calling this `unpublish` for some side effect - of unregistering listener right? Can you clarify which event does this clear.

test/integration/spec/localparticipant.js
```diff
@@ -1394,4 +1394,48 @@ describe('LocalParticipant', function() {
       assert(!error);
     });
   });
+
+  describe.only('JSDK-2985 - Event Listeners cleaned up after leaving room', () => {
```
makarandp0, line 1398: Please remove `.only`  - I have a precomit hooks that would help avoid this. @manjeshbhargav - is it okay to merge https://github.com/twilio/twilio-video.js/pull/1069 ?

test/integration/spec/localparticipant.js
```diff
@@ -1394,4 +1394,48 @@ describe('LocalParticipant', function() {
...
+        const token = getToken(randomName());
+        const options = Object.assign({ audio: true, fake: true }, defaults);
+        // eslint-disable-next-line no-await-in-loop
+        room = await setupRoom(token, options);
+        if (room) {
+          room.disconnect();
+        }
+        events.forEach(event => {
+          room.localParticipant.tracks.forEach(publication => {
+            listenerCount += publication.track.listenerCount(event);
```
makarandp0, line 1419: Looks like a good test - does this test fail w/o your fix ? How about adding multiple connect/disconnect in this test ?

makarandp0 (changes requested):
@PikaJoyce,  This is great start, Can you add some more explanation on the events that were leaking and how this fixes them?

test/integration/spec/localparticipant.js
manjeshbhargav: @makarandp0 👍  as long as it's not enabled by default.

manjeshbhargav, line 1398:
```suggestion
  describe('JSDK-2985 - Clean up LocalTrack event listeners added by the LocalTrackPublication after leaving room', () => {
```

manjeshbhargav (changes requested):
@PikaJoyce Please make the following changes:
- [ ] Open the PR against `develop`, not `master`.
- [ ] Address PR feedback.
- [ ] Add CHANGELOG.md entry for the bug fix.

test/integration/spec/localparticipant.js
PikaJoyce: Yes! I took out the extra connects/disconnects because I realized I only wanted to see if the listeners were being cleaned up after we leave the room. Without the fix, the test fails and catches 7 listeners. With the fix, it catches 0 listeners and the test passes.

lib/room.js
PikaJoyce: Calling unpublish on all of the tracks cleans up all of the listeners that have been added. 

Since I check for the listeners on these events `enabled, disabled, started, stopped, message`, they all get cleared when the publications are unpublished on `disconnected`

manjeshbhargav (changes requested):
@PikaJoyce Please update your branch with changes from `develop` before opening the PR. Otherwise we will see all these unnecessary changes in the PR.

manjeshbhargav (approved):

PR #1186, opened 2020-09-22
JSDK 2998 - preflight safari audiocontext

makarandp0:
preflight uses web audio apis - On safari we do not get any bytes on audio tracks created using these apis if there was a big delay between user action and creating audio context.  here is a demo of this issue (https://makarandp0.github.io/safariAudio/)

This change 1) adds doc to clarify this 2) small code change to ensure we acquire tracks on the same tick

- [ ] todo: fix unit test

**Contributing to Twilio**


makarandp0:
the changelog entry for this issue is included in https://github.com/twilio/twilio-video.js/pull/1200

PR #1199, opened 2020-09-25
add preflight field to ice and connect messages

makarandp0:
This adds a new boolean field `preflight` for `ice` and `connect` messages. This field is set to true from preflight tests.

related rsp pr: https://code.hq.twilio.com/client/room-signaling-protocol/pull/67
**Contributing to Twilio**


idelgado (approved):
Looks good! I would confirm that you see the property for both messages logged in Kibana on the VSG role.

makarandp0:
@idelgado - I looked at Kibana but did not see this property - I suspect VSG does not log unknown properties. But I did verify that client sends the new field in both connect and Ice messages. I will merge this now.

PR #1216, opened 2020-10-06
JSDK-2955 - RemoteTrackPublications and LocalTrackPublications definitions

PikaJoyce:
JIRA Ticket: [JSDK-2955](https://issues.corp.twilio.com/browse/JSDK-2955)

This PR includes all of the Track Publications definitions.

WIP:
- [x] Write some "tests"

**Contributing to Twilio**

### DIFF
--- a/tsdef/AudioTrack.d.ts
+++ b/tsdef/AudioTrack.d.ts
@@ -9,7 +9,7 @@ export class AudioTrack extends Track {
   attach(element?: HTMLMediaElement | string): HTMLMediaElement;
   detach(element?: HTMLMediaElement | string): HTMLMediaElement[];
 
-  on(event: 'disabled', listener: (track: this) => void): this;
-  on(event: 'enabled', listener: (track: this) => void): this;
-  on(event: 'started', listener: (track: this) => void): this;
+  on(event: 'disabled', listener: (track: this) => any): this;
+  on(event: 'enabled', listener: (track: this) => any): this;
+  on(event: 'started', listener: (track: this) => any): this;
 }
--- a/tsdef/AudioTrack.d.ts
+++ b/tsdef/AudioTrack.d.ts
@@ -9,7 +9,7 @@ export class AudioTrack extends Track {
   attach(element?: HTMLMediaElement | string): HTMLMediaElement;
   detach(element?: HTMLMediaElement | string): HTMLMediaElement[];
 
-  on(event: 'disabled', listener: (track: this) => any): this;
-  on(event: 'enabled', listener: (track: this) => any): this;
-  on(event: 'started', listener: (track: this) => any): this;
+  on(event: 'disabled', listener: (track: this) => void): this;
+  on(event: 'enabled', listener: (track: this) => void): this;
+  on(event: 'started', listener: (track: this) => void): this;
 }
--- a/tsdef/LocalAudioTrackPublication.d.ts
+++ b/tsdef/LocalAudioTrackPublication.d.ts
@@ -1,0 +1,12 @@
+import { LocalTrackPublication } from './LocalTrackPublication';
+
+export class LocalAudioTrackPublication extends LocalTrackPublication {
+  kind: 'audio';
+  track: LocalAudioTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
+}
--- a/tsdef/LocalAudioTrackPublication.d.ts
+++ b/tsdef/LocalAudioTrackPublication.d.ts
@@ -1,12 +1,7 @@
+import { LocalAudioTrack } from './LocalAudioTrack';
 import { LocalTrackPublication } from './LocalTrackPublication';
 
 export class LocalAudioTrackPublication extends LocalTrackPublication {
   kind: 'audio';
   track: LocalAudioTrack;
-
-  setPriority(priority: Track.Priority): this;
-  unpublish(): this;
-
-  on(event: 'trackDisabled', listener: () => any): this;
-  on(event: 'trackEnabled', listener: () => any): this;
 }
--- a/tsdef/LocalDataTrackPublication.d.ts
+++ b/tsdef/LocalDataTrackPublication.d.ts
@@ -1,12 +1,7 @@
+import { LocalDataTrack } from './LocalDataTrack';
 import { LocalTrackPublication } from './LocalTrackPublication';
 
 export class LocalDataTrackPublication extends LocalTrackPublication {
   kind: 'data';
   track: LocalDataTrack;
-
-  setPriority(priority: Track.Priority): this;
-  unpublish(): this;
-
-  on(event: 'trackDisabled', listener: () => any): this;
-  on(event: 'trackEnabled', listener: () => any): this;
 }
--- a/tsdef/LocalDataTrackPublication.d.ts
+++ b/tsdef/LocalDataTrackPublication.d.ts
@@ -1,0 +1,12 @@
+import { LocalTrackPublication } from './LocalTrackPublication';
+
+export class LocalDataTrackPublication extends LocalTrackPublication {
+  kind: 'data';
+  track: LocalDataTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
+}
--- a/tsdef/LocalTrackPublication.d.ts
+++ b/tsdef/LocalTrackPublication.d.ts
@@ -9,6 +9,6 @@ export class LocalTrackPublication extends TrackPublication {
   setPriority(priority: Track.Priority): this;
   unpublish(): this;
 
-  on(event: 'trackDisabled', listener: () => any): this;
-  on(event: 'trackEnabled', listener: () => any): this;
+  on(event: 'trackDisabled', listener: () => void): this;
+  on(event: 'trackEnabled', listener: () => void): this;
 }
--- a/tsdef/LocalTrackPublication.d.ts
+++ b/tsdef/LocalTrackPublication.d.ts
@@ -1,0 +1,15 @@
+import { Track } from './Track';
+import { TrackPublication } from './TrackPublication';
+
+export class LocalTrackPublication extends TrackPublication {
+  isTrackEnabled: boolean;
+  kind: Track.Kind;
+  priority: Track.Priority;
+  track: LocalTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
+}
--- a/tsdef/LocalTrackPublication.d.ts
+++ b/tsdef/LocalTrackPublication.d.ts
@@ -1,9 +1,8 @@
+import { LocalTrack } from './types';
 import { Track } from './Track';
 import { TrackPublication } from './TrackPublication';
 
 export class LocalTrackPublication extends TrackPublication {
-  isTrackEnabled: boolean;
-  kind: Track.Kind;
   priority: Track.Priority;
   track: LocalTrack;
 
--- a/tsdef/LocalVideoTrackPublication.d.ts
+++ b/tsdef/LocalVideoTrackPublication.d.ts
@@ -1,12 +1,7 @@
 import { LocalTrackPublication } from './LocalTrackPublication';
+import { LocalVideoTrack } from './LocalVideoTrack';
 
 export class LocalVideoTrackPublication extends LocalTrackPublication {
-  kind: 'data';
+  kind: 'video';
   track: LocalVideoTrack;
-
-  setPriority(priority: Track.Priority): this;
-  unpublish(): this;
-
-  on(event: 'trackDisabled', listener: () => any): this;
-  on(event: 'trackEnabled', listener: () => any): this;
 }
--- a/tsdef/LocalVideoTrackPublication.d.ts
+++ b/tsdef/LocalVideoTrackPublication.d.ts
@@ -1,0 +1,12 @@
+import { LocalTrackPublication } from './LocalTrackPublication';
+
+export class LocalVideoTrackPublication extends LocalTrackPublication {
+  kind: 'data';
+  track: LocalVideoTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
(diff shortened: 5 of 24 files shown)


tsdef/RemoteTrackPublication.d.ts
```diff
@@ -0,0 +1,19 @@
+import { Track } from './Track';
```
manjeshbhargav, line 1: Please export
```ts
export type RemoteTrack = RemoteAudioTrack | RemoteDataTrack | RemoteVideoTrack;
```
from a new file called `types.d.ts` and import it here:
```suggestion
import { Track } from './Track';
import { RemoteTrack } from './types';
```
Without this, references to `RemoteTrack` would point to an undefined type.

tsdef/RemoteTrackPublication.d.ts
```diff
@@ -0,0 +1,19 @@
...
+import { TrackPublication } from './TrackPublication';
+
+export class RemoteTrackPublication extends TrackPublication {
+  isSubscribed: boolean;
+  isTrackEnabled: boolean;
+  kind: Track.Kind;
```
manjeshbhargav, line 7: Is `Track.Kind` defined  in `Track.d.ts`?

tsdef/LocalDataTrackPublication.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class LocalDataTrackPublication extends LocalTrackPublication {
+  kind: 'data';
+  track: LocalDataTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
```
manjeshbhargav, line 11:
```suggestion
```

tsdef/TrackPublication.d.ts
```diff
@@ -0,0 +1,13 @@
+import { EventEmitter } from 'events';
+import { Track } from './Track';
+
+export class TrackPublication extends EventEmitter {
+  trackName: string;
+  trackSid: Track.SID;
```
manjeshbhargav, line 6: Is `Track.SID` defined in `Track.d.ts`?

tsdef/TrackPublication.d.ts
```diff
@@ -0,0 +1,13 @@
+import { EventEmitter } from 'events';
+import { Track } from './Track';
+
+export class TrackPublication extends EventEmitter {
+  trackName: string;
```
manjeshbhargav, line 5:
```suggestion
  isEnabled: boolean;
  kind: Track.kind;
  trackName: string;
```

tsdef/RemoteTrackPublication.d.ts
manjeshbhargav, line 7:
```suggestion
```

tsdef/RemoteDataTrack.d.ts
```diff
@@ -9,4 +9,10 @@ export class RemoteDataTrack extends Track {
   ordered: boolean;
   reliable: boolean;
   sid: Track.SID;
+
+  setPriority(priority: Track.Priority): RemoteDataTrack;
```
manjeshbhargav, line 13:
```suggestion
  setPriority(priority: Track.Priority | null): RemoteDataTrack;
```

tsdef/LocalVideoTrackPublication.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class LocalVideoTrackPublication extends LocalTrackPublication {
+  kind: 'data';
+  track: LocalVideoTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
```
manjeshbhargav, line 11:
```suggestion
```

tsdef/LocalTrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
+import { Track } from './Track';
+import { TrackPublication } from './TrackPublication';
+
+export class LocalTrackPublication extends TrackPublication {
+  isTrackEnabled: boolean;
+  kind: Track.Kind;
```
manjeshbhargav, line 6:
```suggestion
```

tsdef/LocalTrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  priority: Track.Priority;
+  track: LocalTrack;
```
manjeshbhargav, line 8: Please export
```ts
export type LocalTrack = LocalAudioTrack | LocalDataTrack | LocalVideoTrack;
```
from a new file called `types.d.ts` and import it here. Otherwise `LocalTrack` will refer to an undefined type.

tsdef/LocalAudioTrackPublication.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class LocalAudioTrackPublication extends LocalTrackPublication {
+  kind: 'audio';
+  track: LocalAudioTrack;
+
+  setPriority(priority: Track.Priority): this;
+  unpublish(): this;
+
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
```
manjeshbhargav, line 11:
```suggestion
```

manjeshbhargav (changes requested):

tsdef/RemoteTrackPublication.d.ts
PikaJoyce: Yes. There is a `namespace Track` that defines `Kind`. The actual class `Track` is `Track.Kind` from the namespace.

tsdef/AudioTrack.d.ts
```diff
@@ -9,7 +9,7 @@ export class AudioTrack extends Track {
   attach(element?: HTMLMediaElement | string): HTMLMediaElement;
   detach(element?: HTMLMediaElement | string): HTMLMediaElement[];
 
-  on(event: 'disabled', listener: (track: this) => void): this;
-  on(event: 'enabled', listener: (track: this) => void): this;
-  on(event: 'started', listener: (track: this) => void): this;
+  on(event: 'disabled', listener: (track: this) => any): this;
```
timmydoza, line 12: Just curious, why is `void` being changed to `any`? This is nit-picky, but I think `void` is correct. This is consistent with how event listeners are defined in the node.js type definitions.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v12/events.d.ts#L33

PikaJoyce: This was updated to `void` to be consistent with how RTC Diagnostics implements their event listeners. @mhuynh5757 pointed out to me that event listeners should not be opinionated. 

https://github.com/twilio/rtc-diagnostics/blob/master/lib/AudioInputTest.ts#L90

timmydoza: I see. I don't think it makes any real difference whether we use `any` or `void`, but I'd personally lean towards keeping things consistent with the types used in the `EventEmitter` class since these classes are inheriting from it.

Also, the TS Docs do say to use `void` instead of `any`.

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#return-types-of-callbacks

tsdef/RemoteAudioTrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
...
+export class RemoteAudioTrackPublication extends RemoteTrackPublication {
+  kind: 'audio';
+  track: RemoteAudioTrack | null;
+
+  on(event: 'subscribed', listener: (track: RemoteAudioTrack) => any): this;
+  on(event: 'subscriptionFailed', listener: (error: TwilioError) => any): this;
+  on(event: 'trackDisabled', listener: () => any): this;
+  on(event: 'trackEnabled', listener: () => any): this;
+  on(event: 'unsubscribed', listener: (track: RemoteAudioTrack) => any): this;
+  on(event: string, listener: (...args: any[]) => any): this;
```
timmydoza, line 14: Adding this line does get rid of the error: `Property 'on' in type 'RemoteAudioTrackPublication' is not assignable to the same property in base type 'RemoteTrackPublication'.`. It's a really simple approach, but in my eyes it has one drawback.

This approach works well when the developer spells the event name correctly. But when the event name is not spelled correctly, TS does not throw an error:

image

It would be nice if TS could alert the user when an incorrect event name is used. Like this: 

image

This is how it works with the diagnostics SDK (although an enum is used for the event name instead of a string). I think it would be great if we could implement something similar here too. I'm thinking about the best way to do this - it's a little tricky. I'd love to hear what others think.

tsdef/AudioTrack.d.ts
PikaJoyce: @mhuynh5757 @manjeshbhargav 

@timmydoza I hear you. I'm also open to more input and discussion as well on which to use. I think if we go with `void`, then should we not be doing that for RTC Diagnostics as well? I'd like to aim for consistency, I think.

mhuynh5757: Thanks for bringing that up @timmydoza , didn't know that TS prefers `void` over `any`. Makes sense 👍

tsdef/RemoteAudioTrackPublication.d.ts
mhuynh5757: Yeah it would be nice but Diagnostics SDK can only do it because the inheritance is direct, i.e. we only inherit from `EventEmitter` once. TS is narrowing the type of `event` to the specific string, like `'trackSubscribed'`. Then that class will only accept `event: 'trackSubscribed'` which makes sense, but then when we inherit again then we can't define something like `event: 'trackAdded'` because `'trackAdded'` is not of type `'trackSubscribed'`.

timmydoza: Yeah it's much easier with direct inheritance. I think I may have found a solution though:

This is the end result for the user. An object like `RemoteVideoTrackPublication` (which indirectly inherits from `EventEmitter`) lists all of its events and all of its inherited events:

image

It also knows the type of listener that's needed for a particular event:
image


And it provides useful error messages when a non-existent event name is used:

image



To achieve this, we shouldn't use the node.js types for `EventEmitter` because it has no type-checking. Instead we should use something like this: https://github.com/binier/tiny-typed-emitter

This library provides type definitions for a node.js `EventEmitter` class that is type-aware. There's no JS in that library. It's a small library (and MIT licensed), so I just copied [this file](https://github.com/binier/tiny-typed-emitter/blob/master/lib/index.d.ts) into the twilio-video repo in order to get the following examples to work.

This is what TrackPublication.d.ts looks like:

```
import { ListenerSignature, TypedEventEmitter } from './TypedEventEmitter';
import { Track } from './Track';

export interface TrackPublicationEvents {
  trackDisabled: () => void;
  trackEnabled: () => void;
}

export class TrackPublication<T extends ListenerSignature<T>> extends TypedEventEmitter<TrackPublicationEvents & T> {
  isEnabled: boolean;
  kind: Track.Kind;
  trackName: string;
  trackSid: Track.SID;

  toJSON(): string;
  toString(): string;
}
```
A few things are happening here:
- The event names and their respective listeners are defined on the `TrackPublicationEvents` interface (I like how clean this is). 
- The `TrackPublication` class accepts a type variable `T` which extends `ListenerSignature<T>` 
- The `TrackPublication` class extends `TypedEventEmitter<TrackPublicationEvents & T>`. Note: here it is essential that the intersection type operator (`&`) is used to combine the `TrackPublicationEvents` interface with the type variable `T`.

This enables another class to extend TrackPublication and pass in it's own interface of event names and listeners as a type variable. Let's look at `RemoteTrackPublication.d.ts`:

```
import { RemoteTrack } from './types';
import { Track } from './Track';
import { TrackPublication } from './TrackPublication';
import { TwilioError } from './TwilioError';
import { ListenerSignature } from './TypedEventEmitter';

export interface RemoteTrackPublicationEvents {
  publishPriorityChanged: (priority: Track.Priority) => void;
  subscribed: (track: RemoteTrack) => void;
  subscriptionFailed: (error: TwilioError) => void;
  trackSwitchedOff: (track: RemoteTrack) => void;
  trackSwitchedOn: (track: RemoteTrack) => void;
  unsubscribed: (track: RemoteTrack) => void;
}

export class RemoteTrackPublication<T extends ListenerSignature<T>> extends TrackPublication<RemoteTrackPublicationEvents & T> {
  isSubscribed: boolean;
  publishPriority: Track.Priority;
  track: RemoteTrack | null;
}
```
Same thing here, but this class extends `TrackPublication` instead of `TypedEventEmitter`.  Again, the use of the type intersection operator here is important.

Finally, `RemoteVideoTrackPublication.d.ts`:
```
import { RemoteTrackPublication } from './RemoteTrackPublication';
import { RemoteVideoTrack } from './RemoteVideoTrack';
import { TwilioError } from './TwilioError';

export interface RemoteVideoTrackPublicationEvents {
  subscribed: (track: RemoteVideoTrack) => void;
  subscriptionFailed: (error: TwilioError) => void;
  trackSwitchedOff: (track: RemoteVideoTrack) => void;
  trackSwitchedOn: (track: RemoteVideoTrack) => void;
  unsubscribed: (track: RemoteVideoTrack) => void;
}

export class RemoteVideoTrackPublication extends RemoteTrackPublication<RemoteVideoTrackPublicationEvents> {
  kind: 'video';
  track: RemoteVideoTrack | null;
}
```

Since no other classes need to extend `RemoteVideoTrackPublication`, we don't need to set it up to receive a type variable.

What do you all think?

PikaJoyce: I think that this is a really good idea. I would be more than happy to add this. I think this will be much more helpful to users in the future.

tsdef/TrackPublication.d.ts
manjeshbhargav: @PikaJoyce Please confirm this.

manjeshbhargav (changes requested):

tsdef/TrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
+import { EventEmitter } from 'events';
+import { Track } from './Track';
+
+export class TrackPublication extends EventEmitter {
+  isEnabled: boolean;
+  kind: Track.Kind;
```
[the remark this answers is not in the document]
PikaJoyce: `Kind` is defined as `'audio' | 'video' | 'data'` within the `Track` namespace inside of `Track.d.ts`. Since both the class and namespace are exported, there is type merging. So we can import `Track` from `Track.d.ts` and then have access to `Track.Kind` type definition which is `string`.
image
image

manjeshbhargav (approved):

PR #1225, opened 2020-10-13
JSDK-3022 DTX for Opus codec.

manjeshbhargav:
@makarandp0 @ceaglest 

This pull request implements the ConnectOptions API spec changes specified [here](https://code.hq.twilio.com/client/video-sdk-api/pull/73). The LocalTrackPublishOptions API spec changes will be implemented in a separate release.

## Checklist

- [x] Implementation
- [x] Manual verification with React App
- [x] Unit tests
- [x] Integration tests
- [x] Changelog entry
- [x] +1 from at least one reviewer

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 New Features
 ------------
 
-- Enabled discontinuous transmission (DTX) in the Opus audio codec, which will result in
-  bandwidth and CPU savings during silence and background noise. You can control this
-  feature using the ConnectOptions property `preferredAudioCodecs`. (JSDK-3022)
+- Enabled discontinuous transmission (DTX) in the Opus audio codec by default, which
+  will result in bandwidth and CPU savings during silence and background noise. You
+  can control this feature using the ConnectOptions property `preferredAudioCodecs`. (JSDK-3022)
   
   ```js
   const { connect } = require('twilio-video');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,28 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Support for 1.x will cease on December 4th, 2020**. This branch will only receive fixes for critical issues until that date. Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) when planning your migration to 2.x. For details on the 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.7.3 (in progress)
+2.8.0 (in progress)
 ===================
+
+New Features
+------------
+
+- Enabled discontinuous transmission (DTX) in the Opus audio codec, which will result in
+  bandwidth and CPU savings during silence and background noise. You can control this
+  feature using the ConnectOptions property `preferredAudioCodecs`. (JSDK-3022)
+  
+  ```js
+  const { connect } = require('twilio-video');
+  
+  // Disable DTX for Opus.
+  connect('token', {
+    preferredAudioCodecs: [{ codec: 'opus', dtx: false }]
+  });
+  ```
 
 Bug Fixes
 ---------
 - Fixed a bug where LocalTrack event listeners were not being cleaned up after disconnecting from a room. (JSDK-2985)
-
 
 2.7.2 (August 12, 2020)
 =======================
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -259,8 +259,8 @@ function connect(token, options) {
   });
 
   const preferredCodecs = {
-    audio: options.preferredAudioCodecs,
-    video: options.preferredVideoCodecs.map(normalizeVideoCodecSettings)
+    audio: options.preferredAudioCodecs.map(normalizeCodecSettings),
+    video: options.preferredVideoCodecs.map(normalizeCodecSettings)
   };
 
   const networkQualityConfiguration = new NetworkQualityConfigurationImpl(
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,14 +2,29 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
 **Support for 1.x will cease on December 4th, 2020**. This branch will only receive fixes for critical issues until that date. Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) when planning your migration to 2.x. For details on the 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.7.3 (in progress)
+2.8.0 (in progress)
 ===================
 
+New Features
+------------
+
+- Enabled discontinuous transmission (DTX) in the Opus audio codec, which will result in
```
idelgado, line 11: > in the Opus audio codec by default,

idelgado (changes requested):
Should I expect to see tests here too?

For example,

When I join with DTX enabled to a Group Room I see my sender bitrate lower.
When I join P2P and both peers enables DTX
When I join P2P and 1 peer enables DTX

manjeshbhargav:
@idelgado 

> Should I expect to see tests here too?
>
> For example,
>
> When I join with DTX enabled to a Group Room I see my sender bitrate lower.
When I join P2P and both peers enables DTX
When I join P2P and 1 peer enables DTX

We do have tests to verify the application of `usedtx=1` to the local sdps. However, I don't know if there is a way to write deterministic integration tests because you have to synthesize speech and silence to test the codec behavior. Also, the bitrate changes take some time, so I need to use timeouts to calculate the bitrate, which can be flaky.

idelgado:
@manjeshbhargav what if you were to use disabled? Does the bitrate look different when DTX is on for the sender?

I think integration tests make sense here. Maybe this is something QE should add but it should be somewhere and the behavior should be validated and codified by tests.

manjeshbhargav:
@idelgado 

Yes, the bitrate does go down by a lot during silence, but not all the way to 0. We would have to come up with a bitrate threshold estimate to compare against during silence.

manjeshbhargav:
@idelgado I tried testing with a synthetic audio track, and opus seems to behave differently than microphone tracks. The bandwidth reduction is very less in Chrome and Safari and non-existent in Firefox. But, if I use a microphone track, dtx operates as expected. I’m not sure how to simulate microphone tracks in integration tests. This feature most likely will require manual verification.

manjeshbhargav:
@idelgado I tried testing with a synthetic audio track, and opus seems to behave differently than microphone tracks. The bandwidth reduction is very less in Chrome and Safari and non-existent in Firefox. But, if I use a microphone track, dtx operates as expected. I’m not sure how to simulate microphone tracks in integration tests. This feature most likely will require manual verification.

idelgado (approved):
Please have QE validate the scenarios I listed.

CHANGELOG.md
```diff
@@ -2,14 +2,29 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
-2.7.3 (in progress)
+2.8.0 (in progress)
 ===================
 
+New Features
+------------
+
+- Enabled discontinuous transmission (DTX) in the Opus audio codec by default, which
+  will result in bandwidth and CPU savings during silence and background noise. You
+  can control this feature using the ConnectOptions property `preferredAudioCodecs`. (JSDK-3022)
```
makarandp0, line 13: I think we should add more details about what should application consider disabling dtx.

CHANGELOG.md
```diff
@@ -2,14 +2,29 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Support for 1.x will cease on December 4th, 2020**. This branch will only receive fixes for critical issues until that date. Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) when planning your migration to 2.x. For details on the 1.x changes, go [here](https://github.com/twilio/twilio-video.js/blob/support-1.x/CHANGELOG.md).
 
-2.7.3 (in progress)
+2.8.0 (in progress)
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Yes. We need to do that.

CHANGELOG.md
manjeshbhargav: @makarandp0 I want to test with a non-speech audio track (like music). I'll update the changelog after that.

CHANGELOG.md
```diff
@@ -2,14 +2,29 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
 ===================
 
+New Features
+------------
+
+- Enabled discontinuous transmission (DTX) in the Opus audio codec, which will result in
+  bandwidth and CPU savings during silence and background noise. You can control this
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 I don't think that's needed.

ceaglest (commented):
Overall the changes look good.

From an API compatibility standpoint, do you think it is too aggressive to enable DTX by default in a minor version? I am not totally opposed since it is an optimization, and the primary use case is voice traffic.

It would be nice to observe the reduction in bandwidth after enabling DTX in a test. Could you inject silence and then patch in a voice recording of the say verb?

manjeshbhargav:
@ceaglest ,

I don't think it is too aggressive to enable DTX by default, since its an optimization. Also, making this opt-in requires customer education and proactive steps on our part to get them to adopt it.

ceaglest (approved):
Ack, we should be prepared for the possibility that some customers will not want DTX and will reach out to us about it after trying 2.8.0. This PR is a good start on the education front.

Use your discretion on the SDK level test coverage.

PR #1230, opened 2020-10-15
JSDK-2807 - another race condition when Recycling  transceivers

makarandp0:
We have a logic to reuse the transceiver (so that we do not grow m-lines). The logic had issue where in some conditions it caused the transceivers to get recycled even though there was a sender track associated with it.

The [repro](https://app.circleci.com/pipelines/github/twilio/twilio-video.js/2553/workflows/311459fc-71ff-41a0-bdbc-0d97478279ca/jobs/27862/parallel-runs/5/steps/5-105) followed following sequence :

Room: RM2f02ea350c28cde64bfe177c55a64106
1) Negotiation (A) Finishes.
2) Transceiver are recycled after (A)
3) Negotiation (B) starts: `setRemoteDesciption(offer)`
4) Track (T1) is Added - `addTrack` (uses recycled transceiver (TT1) and marks it as 'sendrecv')
5) Negotiation (B) finishes. 
6) Transceiver are recycled after (B) - Here we end-up recycling the transceiver (TT1) that has a sender track associated and direction === 'sendrecv'
7) Another Track (T2) is Added. (ends up replacing T1! )
8) Next negotiation does not include T1!!
 
The fix is to exclude any transceivers with direction = sendrecv or direction = sendonly.

**Contributing to Twilio**


lib/signaling/v2/peerconnection.js
```diff
@@ -1499,7 +1499,9 @@ function shouldRecycleTransceiver(transceiver, pcv2) {
   return !transceiver.stopped && !pcv2._replaceTrackPromises.has(transceiver)
     && (transceiver.currentDirection === 'inactive'
       || transceiver.currentDirection === 'recvonly'
-      || transceiver.direction === 'recvonly');
+      || transceiver.direction === 'recvonly')
+    && transceiver.direction !== 'sendrecv'
+    && transceiver.direction !== 'sendonly';
```
manjeshbhargav, line 1504: @makarandp0 Can't this be simplified to:
```suggestion
      || transceiver.direction === 'recvonly')
    && !transceiver.sender.track;
```
?

makarandp0: @manjeshbhargav - I think your suggestion would work, but using direction/currentDirection makes the intent more explicit.

makarandp0: also I am also not sure if `transceiver.sender.track`  would always be null. but if you think `transceiver.sender.track` is better indicator how about:

```
function shouldRecycleTransceiver(transceiver, pcv2) {
  return !transceiver.stopped && !pcv2._replaceTrackPromises.has(transceiver) && !transceiver.sender.track
}
```

manjeshbhargav: @makarandp0 I think the trouble here is that we are using `currentDirection`, which specifies the current transceiver direction. So, in this case, even though `direction` was set to `sendrecv`, it still got recycled because `currentDirection` is `recvonly`. So, I think we should just get rid of the dependence on `currentDirection`:
```js
return !transceiver.stopped
  && !pcv2._replaceTrackPromises.has(transceiver)
  && ['inactive', 'recvonly'].includes(transceiver.direction);
```

manjeshbhargav (changes requested):
@makarandp0 On second thought, I think depending on directions

makarandp0:
Thanks @manjeshbhargav I have incorporated your suggestion and updated to only look at `transceiver.direction`, Can you take another look?

manjeshbhargav (approved):
+1 pending integration tests passing.

PR #1269, opened 2020-11-11
JSDK-3063 Don't send the local SDP with DTX enabled over the wire.

manjeshbhargav:
@makarandp0 @PikaJoyce 

This PR ensures that the local SDP with `usedtx=1` is not sent over the wire, only the original SDP is sent over.
This makes sure that older SDKs do not enable DTX even though the feature is not supported. Newer SDKs add this flag by themselves if DTX is enabled, so it should not affect them.

**Contributing to Twilio**


makarandp0 (approved):
👍

PikaJoyce (approved):

PR #1271, opened 2020-11-13
WIP: Adding logger module

charliesantos:
This is a work in progress but requires initial review. Please do not merge yet. Reference [document](https://docs.google.com/document/d/14VEOOrliOX4hi8bxtHSDfNk7pKzeLvpMxuvD6o61UXY/edit#heading=h.zdro3cj5pdko).


lib/util/log.js
```diff
@@ -57,6 +59,29 @@ class Log {
...
+          let level = this._logLevels[moduleName] || DEFAULT_LOG_LEVEL;
+
+          // There is no 'off' in the logger module. It uses 'silent' instead
+          level = level === 'off' ? 'silent' : level;
+
+          logger.setDefaultLevel(level);
+          return logger;
+        }
+      },
+      _logLevelsEqual: {
```
makarandp0, line 79: I think we do not need to keep old logger. we should either 
a) remove support for logLevels object - which allows different log levels for different modules
b) use different loggerName for different modules. that way we can keep the logLevels object support but still use "loglevel" library.
for example all logs for a given modulename will go in: 
var moduleLogger = require("loglevel").getLogger(this._loggerName + "-" + moduleName);

charliesantos: Per our slack discussion, we'll go with b

charliesantos:
Will reopen against a feature branch.

PR #1275, opened 2020-11-16
JSDK-2373 | Implementing loglevel module

charliesantos:
**Contributing to Twilio**

- [x] Implement
- [x] Update changelog
- [x] Update docs
- [x] Add tests
- [x] Manual test using quickstart


lib/createlocaltrack.js
```diff
@@ -92,7 +95,14 @@ function createLocalVideoTrack(options) {
  * also specify any of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints" target="_blank">MediaTrackConstraints</a>
  * properties.
  * @typedef {MediaTrackConstraints} CreateLocalTrackOptions
- * @property {LogLevel|LogLevels} logLevel
```
makarandp0, line 95: I think we should support the loglevel (but mark it as deprecated)

lib/index.js
```diff
@@ -52,6 +57,10 @@ Object.defineProperties(Video, {
     enumerable: true,
     value: LocalVideoTrack
   },
+  Logger: {
```
makarandp0, line 60: This is interesting. Just curious - why do we expose Logger here instead of asking developers `require('loglevel')` themselves? What are the tradeoffs?

CHANGELOG.md
```diff
@@ -21,6 +21,47 @@ New Features
   });
   ```
 
+- Exposed SDK logger using [loglevel](https://www.npmjs.com/package/loglevel) module to allow for updating logging configuration after calling `Video.connect`, `Video.createLocalAudioTrack`, `Video.createLocalTracks` and `Video.createLocalVideoTrack`. With this feature, logs can now be intercepted at runtime to allow realtime processing of the logs which include but not limited to inspecting the log data and sending it to your own server.
```
makarandp0, line 24: we should mention that `eventListener` parameter to connect is now deprecated with the implementation of this custom logger, and specify an example that shows usage of logger for listening to events.

makarandp0 (changes requested):
@charliesantos - This is looking good! I have added my initial hi-level review.  Mainly I would like to see

1) logLevel marked as deprecated parameter (and remove documentation for this parameter) - Although we would still keep support for it.  
But let's wait see what @manjeshbhargav has to say about it.

2) de-emphasize `loggerName` - it need not be used in main examples.

3) update/remove eventObserver to not emit events on eventListener. 

4) sorry we do not have any integration tests that demonstrate `eventListener` - But I would like to see some new integration tests that use this functionality. 
which could also demonstrate how it's used to substitute `eventListener`

lib/createlocaltrack.js
charliesantos: I don't think we should mark this as deprecated because we don't plan on removing this functionality, including the ability to specify different levels for each module. Let's hear what @manjeshbhargav has to say.

lib/index.js
charliesantos: If user require the module themselves, they won't be able to see the effects of `setLevel` until a page reloads happen. Loglevel uses local storage to save the state and the SDK will pick up the state after page reload. The other downside is that adding a plugin to the loglevel module only work if npm bundles the same loglevel version. So if the user happens to use a different loglevel, they won't be able to intercept the logs.

CHANGELOG.md
charliesantos: I think deprecating `eventListener` is out of scope and should be on a separate PR.

CHANGELOG.md
```diff
@@ -21,6 +21,47 @@ New Features
...
+
+  ```js
+  var { Logger, connect } = require('twilio-video');
+  var token = getAccessToken();
+
+  // Connect using a custom logger name
+  connect(token, {
+    loggerName: 'logger-alice',
```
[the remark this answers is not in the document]
charliesantos: 👍  to de-emphasize `loggerName`. But I don't think we should deprecate `logLevel` since we don't plan on removing it, especially the ability to specify different levels for individual components.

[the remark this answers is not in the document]
makarandp0: @charliesantos - 

Re: `logLevel` parameter:
This parameter is only there so that developer could specify individual logLevels for subModules right? I think we should deprecate/remove this parameter - as this feature is only useful for sdk developers and not useful for app developers. As @manjeshbhargav suggested, we can continue to support the parameter while its marked as deprecated.

CHANGELOG.md
makarandp0: I think it's okay to deprecate the eventListener in separate PR - I was pushing for doing it in this PR to just be sure that it can be done easily with logLevel.

CHANGELOG.md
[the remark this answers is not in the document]
charliesantos: Understood @makarandp0 . I will mark it as deprecated.

CHANGELOG.md
charliesantos: Thanks @makarandp0 . It should be super easy. Please check the integration tests. I added a test that listens for room name. The user can do a similar way when we deprecate eventListener.

lib/createlocaltrack.js
manjeshbhargav: @charliesantos The goal is to replace the existing log functionality with the new logger in the next major version. So, we can mark the `logLevel` parameter as deprecated and ask them to use the new logger instead.

manjeshbhargav: @charliesantos You should also print a console warning if `logLevel` is specified in ConnectOptions.

manjeshbhargav (commented):
@charliesantos Great work! I have a couple of comments, but I'll leave it to @makarandp0 to approve the PR, since you have been working with him closely.

test/integration/spec/logger.js
```diff
@@ -0,0 +1,101 @@
...
+    assert(!!callSpies.length);
+    callSpies.forEach(callSpy => {
+      ['datetime', 'logLevel', 'component', 'message'].forEach(argName => {
+        assert(!!callSpy.args[0]);
+        assert(!!callSpy.args[0][argName]);
+      });
+    });
+  });
+
+  it('should provide optional data via the plugin', async () => {
```
makarandp0, line 86: I think we need couple of tests to verify that 
the specified loglevel is respected. 
a) That is when logLevel = 'warn' we should not see logs debug logs.  
b) when loglevel is 'debug' we should see 'debug' logs.
I tried b) and did not see debug logs.

test/integration/spec/logger.js
```diff
@@ -0,0 +1,101 @@
...
+    const room = await connect(token, Object.assign({ name: sid }, defaults));
+    const callSpies = loggerCb.getCalls();
+    assert(!!callSpies.length);
+
+    let loggedRoomName;
+    callSpies.forEach(callSpy => {
+      const { message, data } = callSpy.args[0];
+      if (message === 'Room name:') {
+        loggedRoomName = data;
```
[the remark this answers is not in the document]
makarandp0: we should also add an example that connects with two different moduleNames (alice and bob), and then verifies that logs can be distinguished and decorated.

makarandp0 (changes requested):
@charliesantos I have few more change request.
also I was trying this branch - and did not see it work as expected. with logLevel: 'debug' - the output was missiing .debug logs. Maybe I did something wrong, but please see if you see it working with logLevel: 'debug'

lib/index.js
charliesantos: >by state - you mean logLevel previously set right? if they requireed and set logLevel before calling twilio-video APIs then issue #1 can be addressed right?

Yes. If user only uses loglevel to set the level, they can do so freely by just requiring loglevel module themselves.

>regarding loglevel version - How would this work if they are using multiple libraries that depend on loglevel. Say voice and video - if video requires one version and voice another ?

It will still work if the user is only using the `setLevel` APIs because loglevel saves this state to localStorage. However, it won't work if the user wants to intercept the logs. This seems to be a limitation of the loglevel module because it saves its list of logger objects inside the module's scope, not globally.

charliesantos:
> @charliesantos I have few more change request.
also I was trying this branch - and did not see it work as expected. with logLevel: 'debug' - the output was missiing .debug logs. Maybe I did something wrong, but please see if you see it working with logLevel: 'debug'

@makarandp0  No worries. Thanks for the review! Not sure why it's not working on yours. Please check one of my integration tests as a reference. I tried it manually and I'm able to get all loglevels to work.

lib/connect.js
```diff
@@ -373,10 +456,14 @@ function connect(token, options) {
...
  *  here</a> for more information.
- * @property {LogLevel|LogLevels} [logLevel='warn'] - Set the log verbosity
- *   of logging to console. Passing a {@link LogLevel} string will use the same
+ * @property {LogLevel|LogLevels} [logLevel='warn'] - Set the default log verbosity
+ *   of logging. Passing a {@link LogLevel} string will use the same
  *   level for all components. Pass a {@link LogLevels} to set specific log
  *   levels.
+ * @property {string} [loggerName='twilio-video'] - The name of the logger. Use this name when accessing the logger
+ *   used by the SDK. If you provided a {@link LogLevels} object with different levels via the <code>logLevel</code>
+ *   parameter, you can access the loggers used by the SDK using the naming format <code>loggerName + '-' + component</code>.
```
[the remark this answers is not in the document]
charliesantos: The list of valid components are already documented and mentioned here. It's the `{@link LogLevels}` 😃 . Maybe the wording is not correct?

lib/createlocaltrack.js
```diff
@@ -92,7 +95,14 @@ function createLocalVideoTrack(options) {
...
+ * @property {LogLevel|LogLevels} [logLevel='warn'] - Set the default log verbosity
+ *   of logging. Passing a {@link LogLevel} string will use the same
+ *   level for all components. Pass a {@link LogLevels} to set specific log
```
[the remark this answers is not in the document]
charliesantos: The list of valid components are already documented and mentioned here. It's the {@link LogLevels} 😃 . Maybe the wording is not correct?

test/integration/spec/logger.js
charliesantos: I think the reason why b) is not working for you is because you previous called `setLevel('warn')` which would save the state to localStorage. Just a reminder, `logLevel` parameter in connect options will be used as the default loglevel. See [here](https://github.com/twilio/twilio-video.js/blob/JSDK-2373/lib/util/log.js#L82). Meaning, if there's no previously saved state, the logger will use the default one. But if there's a previously saved loglevel, logger will use that instead. I added this in the `logLevel` parameter documentation.

makarandp0: ah got it! thanks for the explaination - I guess it sounds fine then.

test/integration/spec/logger.js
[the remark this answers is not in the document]
charliesantos: >I think we should should talk about bit more. But what you do think about -
always require that optional data is an object and not a value type. This will help developers handle it in a generic way.

As discussed, we will address this when we deprecate eventListener. We need to involve other people.

lib/connect.js
```diff
@@ -373,10 +450,16 @@ function connect(token, options) {
...
- *   of logging to console. Passing a {@link LogLevel} string will use the same
+ * @property {LogLevel|LogLevels} [logLevel='warn'] - <code>(deprecated: use [Video.Logger](module-twilio-video.html) instead.
+ *   See [examples](module-twilio-video.html#.connect) for details)</code>
+ *   Set the default log verbosity
+ *   of logging. Passing a {@link LogLevel} string will use the same
  *   level for all components. Pass a {@link LogLevels} to set specific log
  *   levels.
+ * @property {string} [loggerName='twilio-video'] - The name of the logger. Use this name when accessing the logger
+ *   used by the SDK. If you provided a {@link LogLevels} object with different levels via the <code>logLevel</code>
+ *   parameter, you can access the loggers used by the SDK using the naming format <code>loggerName + '-' + component</code>.
```
makarandp0, line 461: we should remove mention of logLevel parameter from here. If they supplied `loggerName` then they definitely should not supply custom logLevels.
```suggestion
 * @property {string} [loggerName='twilio-video'] - The name of the logger. Use this name when accessing the logger used by the SDK. 
```

makarandp0 (changes requested):
Hey @charliesantos - I have just couple more change requests and comments.

lib/connect.js
```diff
@@ -737,21 +820,24 @@ const EventListenerGroup = {
...
 ]);
 
-function removeDeprecatedConnectOptionsPropsWithWarning(options, log) {
+function deprecateOptions(options, log) {
   deprecatedConnectOptionsProps.forEach(prop => {
-    const { didWarn, name, newName } = prop;
-    if (name in options) {
+    const { didWarn, name, newName, shouldDelete } = prop;
+    if (name in options && typeof options[name] !== 'undefined') {
       if (newName) {
```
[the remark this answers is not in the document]
charliesantos: Good catch. But should be `if (newName && shouldDelete)` 😄

CHANGELOG.md
```diff
@@ -33,6 +33,45 @@ New Features
   });
   ```
 
+- twilio-video now allows customizing logger using [loglevel](https://www.npmjs.com/package/loglevel) module. With this feature, logs can now be intercepted at runtime to allow real-time processing of the logs which include but not limited to inspecting the log data and sending it to your own server.
+
+With this `ConnectOptions`'s `logLevel` property is now deprecated and you should start using this new SDK logger instead.
+
+
+  In addition, the `ConnectOptions`'s `logLevel` property is now deprecated and you should start using this new SDK logger instead.
```
makarandp0: duplicate line...also some formatting need to be updated.

makarandp0 (approved):
this is super exciting work! 👍

PR #1285, opened 2020-11-21
Deprecate ConnectOptions.eventlistener property

makarandp0:
With the introduction of[ customizable logger](https://github.com/twilio/twilio-video.js/pull/1275). Applications can use the same logger to listen for signaling events that used be fired on 'eventListener'. 

This change deprecates 'eventListener' and logs the event on logger instead.

**Contributing to Twilio**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You
 
       // check for signaling events that previously were
       // emitted on (now deprecated) eventListener
-      if (message === 'event') {
+      if (message === 'event' && data.group === 'signaling') {
         if (data.name === 'waiting') {
           assert.equal(data.level, 'warning');
           console.warn('Twilio\'s signaling server is busy, so we wait a little while before trying again.');
@@ -50,7 +50,7 @@ Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You
           }
         }
       }
-      // You can send to your own server
+      // You may also send the logs to your own server
       postDataToServer(arguments);
 
     };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ New Features
 
 With this change `ConnectOptions`'s `logLevel` property is now deprecated. You can instead use `logger.setLevel` to set the desired log level.
 
-Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting logger as shown in example below.
+Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting the logger's messages as shown in the example below.
 
   Example:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 New Features
 ------------
 
-- twilio-video now allows customizing logger using [loglevel](https://www.npmjs.com/package/loglevel) module. With this feature, logs can now be intercepted at runtime to allow real-time processing of the logs which include but not limited to inspecting the log data and sending it to your own server.
+- twilio-video now allows customizing logger using [loglevel](https://www.npmjs.com/package/loglevel) module. With this feature, logs can now be intercepted at runtime to allow real-time processing of the logs which include but not limited to inspecting the log data and sending it to your own server. (JSDK-2373)
 
 With this change `ConnectOptions`'s `logLevel` property is now deprecated. You can instead use `logger.setLevel` to set the desired log level.
   ```js
@@ -28,7 +28,7 @@ With this change `ConnectOptions`'s `logLevel` property is now deprecated. You c
   });
   ```
 
-Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting the logger's messages as shown in the example below.
+Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting the logger's messages as shown in the example below. (JSDK-2977)
 
   Example:
 
@@ -79,8 +79,6 @@ Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You
     console.log('Could not connect to the Room:', error.message);
   });
   ```
-
-
 
 2.9.0 (in progress)
 ===================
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -26,10 +28,35 @@ New Features
 
     return function (datetime, logLevel, component, message, data) {
       method(datetime, logLevel, component, message, data);
-      // Send to your own server
+
+      // check for signaling events that previously were
+      // emitted on (now deprecated) eventListener
+      if (message === 'event' && data.group === 'signaling') {
+        if (data.name === 'waiting') {
+          assert.equal(data.level, 'warning');
```
charliesantos, line 36: Should we remove these assertions and provide an example data object instead?

lib/util/eventobserver.js
```diff
@@ -56,15 +57,16 @@ class EventObserver extends EventEmitter {
         const publisherPayload = Object.assign(payload ? payload : {}, { elapsedTime, level });
         this._publisher.publish(group, name, publisherPayload);
       }
+      const event = Object.assign(payload ? { payload } : {}, {
+        elapsedTime,
+        group,
+        level,
```
charliesantos, line 63: Should we remove `level`? This is already available in the original logger method factory. It'd be good to advise users to use that when determining log level for any type of log.

```js
var originalFactory = logger.methodFactory;
logger.methodFactory = function (methodName, level, loggerName) {
  var method = originalFactory(methodName, level, loggerName);
  return function (datetime, logLevel, component, message, data) {
    method(datetime, logLevel, component, message, data);

    console.log(level); // prints debug, info, warn, or error
    console.log(logLevel); // prints 0 - 3. These are coming from Log.DEBUG, Log.INFO, etc
  };
};
```

lib/util/eventobserver.js
```diff
@@ -56,15 +57,16 @@ class EventObserver extends EventEmitter {
...
+        name,
+        timestamp
+      });
 
+      log.info('event', event);
```
charliesantos, line 68: Since logger requires a minimum of `info` level, should we mention this in the changelog and documentation?

test/integration/spec/logger.js
```diff
@@ -124,6 +124,32 @@ describe('logger', function() {
     assert.equal(room.name, loggedRoomName);
   });
 
+  it('should log signaling events', async () => {
```
charliesantos, line 127: 👍

test/integration/spec/logger.js
```diff
@@ -124,6 +124,32 @@ describe('logger', function() {
...
+    callSpies.forEach(callSpy => {
+      const { message, data } = callSpy.args[0];
+      if (message === 'event' && data.group === 'signaling') {
+        assert(typeof data.elapsedTime === 'number');
+        assert(typeof data.timestamp === 'number');
+        assert(typeof data.level === 'string');
+        assert(typeof data.name === 'string');
+        early = early || data.name === 'early';
+        connecting = connecting || data.name === 'connecting';
+        open = open || data.name === 'open';
```
charliesantos, line 145: Should we also test for `closed`? Also, `waiting` was mentioned in the change log. Should we also test for `waiting`?

charliesantos (changes requested):
Nice @makarandp0 ! Looks great! Just left a few comments.

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ New Features
 
 - twilio-video now allows customizing logger using [loglevel](https://www.npmjs.com/package/loglevel) module. With this feature, logs can now be intercepted at runtime to allow real-time processing of the logs which include but not limited to inspecting the log data and sending it to your own server.
 
-  With this `ConnectOptions`'s `logLevel` property is now deprecated and you should start using this new SDK logger instead.
+With this change `ConnectOptions`'s `logLevel` property is now deprecated. You can instead use `logger.setLevel` to set the desired log level.
+
```
manjeshbhargav, line 13:
```suggestion
With this change `ConnectOptions`'s `logLevel` property is now deprecated. You can instead use `logger.setLevel` to set the desired log level.
```

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ New Features
...
+Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting logger as shown in example below.
```
manjeshbhargav, line 14:
```suggestion
Additionally  `ConnectOptions`'s `eventListener` property is now deprecated. You can listen for the signaling events by intercepting the logger's messages as shown in the example below.
```

manjeshbhargav (commented):
@makarandp0 Please add the ticket number for the logger implementation and deprecation entries. I'll let @charliesantos be the approver.

test/integration/spec/logger.js
makarandp0: we cant test for closed since it happens only after we disconnect from room. and `waiting` is not a reproducible state.

lib/util/eventobserver.js
makarandp0: event level tells the severity of the event ( for example signaling `closed` would fire at level='error'. I am taking your suggestion to use log those at appropriate logLevel. But I think we do need to keep these.

test/integration/spec/logger.js
makarandp0: I will update to check the call spies after disconnect. This will let us check for 'closed' event as well.

CHANGELOG.md
makarandp0: I have now removed the asserts against `data.level` - But I am not sure if we should reproduce the documentation of the events object here. They are documented in https://media.twiliocdn.com/sdk/js/video/releases/2.8.0/docs/global.html
I tried to keep the sample similar to the original one that shows how to use `eventListener`

lib/util/eventobserver.js
makarandp0: yes, I have added a comment about this in the example.

charliesantos (approved):
Looks great!

charliesantos:
Per @manjeshbhargav , please add the ticket number for the logger implementation and deprecation entries. I'll let @charliesantos be the approver.

test/unit/spec/connect.js
```diff
@@ -82,6 +82,12 @@ describe('connect', () => {
       value: true,
       shouldDelete: true
     },
+    {
+      name: 'eventListener',
```
idelgado, line 86: The name might be a bit confusing here, should this be called "loggerListener" or "logEntryListener"?

I'm assuming this is a listener for when the logger logs an event, as opposed to the deprecated interface which I believe was also called `eventListener`

idelgado (changes requested):

test/unit/spec/connect.js
makarandp0: @idelgado  This test is verifying that we handle the deprecated 'eventListener' correctly. The name need to be `eventListener`

idelgado (approved):

PR #1292, opened 2020-11-24
Dev-TypeScript

PikaJoyce:
This draft PR is to review anything that is currently missing from the TS Definitions. 

**Contributing to Twilio**


tsdef/Room.d.ts
```diff
@@ -0,0 +1,52 @@
+import { MediaConnectionError, SignalingConnectionDisconnectedError, TwilioError } from './TwilioError';
```
manjeshbhargav, line 1: Please remove the signaling and media error definitions. Let's just stick with `TwilioError` for now:
```suggestion
import { TwilioError } from './TwilioError';
```

tsdef/Room.d.ts
```diff
@@ -0,0 +1,52 @@
...
+  getStats(): Promise<StatsReport[]>;
+
+  on(event: 'disconnected', listener: (room: Room, error: TwilioError) => void): this;
+  on(event: 'dominantSpeakerChanged', listener: (dominantSpeaker: RemoteParticipant) => void): this;
+  on(event: 'participantConnected', listener: (participant: RemoteParticipant) => void): this;
+  on(event: 'participantDisconnected', listener: (participant: RemoteParticipant) => void): this;
+  on(event: 'participantReconnected', listener: (participant: RemoteParticipant) => void): this;
+  on(event: 'participantReconnecting', listener: (participant: RemoteParticipant) => void): this;
+  on(event: 'reconnected', listener: () => void): this;
+  on(event: 'reconnecting', listener: (error: MediaConnectionError | SignalingConnectionDisconnectedError) => void): this;
```
manjeshbhargav, line 36:
```suggestion
  on(event: 'reconnecting', listener: (error: TwilioError) => void): this;
```

tsdef/types.d.ts
```diff
@@ -0,0 +1,156 @@
...
+export class NetworkQualityMediaStats {
+  send: NetworkQualityLevel;
+  recv: NetworkQualityLevel;
+  sendStats: NetworkQualitySendOrRecvStats | null;
+  recvStats: NetworkQualitySendOrRecvStats | null;
+}
+
+export class NetworkQualityAudioStats extends NetworkQualityMediaStats {}
+
+export class NetworkQualityVideoStats extends NetworkQualityMediaStats {}
```
manjeshbhargav, line 46:
```suggestion
```

tsdef/types.d.ts
```diff
@@ -0,0 +1,156 @@
...
+}
+
+export type NetworkQualityLevel = number;
+
+export type NetworkQualityVerbosity = 0 | 1 | 2 | 3;
+
+export class NetworkQualityStats {
+  level: NetworkQualityLevel;
+  audio: NetworkQualityAudioStats | null;
+  video: NetworkQualityVideoStats | null;
```
manjeshbhargav, line 34:
```suggestion
  audio: NetworkQualityMediaStats | null;
  video: NetworkQualityMediaStats | null;
```

tsdef/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+export { LocalTrack, RemoteTrack, LogLevel, LogLevels } from './types';
+export { LocalAudioTrack } from './LocalAudioTrack';
+export { LocalAudioTrackPublication } from './LocalAudioTrackPublication';
+export { LocalDataTrack } from './LocalDataTrack';
+export { LocalDataTrackPublication } from './LocalDataTrackPublication';
+export { LocalParticipant } from './LocalParticipant';
+export { LocalTrackPublication } from './LocalTrackPublication';
+export { LocalVideoTrack } from './LocalVideoTrack';
+export { LocalVideoTrackPublication } from './LocalVideoTrackPublication';
+export { Participant } from './Participant';
```
manjeshbhargav, line 23: `Participant` is a template class, so you can't export that. Instead you can export a type:
```suggestion
type Participant = LocalParticipant | RemoteParticipant;
export Participant;
```

tsdef/types.d.ts
```diff
@@ -0,0 +1,156 @@
...
+  insights?: boolean;
+  maxAudioBitrate?: number | null;
+  maxVideoBitrate?: number | null;
+  name?: string | null;
+  networkQuality?: boolean | NetworkQualityConfiguration;
+  region?: string;
+  preferredAudioCodecs?: AudioCodec[];
+  preferredVideoCodecs?: Array<VideoCodec | VideoCodecSettings | VP8CodecSettings>;
+  logLevel?: LogLevel | LogLevels;
+  tracks?: LocalTrack[] | MediaStreamTrack[];
```
manjeshbhargav, line 143:
```suggestion
  tracks?: Array<LocalTrack | MediaStreamTrack>;
```

tsdef/LocalParticipant.d.ts
```diff
@@ -0,0 +1,34 @@
...
+import { TwilioError } from './TwilioError';
+
+export class LocalParticipant extends Participant<LocalAudioTrackPublication, LocalDataTrackPublication, LocalVideoTrackPublication, LocalTrackPublication> {
+  audioTracks: Map<Track.SID, LocalAudioTrackPublication>;
+  dataTracks: Map<Track.SID, LocalDataTrackPublication>;
+  tracks: Map<Track.SID, LocalTrackPublication>;
+  videoTracks: Map<Track.SID, LocalVideoTrackPublication>;
+  signalingRegion: string;
+
+  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackOptions): Promise<LocalTrackPublication>;
```
manjeshbhargav, line 18:
```suggestion
  publishTrack(track: LocalTrack, options?: LocalTrackPublishOptions): Promise<LocalTrackPublication>;
  publishTrack(track: MediaStreamTrack, options?: MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
```

tsdef/TrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
+import { EventEmitter } from 'events';
+import { Track } from './Track';
+
+export class TrackPublication extends EventEmitter {
+  isEnabled: boolean;
+  kind: Track.Kind;
+  trackName: string;
+  trackSid: Track.SID;
+
+  toJSON(): string;
```
manjeshbhargav, line 10:
```suggestion
```

tsdef/TrackPublication.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  toString(): string;
```
manjeshbhargav, line 11:
```suggestion
```

tsdef/TwilioError.d.ts
```diff
@@ -0,0 +1,6 @@
+export class TwilioError extends Error {
+  code: number;
+  message: string;
+
+  toString(): string;
```
manjeshbhargav, line 5:
```suggestion
```

tsdef/Track.d.ts
```diff
@@ -0,0 +1,13 @@
+import { EventEmitter } from 'events';
+
+export namespace Track {
+  type Kind = 'audio' | 'video' | 'data';
```
manjeshbhargav, line 4:
```suggestion
  enum Kind {
    audio = 'audio',
    data = 'data',
    video = 'video'
  }
```

manjeshbhargav (changes requested):

manjeshbhargav:
Please address the circle ci build error as well.

tsdef/Track.d.ts
PikaJoyce: @manjeshbhargav After chatting with @timmydoza a bit to go over the definitions. We learned that you actually cannot have enums inside of definition files because the compiler does not compile the definitions into anything. So if we tried to use `Kind`, it would come back as undefined. I think we should keep them as types since there actually is a drop down menu when I re-tried today!

This is what it would look like trying to access LogLevel as a `type`:
image

Which still has a drop down menu to select which `LogLevel` we want!

timmydoza:
Nice work @PikaJoyce!

I tried importing these types into the React app, and I encountered a few errors. I think the changes that @manjeshbhargav suggested above will fix a lot of them (I get errors that `ConnectOptions` and `CreateLocalTracksOptions` are missing, for instance). But there are a few other things that I wanted to mention here.

#### `Participant` Class
I think we need to change how we export the `Participant` class. Look at this [line](https://github.com/twilio/twilio-video-app-react/blob/38be43a697f61c8b2b701d24fc0398682b989250/src/components/MainParticipantInfo/MainParticipantInfo.tsx#L75) from the MainParticipant component from the React app. This component is expecting a `Participant` as one of the props. The `Participant` class that these definitions [export](https://github.com/twilio/twilio-video.js/blob/dev-typescript/tsdef/Participant.d.ts#L11) requires that we supply four type arguments. Because of this, I'm getting an error when I try to compile the React app.

I think the solution here is to not export our `Participant` class as is. I think the name should be changed to something like, `ParticipantImpl` (the namespace name would change as well).  Then, the `RemoteParticipant` and `LocalParticipant` classes would inherit from `ParticipantImpl`, and they would still pass the four type arguments to it. 

Then, in [types.d.ts](https://github.com/twilio/twilio-video.js/blob/dev-typescript/tsdef/types.d.ts), we can define and export a `Participant` type as such:

```diff
export type LocalTrack = LocalAudioTrack | LocalVideoTrack | LocalDataTrack;
export type RemoteTrack = RemoteAudioTrack | RemoteVideoTrack | RemoteDataTrack;
export type AudioTrackPublication = LocalAudioTrackPublication | RemoteAudioTrackPublication;
export type DataTrackPublication = LocalDataTrackPublication | RemoteAudioTrackPublication;
export type VideoTrackPublication = LocalVideoTrackPublication | RemoteVideoTrackPublication;
+export type Participant = LocalParticipant | RemoteParticipant;
``` 
I included some of the other types to illustrate that this same pattern is already used in these type definitions. 

When I make the above changes, I can continue to use the `Participant` class in the React app without issue.

#### Classes that emit events

Another error that I encounter occurs when I try to use my own union types. See this [line](https://github.com/twilio/twilio-video-app-react/blob/master/src/hooks/useIsTrackEnabled/useIsTrackEnabled.tsx#L4) from the React app.  This `useIsTrackEnabled` hook is expecting that any local or remote media track be passed in.

The problem, is that the type signatures for the `LocalAudioTrack`, `LocalVideoTrack`, `RemoteAudioTrack`, and `RemoteVideoTrack` classes are incompatible. The way to fix this, is to do something that we've already done. We need to add the following type signature to these classes:

`on(event: string, listener: (...args: any[]) => void): this;`

With this, the above mentioned types have compatible signatures, and as a result we are able to successfully create a type union between them.
 
I think this needs to be done on any class that extends (no matter how distant) from the `EventEmitter` class. 

#### Other

There are a number of smaller issues that prevented the React app from compiling. For example, `LocalVideoTrack` is missing the `isEnabled` boolean. Many of these may be fixed by adding in the suggestions that Manjesh enumerated above. But I also recommend that you try to import these types in the React app, and see if you can compile it (with `npm start` or `npm run build`). 

Really nice work overall. This is so close! Please let me know if you have any questions.

tsdef/Track.d.ts
timmydoza: Correct. In an ordinary .ts file, an enum would be compiled into a JavaScript object. But an enum inside a .d.ts file never gets compiled into an object, so the compiled JS file will try to `require()` an object that doesn't exist.

tsdef/types.d.ts
PikaJoyce: @manjeshbhargav I can do this but I would need them for [NetworkQualityStats](https://media.twiliocdn.com/sdk/js/video/releases/2.8.0/docs/NetworkQualityStats.html). Should I still remove them? And if so, what should go in their place within `NetworkQualityStats`?

tsdef/index.d.ts
PikaJoyce: I can put this in types and export it out of there. I think it will look a little cleaner!

PR #1311, opened 2020-12-09
[WIP] Instrumentation to support video-inspector

makarandp0:
This adds some instrumentation to support video-inspector. 
- added a global variable to maintain list of rooms
- added an integration test that verifies that the global rooms array is maintained and logger emits a log entry when a new room is created.

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -6,6 +6,10 @@ const createCancelableRoomPromise = require('./cancelableroompromise');
 const createLocalTracks = require('./createlocaltracks');
 const EncodingParametersImpl = require('./encodingparameters');
 const LocalParticipant = require('./localparticipant');
+const topLevel = global.window || global;
+const rooms = [];
+topLevel._TwilioVideo = topLevel._TwilioVideo || {};
+topLevel._TwilioVideo.rooms = rooms || [];
 
 const {
   LocalAudioTrack,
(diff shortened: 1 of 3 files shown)


makarandp0:
@manjeshbhargav can you please take a look at this PR? we would like to get this in so that we can use video inspector against

manjeshbhargav (changes requested):
I don't think the SDK should be adding any object to the global namespace, which is shared by the application code as well. I don't want the SDK to pollute the application global namespace with extra objects that are not part of the public API. If we want to support this, we should communicate with the extension directly. I think this needs a proposal and discussion among the JS team and Ahoy.

timmydoza:
> I think this needs a proposal and discussion among the JS team and Ahoy.

I'd be happy to discuss this. There are a number of different approaches we can take here, and we should evaluate them. I think it would be also worthwhile to look at the implementation of other similar tools (like [React Devtools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en), for example).

For now we can continue to let the inspector look for the `room` object on a global window variable. This allows app developers to "opt-in" to using the inspector by exposing the `room` object themselves (just how the react app does with `window.twilioRoom`).

PR #1339, opened 2021-01-13
JSDK-3045 - Framework Tests

PikaJoyce:
Ticket : [JSDK-3045](https://issues.corp.twilio.com/browse/JSDK-3045)

This ticket is a small fix for the `no-frame` and `react` framework tests. The fix is just so it can take in the environment variable other than prod so it does not throw errors when we use credentials from different environments. 

This PR does **not** include a fix for the `angular` test.

**Contributing to Twilio**


test/framework/twilio-video-no-framework/app.js
```diff
@@ -18,8 +18,9 @@ function getQueryParameters(location) {
 }
 
 const token = (getQueryParameters(location).get('token') || [])[0] || '';
+const environment = (getQueryParameters(location).get('environment') || [])[0] || '';
 
-Twilio.Video.connect(token).then(room => {
+Twilio.Video.connect(token, Object.assign({ environment })).then(room => {
```
makarandp0, line 23:
```suggestion
Twilio.Video.connect(token, { environment }).then(room => {
```

test/framework/twilio-video-react/src/App.js
```diff
@@ -7,13 +7,23 @@ export default class App extends Component {
...
-    connect(this.props.token).then(room => {
-      this.setState({ room });
-      room.once('disconnected', () => this.forceUpdate());
-      room.disconnect();
-    }, error => {
-      this.setState({ error });
-    });
+    if (this.props.environment) {
+      connect(this.props.token, Object.assign({ environment: this.props.environment })).then(room => {
+        this.setState({ room });
```
makarandp0, line 12: instead of duplicating this code block, I would just create connectOptions outside like
```
const connectOptions =  this.props.environment ? { environment: this.props.environment } : {};
connect(this.prop.token, connectOptions).then(room => {
...
```

test/framework/twilio-video-react/src/index.js
```diff
@@ -19,8 +19,17 @@ function getQueryParameters(location) {
 
 // eslint-disable-next-line
 const token = (getQueryParameters(location).get('token') || [])[0] || '';
+// eslint-disable-next-line
+const environment = (getQueryParameters(location).get('environment') || [])[0] || '';
```
makarandp0, line 23: by ensuring that you always have correct `environment` you can avoid code changes in react render block below.
```suggestion
const environment = (getQueryParameters(location).get('environment') || [])[0] || 'prod';
```

makarandp0 (changes requested):
Glad that you found the root cause! I have few suggestions to simplify some code changes.

makarandp0:
btw, i still see that framework test is red above - I guess thats because of `angular` right?

PikaJoyce:
> btw, i still see that framework test is red above - I guess thats because of `angular` right?

Yes, I see this in CircleCI Log: `"test:framework:angular" exited with 1.`

I also haven't went in to fix the Angular test yet.

test/framework/twilio-video-no-framework/app.js
```diff
@@ -18,8 +18,9 @@ function getQueryParameters(location) {
 }
 
 const token = (getQueryParameters(location).get('token') || [])[0] || '';
+const environment = (getQueryParameters(location).get('environment') || [])[0] || '';
```
manjeshbhargav, line 21:
```suggestion
const environment = (getQueryParameters(location).get('environment') || [])[0];
```

test/framework/twilio-video-no-framework/app.js
manjeshbhargav:
```suggestion
Twilio.Video.connect(token, environment ? { environment } : {}).then(room => {
```

manjeshbhargav (approved):

makarandp0 (approved):
👍

package.json
```diff
@@ -135,7 +136,10 @@
   "dependencies": {
     "@twilio/webrtc": "4.3.2",
     "backoff": "^2.5.0",
+    "chromedriver": "^87.0.5",
```
makarandp0, line 139: @PikaJoyce - These dependencies should not go here.  We do have a step to install these as part of `test:framework:instsall' - We removed these packages from `devDependancies` to avoid downloading these as part of every circleci job.

makarandp0 (changes requested):
@PikaJoyce - we need to remove the dependencies. Was there any reason test:framework:install not enough for these?

package.json
PikaJoyce: Weird this got left in, had a conflict but this is now fixed.

PR #1341, opened 2021-01-15
JSDK-3020: Large room tests

makarandp0:
verifies that large rooms 
- can be created with MaxParticipant > 50
- does not cause participantConnected when a participant with no tracks joins
- track publish fails when max tracks limit is reached
- synthetic disconnect is emitted when participant unpublishes last track. 

**Contributing to Twilio**


test/integration/spec/room.js
```diff
@@ -147,6 +147,74 @@ describe('Room', function() {
...
+    });
+
+    it('does not send participantConnected when a participant joins with no tracks', async () => {
+      const bobConnectedPromise = new Promise(resolve => aliceRoom.on('participantConnected', resolve));
+      bobRoom = await connect(getToken('Bob'), Object.assign({ name: roomSid }, defaults, { tracks: [] }));
+
+      await waitForNot(bobConnectedPromise, `received unexpected participantConnected: ${roomSid}`);
+    });
+
+    it('emits participantConnected when a participant publishes track', async () => {
```
manjeshbhargav, line 183: @makarandp0 Can you add a test for the case where "participantConnected" is emitted for Bob when they join a Room with a track in ConnectOptions?

test/integration/spec/room.js
```diff
@@ -147,6 +147,74 @@ describe('Room', function() {
...
+      await bobRoom.localParticipant.publishTrack(track);
+      await waitFor(bobConnectedPromise, `waiting for participantConnected: ${roomSid}`);
+    });
+
+    it('returns error when published tracks increase the limit', async () => {
+      const tracksToPublish = Array(roomDetails.max_concurrent_published_tracks - 1).fill(0);
+      await waitFor(tracksToPublish.map(async () => {
+        const anotherTrack = await createLocalAudioTrack({ fake: true });
+        return bobRoom.localParticipant.publishTrack(anotherTrack);
+      }), 'max track to get published');
```
manjeshbhargav, line 195: @makarandp0 This can be simplified to:
```suggestion
      const tracks = await waitFor(tracksToPublish.map(() => createLocalAudioTrack({ fake: true })));
      await waitFor(room.localParticipant.publishTracks(tracks), 'max Tracks to get published');
```

test/integration/spec/room.js
```diff
@@ -147,6 +147,74 @@ describe('Room', function() {
...
+    });
+
+    it('emits participantConnected when a participant publishes track', async () => {
+      const bobConnectedPromise = new Promise(resolve => aliceRoom.on('participantConnected', resolve));
+
+      await bobRoom.localParticipant.publishTrack(track);
+      await waitFor(bobConnectedPromise, `waiting for participantConnected: ${roomSid}`);
+    });
+
+    it('returns error when published tracks increase the limit', async () => {
```
manjeshbhargav, line 190: @makarandp0 We can generalize this test case as follows:

* `MAX_TRACKS - x` tracks are published.
* Attempted to publish `x + 1` tracks.
* Expected behavior: `x + 1` publication promises to fail with the appropriate error.

With this test, you can now have a test case for `x = 0` (your current test) and for `x = 2` (for example).

manjeshbhargav (changes requested):

test/lib/post.js
```diff
@@ -35,11 +35,25 @@ function request(config, data) {
...
 
+/**
+ * Make an HTTP(S) get request.
+ * @param {*} config
+ * @returns {Promise<*>}
+ */
+function get(config) {
+  return request(Object.assign({
+    method: 'GET'
+  }, config));
```
charliesantos, line 53: nit: I noticed we're still using `Object.assign(..` everywhere. Should we start using spread operator?
```js
  return request({
    method: 'GET',
    ...config,
  });
```

makarandp0: I think we should start using spread (and lot more) once we start moving to typescript. For now I will keep Object.assign to be consistant with rest of the file.

test/integration/spec/room.js
makarandp0: not sure I completely understand your suggestion, but I have updated the code to read it more generic. Note that these tests share same `before` and do not create new room, so in this `it` we will have one track already published.

test/integration/spec/room.js
makarandp0: sure, added a new case now.

makarandp0:
Thanks @manjeshbhargav and @charliesantos for your feedback. I have incorporated your feedback, can you take another look?

charliesantos (approved):

PikaJoyce (approved):
LGTM!

makarandp0:
@manjeshbhargav - I have added a new test as you suggested https://github.com/twilio/twilio-video.js/pull/1341/commits/62f02d1dda553c9432793f93aeadfcd2bf3500dd - Please take a look.

test/integration/spec/room.js
```diff
@@ -713,6 +702,130 @@ describe('Room', function() {
...
+        const bobConnectedPromise = new Promise(resolve => aliceRoom.once('participantConnected', resolve));
+        const bobRoom = await connect(getToken('Bob'), Object.assign({ name: roomSid }, defaults, { tracks: [] }));
+
+        // it does not invoke participantConnected.
+        await waitForNot(bobConnectedPromise, `received unexpected participantConnected: ${roomSid}`);
+
+        // bob publishes max_concurrent_published_tracks - nTracksToPublishLater + 1
+        const nTracksToPublishInitially = roomDetails.max_concurrent_published_tracks - nTracksToPublishLater + 1;
+        const tracksToPublishInitially = await waitFor(Array(nTracksToPublishInitially).fill(0).map(() => createLocalAudioTrack({ fake: true })), `creating ${nTracksToPublishInitially} tracks`);
+        await waitFor(bobRoom.localParticipant.publishTracks(tracksToPublishInitially), `initial ${nTracksToPublishInitially} tracks to get published: ${roomSid}`);
```
manjeshbhargav, line 809: @makarandp0 You can pass the `nTracksToPublishInitially` tracks in ConnectOptions instead of publishing after joining the Room.

manjeshbhargav (approved):

PR #1346, opened 2021-01-21
JSDK-3045 Framework Tests

PikaJoyce:
Ticket : [JSDK-3045](https://issues.corp.twilio.com/browse/JSDK-3045)

This ticket is a small fix for the no-frame and react framework tests. The fix is just so it can take in the environment variable other than prod so it does not throw errors when we use credentials from different environments.

This PR does not include a fix for the angular test.

**Contributing to Twilio**


package.json
```diff
@@ -121,7 +121,7 @@
     "test:framework:react:run": "mocha ./test/framework/twilio-video-react.js",
     "test:framework:react": "npm-run-all test:framework:react:*",
     "test:framework:install": "npm install chromedriver && npm install selenium-webdriver && npm install geckodriver && npm install puppeteer",
-    "test:framework": "npm-run-all test:framework:install test:framework:angular test:framework:no-framework test:framework:react",
```
charliesantos, line 124: Why are we removing angular tests?

.circleci/config.yml
```diff
@@ -143,12 +143,9 @@ commands:
   framework-tests:
     steps:
       - build
-      - run:
-          name: Installing dependencies
-          command: npm install
       - run:
           name: Running Framework Tests
-          command: npm run test:framework
+          command: scripts/circleci-run-tests.sh
```
charliesantos, line 148: Just curious, why can't we run framework tests directly from here?

charliesantos (commented):
Sorry if these has already been discussed. I left a couple of questions.

.circleci/config.yml
makarandp0: because framework tests require account creds, and the creds are setup in `scripts/circleci-run-tests.sh` (they are different for stage / prod / dev)

makarandp0 (approved):

.circleci/config.yml
PikaJoyce: @charliesantos As Makarand said, we were still hitting credential errors when running from npm script even with the changes. So we had to do what we did for integration tests in order for us to pass in stage/dev creds.

package.json
PikaJoyce: After spending some time with angular tests, I've added the work to address the angular test failures in a future ticket here: [JSDK-3110](https://issues.corp.twilio.com/browse/JSDK-3110). I think fixing the angular tests would be impossible unless we updated some of the packages which caused problems for other older packages as well.

PR #1352, opened 2021-01-21
JSDK-3129 | Fixing loglevel module to work with browserify and requirejs

charliesantos:
**Contributing to Twilio**

The loglevel npm module is handling CommonJS, AMD, and browser environments all in the same file. Which doesn't work if the module is bundled with browserify, and then later used with RequireJS. To fix this issue, the loglevel module should update its build process to output AMD/Browser and CommonJS files separately. Unfortunately, this is a common mistake by some developers. See this discussion and some links to requirejs docs [here](https://stackoverflow.com/questions/15371918/mismatched-anonymous-define-module.). Summary: If using requirejs, all files should be included using requirejs. If using CommonJS, building should only include CommonJS (this is the one that needs to be publish to npm). If you need to support both, you should build for both.

To address this issue on our side, I included the loglevel file directly and only added CommonJS support since that's the only scenario we need it for. The browser/AMD use case is handled by our SDK's build process already.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,6 +149,16 @@ commands:
       - run:
           name: Running Framework Tests
           command: npm run test:framework
+      - save-test-results
+  umd-tests:
+    steps:
+      - build
+      - run:
+          name: Installing dependencies
+          command: npm install
+      - run:
+          name: Installing UMD test dependencies
+          command: npm run test:umd:install
       - run:
           name: Running UMD Tests
           command: npm run test:umd
@@ -222,6 +232,11 @@ jobs:
       BROWSER: "chrome"
     executor: docker-with-browser
     steps: [framework-tests]
+  UMD-tests:
+    environment:
+      BROWSER: "chrome"
+    executor: docker-with-browser
+    steps: [umd-tests]
   Build-Docker-Test-Image:
     executor: machine-executor
     parameters:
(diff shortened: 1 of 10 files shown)


makarandp0:
@charliesantos, I am wondering if not exposing `Logger` from twilio-video is better option, and let developers require('loglevel') separately ? I know it will be breaking change but asking anyways.

While working with typescript I had to get types for loglevel [separately](https://github.com/makarandp0/twilio-video-api/blob/main/src/createRoomControls.ts#L13) for usage like [this](https://github.com/makarandp0/twilio-video-api/blob/main/src/createRoomControls.ts#L49), I wonder how would that work if we got the folded in the implementation of loglevel.

charliesantos:
> @charliesantos, I am wondering if not exposing `Logger` from twilio-video is better option, and let developers require('loglevel') separately ? I know it will be breaking change but asking anyways.
> 
> While working with typescript I had to get types for loglevel [separately](https://github.com/makarandp0/twilio-video-api/blob/main/src/createRoomControls.ts#L13) for usage like [this](https://github.com/makarandp0/twilio-video-api/blob/main/src/createRoomControls.ts#L49), I wonder how would that work if we got the folded in the implementation of loglevel.

@makarandp0 Whether we expose `Logger` from twilio-video or not, the issue with RequireJS still exists. It's because the loglevel build process is not setup properly. Additionally, not exposing it meaning some features such as the plugin system will not work, which is the main reason why we added this module in the first place. Also, loglevel has its own definitions already and you shouldn't have to get them, but I don't think module was setup properly. In this PR, I added the definitions as well.

makarandp0 (approved):
Thanks @charliesantos, Changes LGTM!

PR #1382, opened 2021-02-11
VIDEO-3561 | removeProcessor API

charliesantos:
**Contributing to Twilio**

This PR:
- Implemented removeProcessor API
- Handle track.restart()
- Fixed an issue where processedTrack is not sent to RTCRtpSender after disable/enable
- Added checks before and during capture process whether capturing frames should proceed/stop or not. See `canCaptureFrames()` method
- Do not stop capturing frames on `mute`. This is because the track is automatically muted by the browser when not visible or looking at a different tab. In this case, we don't want to stop capturing frames.
- Stopped using `requestVideoFrameCallback` because this doesn't trigger when the page is not visible. For example, if the user is looking at a different tab. We don't want to stop capturing frames in this case. But we may want to figure out the correct timeout period, depending on the VideoProcessor performance.

### DIFF
--- a/lib/media/track/localvideotrack.js
+++ b/lib/media/track/localvideotrack.js
@@ -39,6 +39,9 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
     super(mediaStreamTrack, options);
 
     Object.defineProperties(this, {
+      _unprocessedTrack: {
+        value: mediaStreamTrack
+      },
       _workaroundSilentLocalVideo: {
         value: options.workaroundSilentLocalVideo
           ? workaroundSilentLocalVideo
@@ -67,6 +70,15 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
    */
   _end() {
     return super._end.apply(this, arguments);
+  }
+
+  /**
+   * @private
+   */
+  _setSenderMediaStreamTrack(mediaStreamTrack) {
+    return this._trackSender.setMediaStreamTrack(mediaStreamTrack).catch(error =>
+      this._log.warn('setMediaStreamTrack failed on LocalVideoTrack RTCRtpSender',
+        { error, mediaStreamTrack }));
   }
 
   /**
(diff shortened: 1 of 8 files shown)


lib/media/track/localvideotrack.js
```diff
@@ -129,17 +176,24 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
+   * or "pause". If a {@link VideoProcessor} is added,
+   * then `processedTrack` is enabled or disabled as well.
    * @param {boolean} [enabled] - Specify false to pause the
    *   {@link LocalVideoTrack}
    * @returns {this}
    * @fires VideoTrack#disabled
    * @fires VideoTrack#enabled
    */
-  enable() {
+  enable(enabled) {
```
manjeshbhargav, line 187:
```suggestion
  enable(enabled = true) {
```

lib/media/track/localvideotrack.js
```diff
@@ -129,17 +176,24 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
     if (this.processedTrack) {
-      this.processedTrack.enabled = true;
-      this._captureFrames();
+      enabled = typeof enabled === 'boolean' ? enabled : true;
```
manjeshbhargav, line 189:
```suggestion
```

lib/media/track/localvideotrack.js
```diff
@@ -39,6 +39,9 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
     super(mediaStreamTrack, options);
 
     Object.defineProperties(this, {
+      _unprocessedTrack: {
```
manjeshbhargav, line 42: Why do we need this? Doesn't `this.mediaStreamTrack` point to the unprocessed track?

lib/media/track/localvideotrack.js
```diff
@@ -183,8 +237,17 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
+    const processor = this.processor;
+    if (processor) {
+      this.removeProcessor(processor);
+    }
     const promise = super.restart.apply(this, arguments);
 
+    if (processor) {
+      promise.then(() => this.addProcessor(processor));
+    }
+
```
manjeshbhargav, line 250: This means that while restart is under process, `LocalVideoTrack.processor` and `LocalVideoTrack.processedTrack` is null. Instead of removing and adding the processor, can we stop and start capture?

lib/media/track/videotrack.js
```diff
@@ -84,18 +109,15 @@ class VideoTrack extends MediaTrack {
 
     this._dummyEl.play().then(() => {
       const captureFrame = cb => {
-        const { enabled, muted, readyState } = this.mediaStreamTrack;
-        if (!enabled || muted || readyState === 'ended') {
-          return;
-        }
-        if (this._dummyEl.requestVideoFrameCallback) {
-          this._dummyEl.requestVideoFrameCallback(cb);
```
manjeshbhargav, line 92: Are you sure `requestVideoFrameCallback` does not execute when the tab is not in view? The spec did not mention anything about this.

lib/media/track/localvideotrack.js
charliesantos: Update mediatrack.js to save the unprocesstrack

lib/media/track/localvideotrack.js
charliesantos: Document this behavior. Cut a ticket to change this behavior later, maybe during beta.

lib/media/track/videotrack.js
charliesantos: Subtrack the processFrame delay for each call.

manjeshbhargav (approved):

Issue #963, opened 2020-04-21
Client is unable to create or apply a local media description


dambusm:
Hi all,

For a while now we've been seeing quite a lot of reports of this error:
```
TwilioError: Client is unable to create or apply a local media description
  at TwilioError(../node_modules/twilio-video/es5/util/twilioerror.js:37:23)
  at new t(../node_modules/twilio-video/es5/util/twilio-video-errors.js:1031:147)
  at ? (../node_modules/twilio-video/es5/signaling/v2/peerconnection.js:519:75)
```

It seems to have started when we upgraded from twilio-video 2.0.0 to 2.0.1. We're now on 2.3.0 and it's still occurring. It's happening on Edge 80.0.361 and on Chrome 80.0.3987.

I hadn't been able to reproduce it, but today I had it happen on my machine. It caused the call to have no video or audio, although the Twilio room (`"RoomSid": "RM5f51ab5a21407e4858f8918b8b3d7d7a"`) reports tracks added from both participants. When I refreshed the page, everything connected and worked fine.

Console output:

```js
2020-04-21 09:15:29.613Z | WARN in [PeerConnectionV2 #1: 51ada9e3-b3e6-468a-a6c8-5e6185da9dde]: Failed to add RTCIceCandidate "candidate:211156821 1 udp 2122260223 192.168.1.5 64547 typ host generation 0 ufrag HToW network-id 1 network-cost 10": Failed to execute 'addIceCandidate' on 'RTCPeerConnection': Error processing ICE candidate
log @ VM207 1.chunk.js:309571
220455:1 Uncaught (in promise) TwilioError: Client is unable to create or apply a local media description
    at MediaClientLocalDescFailedError.TwilioError (http://localhost:3000/static/js/1.chunk.js:313170:23)
    at new MediaClientLocalDescFailedError (http://localhost:3000/static/js/1.chunk.js:312871:147)
    at http://localhost:3000/static/js/1.chunk.js:298596:75
    at async Promise.all (index 0)
    at async Promise.all (index 0)
[...]

PR #1415, opened 2021-03-10
Fix unresolved Promise getting track publication

a-type:
This small fix addresses a theoretical scenario where the Promise returned from `_getOrCreateLocalTrackPublication` might never be resolved or rejected.

I'm not sure of the relevance of the scenario where `trackSignaling.sid` would be falsy, so I just provided a generic error message. A maintainer may want to provide a more helpful contextual error here.

Rejecting the Promise before calling `return;` is vital for all codepaths in a Promise constructor, or else any code awaiting the Promise will never resume.

Although this fix may possibly address #963, it's hard to say for certain without some sort of isolated reproduction, which has been hard to do as it seems to crop up mostly in response to changes in device power state for me. #963 still might not be fixed by this change if it turns out that the `'updated'` event is simply never fired from `trackSignaling` (which would also cause this Promise to never resolve, as all resolution paths are tied to that event).

**Contributing to Twilio**

### DIFF
--- a/lib/localparticipant.js
+++ b/lib/localparticipant.js
@@ -294,6 +294,7 @@ class LocalParticipant extends Participant {
 
         const sid = trackSignaling.sid;
         if (!sid) {
+          reject(new Error(`No SID for track publication for ${localTrack}`));
           return;
         }
 


manjeshbhargav:
Hi @a-type ,

Thanks for the PR. While I agree in general about always resolving/rejecting before `return`, in this case, it is simply a bit of defensive programming as the SID is never actually null. However, we will keep this in mind and update accordingly.

Thanks,

Manjesh

a-type:
With respect, I appreciate the fact that it's purely defensive, but I'd point out that defensive programming which results in an indefinite Promise hang isn't really defensive. It's probably preferable to delete the branch entirely if you think the scenario is logically impossible. Then if the impossible happens, we'll at least have a runtime error which should reject the Promise.

PR #1418, opened 2021-03-11
video_3739: use ResizeObserver to report video element sizes.

makarandp0:
This change adds ResizeObserver - RemoteVideoTrack now watches the video element as they resize and report the size to VMS.

- [x] implementation
- [x] unit tests
- [ ] integration tests

- [x]  I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/lib/media/track/remotevideotrack.js
+++ b/lib/media/track/remotevideotrack.js
@@ -3,7 +3,7 @@
 const mixinRemoteMediaTrack = require('./remotemediatrack');
 const VideoTrack = require('./videotrack');
 const documentVisibilityMonitor = require('../../util/documentvisibilitymonitor.js');
-const NullIntersectionObserver = require('../../util/nullintersectionobserver.js');
+const { NullIntersectionObserver, NullResizeObserver } = require('../../util/nullobserver.js');
 
 const RemoteMediaVideoTrack = mixinRemoteMediaTrack(VideoTrack);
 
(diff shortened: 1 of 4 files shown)


lib/media/track/remotevideotrack.js
```diff
@@ -57,6 +58,17 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
       _invisibleElements: {
         value: new WeakSet(),
       },
+      _resizeObserver: {
+        value: new options.ResizeObserver(entries => {
+          // NOTE(mpatwardhan): we ignore elements in _invisibleElements
+          // to ensure that ResizeObserver does not end-up turning off a track when a fresh Video element is
+          // attached and IntersectionObserver has not had its callback executed yet.
+          const visibleElementResized = entries.find(entry => !this._invisibleElements.has(entry.target));
```
manjeshbhargav, line 66: @makarandp0 Is it possible for the resize observer to fire events for invisible elements? I don't think invisible elements can be resized, no? Unless you are talking about video elements without fixed widths and heights.

manjeshbhargav (approved):

PikaJoyce (approved):

charliesantos (approved):

lib/media/track/remotevideotrack.js
makarandp0: @manjeshbhargav, Yes, ResizeObserver does get fire its callback even when the element is not visible. I found that by changing dimensions of the elements were not visible in the scrollView. 
This check is most obvious  for freshly added elements for which resizeObserver gets called before interSectionObserver.

PR #1410, opened 2021-03-05
Fix publishTrack types

ksocha:
According to the [docs](https://media.twiliocdn.com/sdk/js/video/releases/2.13.0/docs/LocalParticipant.html), `publishTrack` method should accept different options depending on the first argument's type.

Currently, it does only accept `LocalTrackPublishOptions` and not `MediaStreamTrackPublishOptions` if mediaTrack was passed. This PR fixes the issue.

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #1419, opened 2021-03-13
Adding MediaStreamTrackOptions to the LocalParticipant publishTrack m…

PikaJoyce:
JIRA Ticket: [VIDEO-4460](https://issues.corp.twilio.com/browse/VIDEO-4460)

This is a PR which addresses an [issue](https://github.com/twilio/twilio-video.js/pull/1410) with our Type definitions. I have since added `MediaStreamTrackOptions` to the `publishTrack` method on the `LocalParticipant`. I have also double checked that the shape of `MediaStreamTrackOptions` should match the shape of `LocalTrackPublishOptions` so there is a fix for that as well.

**Contributing to Twilio**

### DIFF
--- a/tsdef/LocalParticipant.d.ts
+++ b/tsdef/LocalParticipant.d.ts
@@ -1,4 +1,4 @@
-import { EncodingParameters, LocalTrack, LocalTrackPublishOptions, NetworkQualityConfiguration } from './types';
+import { EncodingParameters, LocalTrack, LocalTrackPublishOptions, MediaStreamTrackPublishOptions, NetworkQualityConfiguration } from './types';
 import { LocalAudioTrackPublication } from './LocalAudioTrackPublication';
 import { LocalDataTrackPublication } from './LocalDataTrackPublication';
 import { LocalTrackPublication } from './LocalTrackPublication';
@@ -15,7 +15,7 @@ export class LocalParticipant extends Participant {
   videoTracks: Map<Track.SID, LocalVideoTrackPublication>;
   signalingRegion: string;
 
-  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions): Promise<LocalTrackPublication>;
+  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions | MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
   publishTracks(tracks: Array<LocalTrack | MediaStreamTrack>): Promise<LocalTrackPublication[]>;
   setNetworkQualityConfiguration(networkQualityConfiguration: NetworkQualityConfiguration): this;
   setParameters(encodingParameters?: EncodingParameters | null): this;
--- a/tsdef/LocalParticipant.d.ts
+++ b/tsdef/LocalParticipant.d.ts
@@ -15,7 +15,8 @@ export class LocalParticipant extends Participant {
   videoTracks: Map<Track.SID, LocalVideoTrackPublication>;
   signalingRegion: string;
 
-  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions | MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
+  publishTrack(track: LocalTrack, options?: LocalTrackPublishOptions): Promise<LocalTrackPublication>;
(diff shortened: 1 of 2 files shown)


charliesantos (approved):

tsdef/LocalParticipant.d.ts
```diff
@@ -15,7 +15,7 @@ export class LocalParticipant extends Participant {
   videoTracks: Map<Track.SID, LocalVideoTrackPublication>;
   signalingRegion: string;
 
-  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions): Promise<LocalTrackPublication>;
+  publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions | MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
```
timmydoza, line 18: Instead of:
```
publishTrack(track: LocalTrack | MediaStreamTrack, options?: LocalTrackPublishOptions | MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
```

Should it be?:
```
publishTrack(track: LocalTrack, options?: LocalTrackPublishOptions ): Promise<LocalTrackPublication>;
publishTrack(track: MediaStreamTrack, options?: MediaStreamTrackPublishOptions): Promise<LocalTrackPublication>;
```


With the second snippet - you can only pass `MediaStreamTrackPublishOptions` when publishing a `MediaStreamTrack`, and you can only pass `LocalTrackPublishOptions`, when publishing a `LocalTrack`.

The first snippet allows users to pass a `MediaStreamTrackPublishOptions` object when publishing a `LocalTrack`, and to pass a `LocalTrackPublishOptions` object when publishing a `MediaStreamTrack`. I don't think this should be allowed. I think the second snippet above is more consistent with our documentation. 

What do you think?

timmydoza (changes requested):

tsdef/LocalParticipant.d.ts
PikaJoyce: Thanks Tim, I think that's a great idea and it makes total sense. I will make those changes now :)

tsdef/types.d.ts
```diff
@@ -148,7 +148,6 @@ export interface LocalTrackPublishOptions {
 }
 
 export interface MediaStreamTrackPublishOptions {
-  name?: string;
```
timmydoza, line 151: @PikaJoyce I think we should keep this value. It's in the docs. Plus, without it, `MediaStreamTrackPublishOptions` and `LocalTrackPublishOptions` would be identical. 

What do you think?

PikaJoyce: @timmydoza, I removed this because the documentation for [MediaStreamTrackPublishOptions](https://media.twiliocdn.com/sdk/js/video/releases/2.13.0/docs/global.html#MediaStreamTrackPublishOptions) does not have a name property. Which is why I removed it. Is the documentation incorrect? If it is, I can change the documentation for `MediaStreamTrackPublishOptions` and put the name property back in. 

cc: @manjeshbhargav

manjeshbhargav: @timmydoza @PikaJoyce Yes, `MediaStreamTrackPublishOptions` should have the `name` property. Please correct the JS documentation to include it.

timmydoza: @PikaJoyce - Looking closer, there isn't a `name` property on the `MediaStreamTrackPublishOptions` type itself, but there is a `name` property on the `LocalTrackOptions` type that `MediaStreamTrackPublishOptions` inherits from. So maybe something like this?

```
export interface MediaStreamTrackPublishOptions extends LocalTrackOptions {
    priority?: Track.Priority;
}
```

This looks like its consistent with the docs: https://media.twiliocdn.com/sdk/js/video/releases/2.13.0/docs/global.html#MediaStreamTrackPublishOptions

manjeshbhargav: @PikaJoyce Looks like you don't need to create the documentation ticket after all! 🎉

PikaJoyce: @timmydoza Oh yay! Thanks for finding that. For some odd reason, I kept reading that `MediaStreamTrackPublishOptions` inherits from `LocalTrackPublishOptions` and not `LocalTrackOptions`. That was  my mistake, I'm sorry. I'll revert the change and add in your suggestion :)

PikaJoyce: Also it looks like I didn't even commit the JSDoc change so there is conveniently nothing for me to change on that end!

manjeshbhargav (approved):

PR #1422, opened 2021-03-15
VIDEO-4383 connect options for render hints

makarandp0:
This PR adds `renderHints` connectOptions property and plumbing to pass the property to `RemoteVideoTrack`
also updated unit test against on/off state of various options, added integration tests for `idleTrackSwitchOff` property

**Contributing to Twilio**

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -284,8 +284,8 @@ function connect(token, options) {
   }
 
 
-  // Note(mpatwardhan): idleTrackSwitchOff and renderHints options are to be Opt-out
-  //  they are enabled by default bandwidth profile is enabled
+  // Note(mpatwardhan): "idleTrackSwitchOff" and "renderHints" are to be Opt-out options
+  //  if "bandwidthProfile" is enabled
   options.idleTrackSwitchOff =  false;
   options.renderHints =  false;
   if (options.bandwidthProfile) {
@@ -469,9 +469,10 @@ function connect(token, options) {
  * @property {boolean} [idleTrackSwitchOff] - Optional parameter that when enabled switches off a
  *    {@link RemoteVideoTrack} when no video element is attached to the track, or when all attached video
  *    elements of the track are not visible, or when the Document is not visible. This is enabled by default
- * @property {boolean} [renderHints] - Optional parameter that when enabled allows the media server
- *   to send video with bitrate and dimensions based on the render dimensions of the video
- *   elements attached to {@link RemoteVideoTrack}. This is enabled by default
+ *    if <code>bandwidthProfile</code> is specified in {@link ConnectOptions}.
+* @property {boolean} [renderHints] -  Optional parameter that when enabled regulates the bandwidth and
+ *    quality of a {@link RemoteVideoTrack} based on the dimensions of the largest <video> element attached to it.
+ *    This is enabled by default if <code>bandwidthProfile</code> is specified in {@link ConnectOptions}.
  * @property {number} [maxTracks] - Optional parameter to specify the maximum number of visible
  *   {@link RemoteVideoTrack}s, which will be selected based on {@link Track.Priority} and an N-Loudest
  *   policy. By default there are no limits on the number of visible {@link RemoteVideoTrack}s.
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -284,12 +284,19 @@ function connect(token, options) {
   }
 
 
+  // Note(mpatwardhan): idleTrackSwitchOff and renderHints options are to be Opt-out
+  //  they are enabled by default bandwidth profile is enabled
   options.idleTrackSwitchOff =  false;
+  options.renderHints =  false;
   if (options.bandwidthProfile) {
     options.idleTrackSwitchOff = true;
+    options.renderHints = true;
     if (options.bandwidthProfile.video) {
       if (options.bandwidthProfile.video.idleTrackSwitchOff === false) {
         options.idleTrackSwitchOff = false;
+      }
+      if (options.bandwidthProfile.video.renderHints === false) {
+        options.renderHints = false;
       }
     }
   }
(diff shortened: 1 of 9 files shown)


lib/connect.js
```diff
@@ -284,13 +284,20 @@ function connect(token, options) {
   }
 
 
+  // Note(mpatwardhan): idleTrackSwitchOff and renderHints options are to be Opt-out
+  //  they are enabled by default bandwidth profile is enabled
```
manjeshbhargav, line 288:
```suggestion
  // Note(mpatwardhan): "idleTrackSwitchOff" and "renderHints" are opt-out options
  // if "bandwidthProfile" is enabled.
```

lib/connect.js
```diff
@@ -460,8 +467,11 @@ function connect(token, options) {
  *   downlink video bandwidth in bits per second (bps). By default, there are no limits on
  *   the downlink video bandwidth.
  * @property {boolean} [idleTrackSwitchOff] - Optional parameter that when enabled switches off a
- *    RemoteVideoTrack when no video element is attached to the track, or when all attached video
+ *    {@link RemoteVideoTrack} when no video element is attached to the track, or when all attached video
  *    elements of the track are not visible, or when the Document is not visible. This is enabled by default
```
manjeshbhargav, line 471: Please mention that this is applicable to Group and Small Group Rooms only.

lib/connect.js
```diff
@@ -460,8 +467,11 @@ function connect(token, options) {
...
+ * @property {boolean} [renderHints] - Optional parameter that when enabled allows the media server
+ *   to send video with bitrate and dimensions based on the render dimensions of the video
+ *   elements attached to {@link RemoteVideoTrack}. This is enabled by default
```
manjeshbhargav, line 474:
```suggestion
 * @property {boolean} [renderHints] - Optional parameter that when enabled regulates the bandwidth and
      quality of a {@link RemoteVideoTrack} based on the dimensions of the largest <video> element attached to it.
      This is enabled by default if <code>bandwidthProfile</code> is specified in {@link ConnectOptions}.
```

lib/media/track/remotevideotrack.js
```diff
@@ -106,8 +110,13 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
...
     this._invisibleElements.add(result);
-    this._intersectionObserver.observe(result);
-    this._resizeObserver.observe(result);
+    if (this._idleTrackSwitchOff) {
+      this._intersectionObserver.observe(result);
+    }
+
+    if (this._renderHints) {
+      this._resizeObserver.observe(result);
+    }
```
manjeshbhargav, line 119: I don't think we need the `if()` check since you are using the `NullObserver`.

lib/media/track/remotevideotrack.js
```diff
@@ -125,8 +134,14 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
...
-      this._intersectionObserver.unobserve(element);
-      this._resizeObserver.unobserve(element);
+      if (this._idleTrackSwitchOff) {
+        this._intersectionObserver.unobserve(element);
+      }
+
+      if (this._renderHints) {
+        this._resizeObserver.unobserve(element);
+      }
+
```
manjeshbhargav, line 144: I don't think we need the `if()` check since you are using the `NullObserver`.

lib/media/track/remotevideotrack.js
```diff
@@ -247,19 +262,24 @@ function setupDocumentVisibilityTurnOff(removeVideoTrack) {
  * updates render hints
  */
 function updateRenderHints(removeVideoTrack) {
-  if (!removeVideoTrack._idleTrackSwitchOff) {
+  if (!removeVideoTrack._idleTrackSwitchOff && !removeVideoTrack._renderHints) {
```
manjeshbhargav, line 265: Why is this check required if you are using `NullObserver`?

manjeshbhargav (changes requested):

lib/connect.js
makarandp0: These two are properties of VideoBandwidthProfileOptions, which itself is group room specific.  We do specify that  `BandwidthProfileOptions` are group room specific on line 447

lib/media/track/remotevideotrack.js
makarandp0: You are right that this check is not necessary, but I it helps with unit tests, as they can verify that observe being called only when option is set.

lib/media/track/remotevideotrack.js
makarandp0: You are right that this check is not necessary, but I it helps with unit tests, as they can verify that unobserve being called only when option is set.

lib/media/track/remotevideotrack.js
makarandp0: With NullObserver the observer callbacks are covered. but this  check is required because we call also this function outside the observe callbacks.  For example this function is called from RemoteVideoTrack.start and RemoteVideoTrack.detach.

lib/media/track/remotevideotrack.js
manjeshbhargav: @makarandp0 I think the unit tests can test if the correct callback (actual or fake) is called. That way, you can remove these checks.

lib/media/track/remotevideotrack.js
manjeshbhargav: @makarandp0 I think the unit tests can test if the correct callback (actual or fake) is called. That way, you can remove these checks.

lib/media/track/remotevideotrack.js
makarandp0: the unit test and the implementation both use same prototype (NullObserver) for fake, making it difficult to distinguish. maybe we can quick chat about whats the best way to do this.

manjeshbhargav: @makarandp0 Since you depend on the global `IntersectionObserver` and `ResizeObserver`, you can use two fake observers, one to verify the case where they are supported and one to verify the case where they are not supported.

```js
// Supported case
before(() => {
  global.IntersectionObserver = FakeObserver;
});

it('...', () => {
  assert(/* FakeObserver constructor is called when the RemoteVideoTrack is created */);
});

// Unsupported case
before(() => {
  delete global.IntersectionObserver;
});

it('...', () => {
  assert(/* NullObserver constructor is called when the RemoteVideoTrack is created */);
});
```

This way, you can assert that the callback is called in both cases.

lib/signaling/v2/room.js
```diff
@@ -433,7 +433,7 @@ class RoomV2 extends RoomSignaling {
   _initTrackSwitchOffSignaling() {
     this._trackSwitchOffSignaling.on('updated', (tracksOff, tracksOn) => {
       try {
-        this._log.warn('received trackSwitch: ', { tracksOn, tracksOff });
+        this._log.info('received trackSwitch: ', { tracksOn, tracksOff });
```
manjeshbhargav, line 436:
```suggestion
        this._log.debug('received trackSwitch: ', { tracksOn, tracksOff });
```

manjeshbhargav (changes requested):

PikaJoyce (approved):
Yay, great work Makarand!

manjeshbhargav (approved):

PR #1431, opened 2021-03-19
video-4524 msp field rename

makarandp0:
we decided to rename two msp fields in render_hints protocol - the msp protocol change is described here: https://code.hq.twilio.com/client/room-signaling-protocol/pull/70

for render_hint protocol:
`render_dimension` => `render_dimensions`
`track_sid` => `track`

This PR updates sdk to accommodate it.  I have verified the client against dev environment. We will merge this change only after related VMS changes are in stage. (Note: circleCi job against this PR is expected fail till then) 

Along with protocol change, this also changes an internal structure from `renderDimension` => `renderDimensions`

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/renderhintssignaling.js
+++ b/lib/signaling/v2/renderhintssignaling.js
@@ -85,13 +85,13 @@ class RenderHintsSignaling extends MediaSignaling {
   setTrackHint(trackSid, renderHint) {
     // convert hint to msp format
     const mspHint = {
-      'track_sid': trackSid,
+      'track': trackSid,
       'enabled': !!renderHint.enabled,
     };
 
-    if (renderHint.renderDimension) {
+    if (renderHint.renderDimensions) {
       // eslint-disable-next-line camelcase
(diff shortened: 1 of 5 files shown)


lib/signaling/v2/renderhintssignaling.js
```diff
@@ -85,13 +85,13 @@ class RenderHintsSignaling extends MediaSignaling {
   setTrackHint(trackSid, renderHint) {
     // convert hint to msp format
     const mspHint = {
-      'track_sid': trackSid,
+      'track': trackSid,
```
charliesantos, line 88: Should we just keep this as `track_sid` since it's really a track's sid?

makarandp0: it's not really up to us. It was decided collectively to use `track` since thats what we use in rest of the MSP protocols. VMS is doing the same change on server side.

charliesantos (approved):

manjeshbhargav (approved):

PR #1455, opened 2021-04-22
VIDEO-5026 unexpected render hint on track started

makarandp0:
this fixes an issue where an empty renderhint was sent (by track started).
With this change:
 - a) do not  send any hints if neither of 'subscribedTrackSwitchOff' and 'contentPreferencesMode' set to auto
 - b) send hint after trackStarted only if  subscribedTrackSwitchOff is set to auto
Also fixed a typo in error message (VIDEO-5008)

**Contributing to Twilio**


lib/media/track/remotevideotrack.js
```diff
@@ -105,7 +105,10 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
    */
   _start(dummyEl) {
     const result = super._start.call(this, dummyEl);
-    updateRenderHints(this);
+    // after emitting started, turn off track if not needed.
```
manjeshbhargav, line 108:
```suggestion
    // NOTE(mpatwardhan): after emitting started, update render hints only if subscribedTrackSwitchOffMode is 'auto'.
```

lib/media/track/remotevideotrack.js
```diff
@@ -105,7 +105,10 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
...
+    if (this._subscribedTrackSwitchOffMode === 'auto') {
```
manjeshbhargav, line 109: Is this check necessary since you are performing the same check within `updateRenderHints()`. Another nitpick: if there is a chance that render hints are not updated when you call `updateRenderHints()`, it should be called `maybeUpdateRenderHints()`.

manjeshbhargav (changes requested):

lib/media/track/remotevideotrack.js
makarandp0: this check is necessary because `updateRenderHints` is no-op only when both `subscribedTrackSwitchOffMode != auto` and  `contentPreferencesMode != auto` 
Good Idea about renaming `maybeUpdateRenderHints` - I will make the change

manjeshbhargav (approved):
Approved after clarification by @makarandp0 for my question. Please create a ticket to refactor the render hints and subscriber track switch off  features into their own modules, which can be included or not depending on the ConnectOptions flags.

PR #1466, opened 2021-05-04
VIDEO-4718 Adding A/V sync metrics, elapsed time and level to "stats-report" insights message.

manjeshbhargav:
@makarandp0 @PikaJoyce 

Adds A/V sync metrics ([example](https://kibana.us1.eak.twilio.com/video/goto/269fed43d045922091455a100a0a7d0a)) to insights, as defined [here](https://issues.corp.twilio.com/browse/VIDEO-4718).

**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -15,11 +15,13 @@ const {
   constants: { DEFAULT_SESSION_TIMEOUT_SEC },
   createBandwidthProfilePayload,
   defer,
+  difference,
   filterObject,
   flatMap,
   oncePerTick
 } = require('../../util');
 
+const MovingAverageDelta = require('../../util/movingaveragedelta');
 const { createTwilioError } = require('../../util/twilio-video-errors');
 
 const STATS_PUBLISH_INTERVAL_MS = 10000;
@@ -655,6 +657,7 @@ function handleTransportEvents(roomV2, transport) {
  * @param {Number} intervalMs
  */
 function periodicallyPublishStats(roomV2, transport, intervalMs) {
+  const movingAverageDeltas = new Map();
   let oddPublishCount = false;
   const interval = setInterval(() => {
     roomV2.getStats().then(stats => {
@@ -665,22 +668,35 @@ function periodicallyPublishStats(roomV2, transport, intervalMs) {
         // StatsReport will add zeros to properties that do not exist.
         const report = new StatsReport(id, response, true /* prepareForInsights */);
 
+        // NOTE(mmalavalli): Since A/V sync metrics are not part of the StatsReport class,
+        // we add them to the insights payload here.
         transport.eventObserver.emit('event', {
           group: 'quality',
           level: 'info',
           name: 'stats-report',
           payload: {
             audioTrackStats: report.remoteAudioTrackStats.map((trackStat, i) =>
-              addAVSyncMetricsToRemoteTrackStats(trackStat, response.remoteAudioTrackStats[i])),
+              addAVSyncMetricsToRemoteTrackStats(trackStat, response.remoteAudioTrackStats[i], movingAverageDeltas)),
             localAudioTrackStats: report.localAudioTrackStats.map((trackStat, i) =>
-              addAVSyncMetricsToLocalTrackStats(trackStat, response.localAudioTrackStats[i])),
+              addAVSyncMetricsToLocalTrackStats(trackStat, response.localAudioTrackStats[i], movingAverageDeltas)),
             localVideoTrackStats: report.localVideoTrackStats.map((trackStat, i) =>
-              addAVSyncMetricsToLocalTrackStats(trackStat, response.localVideoTrackStats[i])),
+              addAVSyncMetricsToLocalTrackStats(trackStat, response.localVideoTrackStats[i], movingAverageDeltas)),
             peerConnectionId: report.peerConnectionId,
             videoTrackStats: report.remoteVideoTrackStats.map((trackStat, i) =>
-              addAVSyncMetricsToRemoteTrackStats(trackStat, response.remoteVideoTrackStats[i])),
+              addAVSyncMetricsToRemoteTrackStats(trackStat, response.remoteVideoTrackStats[i], movingAverageDeltas)),
           }
         });
+
+        // NOTE(mmalavalli): Clean up entries for Tracks that are no longer published or subscribed to.
+        const movingAverageDeltaTrackSids = Array.from(movingAverageDeltas.keys());
+        const trackSids = flatMap([
+          'localAudioTrackStats',
+          'localVideoTrackStats',
+          'remoteAudioTrackStats',
+          'remoteVideoTrackStats'
+        ], prop => report[prop].map(({ trackSid }) => trackSid));
+        const movingAverageDeltaTrackSidsToBeRemoved = difference(movingAverageDeltaTrackSids, trackSids);
+        movingAverageDeltaTrackSidsToBeRemoved.forEach(trackSid => movingAverageDeltas.delete(trackSid));
 
         if (oddPublishCount) {
           // NOTE(mmalavalli): null properties of the "active-ice-candidate-pair"
@@ -733,7 +749,12 @@ function handleSubscriptions(room) {
   });
 }
 
-function addAVSyncMetricsToLocalTrackStats(trackStats, trackResponse) {
+/**
+ * NOTE(mmalavalli): Since A/V sync metrics are not part of the public StatsReport class, we add them
+ * only for reporting purposes.
+ * @private
+ */
+function addAVSyncMetricsToLocalTrackStats(trackStats, trackResponse, movingAverageDeltas) {
   const {
     framesEncoded,
     packetsSent,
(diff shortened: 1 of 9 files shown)


lib/signaling/v2/room.js
```diff
@@ -665,12 +665,21 @@ function periodicallyPublishStats(roomV2, transport, intervalMs) {
...
-          localVideoTrackStats: report.localVideoTrackStats,
-          peerConnectionId: report.peerConnectionId,
-          videoTrackStats: report.remoteVideoTrackStats
+        transport.eventObserver.emit('event', {
+          group: 'quality',
+          level: 'info',
+          name: 'stats-report',
+          payload: {
+            audioTrackStats: report.remoteAudioTrackStats.map((trackStat, i) =>
+              addAVSyncMetricsToRemoteTrackStats(trackStat, response.remoteAudioTrackStats[i])),
```
makarandp0, line 674: @manjeshbhargav, instead of adding the AVSyncMetric after the report is generated, Why not add this in `StatsReport` itself so that you don't have to patch it later?

lib/signaling/v2/room.js
```diff
@@ -719,6 +733,58 @@ function handleSubscriptions(room) {
...
+    framesEncoded,
+    packetsSent,
+    totalEncodeTime,
+    totalPacketSendDelay
+  } = trackResponse;
+  const augmentedTrackStats = Object.assign({}, trackStats);
+
+  if (typeof totalEncodeTime === 'number' && typeof framesEncoded === 'number') {
+    const totalEncodeTimeMs = totalEncodeTime * 1000;
+    augmentedTrackStats.avgEncodeDelay = Math.round(framesEncoded <= 0
```
makarandp0, line 747: @manjeshbhargav, I was thinking that these metrics would calculate the rates only for the duration since last metric was sent. so avgEncodeDelay would be calculated as:
```
deltaEncodeTime = totalEncodeTimeMS - lastTotalEncodeTimeMS;
lastTotalEncodeTimeMS = totalEncodeTimeMS;
deltaFrameEncoded = framesEncoded - lastFramesEncoded;
lastFramesEncoded = framesEncoded;
avgEncodeDelay = deltaFrameEncoded / deltaEncodeTime
``` 

That way those values would have meaningful spikes and delays can be detected by drawing a graph of those delays. If thy are average over the lifetime of peerConnection - I am afraid it wont be noticeable.

Maybe @ceaglest can tell us if these values would be detectable if they are averaged over longer times.

lib/signaling/v2/room.js
makarandp0, line 674: instead of adding the metrics to status report, why not generate them as part of `StatusReport` ?

makarandp0: I think we already generate different status report based on `prepareForInsights`

lib/util/eventobserver.js
```diff
@@ -76,13 +77,6 @@ class EventObserver extends EventEmitter {
       log[logLevel]('event', event);
 
       if (eventListener && group === 'signaling') {
-        const event = Object.assign(payload ? { payload } : {}, {
```
makarandp0, line 79: Thanks for removing this duplicate :)

makarandp0 (commented):
@manjeshbhargav I took a look and have couple of comments.  @ceaglest can you take a look as well.

lib/signaling/v2/room.js
manjeshbhargav: @makarandp0 I don't want these stats to be part of the public StatsReport class. I propose that we add them in a separate release.

lib/signaling/v2/room.js
manjeshbhargav: @makarandp0 If we want to do this, then it cannot be called as the average. We would need to remove the prefix `avg` from the fields.

lib/signaling/v2/room.js
manjeshbhargav: @makarandp0 No, we just set different values, but the fields are the same. I don't want to conditionally add properties to the class.

ceaglest: I wasn't expecting the public APIs to change as a part of this ticket so patching the events reported to insights is 👌. Maybe it deserves a comment explaining why the stats are added later.

lib/signaling/v2/room.js
ceaglest: I don't think an average over all time is going to be very helpful to debug media quality problems. The issue is that the longer the stats are collected the harder it becomes to see momentary spikes or troughs in these metrics.

For context I originally advocated for reporting totals and having the gateway compute the average (total, moving or otherwise).

> @makarandp0 If we want to do this, then it cannot be called as the average. We would need to remove the prefix avg from the fields.

What we intended was a [moving average](https://en.wikipedia.org/wiki/Moving_average) with a window size of two samples. It is an average but over a much smaller data set. If you are concerned that the field naming is not clear we could change the prefix to `movingAvg`, `rollingAvg` or `runningAvg` instead.

lib/util/movingaveragedelta.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+/**
+ * Calculates the moving average delta for the given samples. A sample (S) consists of
+ * a numerator (Sn) and a denominator (Sd). Given the sample size n, the moving average
+ * delta is calculated as follows:
+ *
+ * AvgDelta = Sum((Sn[i] - Sn[i-1]) / (Sd[i] - Sd[i-1])) for all i in [1, n - 1]
```
manjeshbhargav, line 8:
```suggestion
 * AvgDelta = Sum((Sn[i] - Sn[i-1]) / (Sd[i] - Sd[i-1])) / n for all i in [1, n - 1]
```

lib/util/movingaveragedelta.js
```diff
@@ -0,0 +1,67 @@
...
+
+    return Math.round(sampleSumDifferences / (samples.length - 1));
+  }
+
+  /**
+   * Put a sample and maybe get rid of the older samples to maintain sample size.
+   * @param numerator
+   * @param denominator
+   */
+  putSample(numerator, denominator) {
```
makarandp0, line 58: check for denominator being zero, to avoid `divide by zero` crashes

lib/signaling/v2/room.js
```diff
@@ -719,6 +749,76 @@ function handleSubscriptions(room) {
...
+function addAVSyncMetricsToRemoteTrackStats(trackStats, trackResponse, movingAverageDeltas) {
+  const {
+    estimatedPlayoutTimestamp,
+    framesDecoded,
+    jitterBufferDelay,
+    jitterBufferEmittedCount,
+    totalDecodeTime
+  } = trackResponse;
+  const augmentedTrackStats = Object.assign({}, trackStats);
+  const trackMovingAverageDeltas = movingAverageDeltas.get(trackStats.trackSid) || new Map();
```
makarandp0, line 799: with simulcast enabled would you  get multiple tracks with same trackSid and different `ssrc` - Shouldn't they be kept separately?

lib/util/movingaveragedelta.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+/**
+ * Calculates the moving average delta for the given samples. A sample (S) consists of
+ * a numerator (Sn) and a denominator (Sd). Given the sample size n, the moving average
+ * delta is calculated as follows:
+ *
+ * AvgDelta = Sum((Sn[i] - Sn[i-1]) / (Sd[i] - Sd[i-1])) for all i in [1, n - 1]
+ */
+class MovingAverageDelta {
```
makarandp0, line 10: I liked the way you refactored this!

lib/util/movingaveragedelta.js
```diff
@@ -0,0 +1,67 @@
...
+   */
+  get() {
+    const { _samples: samples } = this;
+    if (samples.length <= 0) {
+      return 0;
+    }
+    if (samples.length === 1) {
+      return Math.round(samples[0].numerator / samples[0].denominator);
+    }
+    const sampleDifferencesRight = samples.slice(0, samples.length - 1);
```
makarandp0, line 38: I did not understand this math -  some comments would be useful here.

I did not understand why do you need to keep 3 samples. Should we just keep last sample? and compute averages using
(current Numerator - last Numerator) / (current denominator - last denominator)

makarandp0 (changes requested):
@manjeshbhargav, nice use of the utility class for `MovingAverage'! But I did not understand the computations - I think some comments would help.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -192,14 +198,11 @@ class TwilioConnectionTransport extends StateMachine {
       }
     });
 
-    if (options.eventObserver) {
-      options.eventObserver.setPublisher(this._eventPublisher);
-    }
-
+    this._eventObserver.setPublisher(eventPublisher);
```
makarandp0, line 201: can we instead have caller ensure that publisher is already set before calling this class? That way both publisher and observer need not be sent down.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -308,7 +311,7 @@ class TwilioConnectionTransport extends StateMachine {
       this.preempt('disconnected', null, [error]);
       this._sendConnectOrSyncOrDisconnectMessage();
       this._twilioConnection.close();
-      this._eventPublisher.disconnect();
+      this._disconnectEventPublisher();
```
makarandp0, line 314: again, It does not feel right that _disconnectEventPublisher is declared as property. I am curious what makes it property v/s class method.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -341,13 +344,14 @@ class TwilioConnectionTransport extends StateMachine {
...
+   * @param {string} name - Event name
+   * @param {string} level - Event level
    * @param {object} payload - Event payload
-   * @returns {boolean} true if queued or published, false if disconnected from the Insights gateway
+   * @returns {void}
    */
-  publishEvent(groupName, eventName, payload) {
-    return this._eventPublisher.publish(groupName, eventName, payload);
+  publishEvent(group, name, level, payload) {
+    this._eventObserver.emit('event', { group, name, level, payload });
```
makarandp0, line 354: I do not like this `emit` pattern - where consumers of eventObserver are emitting on it. but I understand that its not exposed by this PR. We should clean this up when we get rid of `eventObserver` conenct option..

lib/util/movingaveragedelta.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+  }
+
+  /**
+   * Get the moving average delta.
+   * @returns {number}
+   */
+  get() {
+    const { _samples: samples } = this;
+    const denominatorDelta = (samples[1].denominator - samples[0].denominator) || 1;
```
makarandp0, line 31: I see that you are avoiding `divide by zero`  by replacing denominator delta with `1` . Would `Number.EPSILON` be more appropriate default?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -118,21 +126,19 @@ class TwilioConnectionTransport extends StateMachine {
       _bandwidthProfile: {
         value: options.bandwidthProfile
       },
+      _disconnectEventPublisher: {
+        value: () => eventPublisher.disconnect()
+      },
       _dominantSpeaker: {
         value: options.dominantSpeaker
       },
+      _eventObserver: {
```
makarandp0, line 135: +1 to replacing _eventObserver with _eventPublisher.

makarandp0 (changes requested):
its good overall, but have a few suggestion and change requests.

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -118,21 +126,19 @@ class TwilioConnectionTransport extends StateMachine {
       _bandwidthProfile: {
         value: options.bandwidthProfile
       },
+      _disconnectEventPublisher: {
+        value: () => eventPublisher.disconnect()
```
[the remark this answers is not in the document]
manjeshbhargav: @makarandp0 Making this a class method will not work since `eventPublisher` is accessible only from the constructor.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 I can add a TODO for this, since making this change now will lead to more changes in other classes and test files, which I want to avoid.

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 See my response [here](https://github.com/twilio/twilio-video.js/pull/1466#discussion_r632741895).

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @makarandp0 In this case, what we are doing is legitimate because the EventObserver is supposed to observe events emitted on it by the consumer. This is different from EventEmitter, where the consumer observes events emitted by it.

lib/util/movingaveragedelta.js
manjeshbhargav: @makarandp0 The number calculated using `EPSILON` wouldn't make sense. Dividing by 1 is closer to what we want.

makarandp0: you are right `EPSILON` is not good choice. But It does not look like 1 is right choice either, I think its better to return 0 from this function and  either a) not include 0 average values in the payload  (preferred) send to insights or b) report 0, but when querying in kibana we could exclude these 0 values.

lib/signaling/v2/twilioconnectiontransport.js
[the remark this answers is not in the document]
makarandp0: I see - I guess we can fix this when we have caller not sending  `eventPublisher` as your TODO comment below.

ceaglest (approved):
Looks good to me. I mainly focused on the moving average calculation and population of the insights reports.

makarandp0 (approved):

PR #1459, opened 2021-04-28
fix (#1458) LocalDataTrackOptions should extend LocalTrackOptions to enable providing a name

xdumaine:
Fixes the type to match documentation, and allowing providing a name in options of LocalDataTrack constructor. 

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

Issue #1458, opened 2021-04-28
LocalDataTrackOptions type should extend LocalTrackOptions


xdumaine:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

```ts
const dataTrack: Twilio.Video.LocalDataTrack = new Twilio.Video.LocalDataTrack({ name: 'whatever' }) // type errror
```

**Expected behavior:**

A name can be provided to the LocalDataTrack, according to [documentation](https://sdk.twilio.com/js/video/releases/2.8.0/docs/global.html#LocalDataTrackOptions)

**Actual behavior:**

Type error

PR #1468, opened 2021-05-06
TSDef Fix - LocalDataTrackOptions

PikaJoyce:
JIRA : [VIDEO-5116](https://issues.corp.twilio.com/browse/VIDEO-5116)

Reopening this branch to address @kseilezundane 's comments

**Contributing to Twilio**

### DIFF
--- a/tsdef/LocalDataTrack.d.ts
+++ b/tsdef/LocalDataTrack.d.ts
@@ -1,4 +1,4 @@
-import { LocalDataTrackOptions } from './types';
+import { LocalDataTrackOptions } from './LocalDataTrackOptions';
 import { Track } from './Track';
 
 export class LocalDataTrack extends Track {
--- a/tsdef/LocalDataTrack.d.ts
+++ b/tsdef/LocalDataTrack.d.ts
@@ -1,4 +1,4 @@
-import { LocalDataTrackOptions } from './LocalDataTrackOptions';
+import { LocalDataTrackOptions } from './types';
 import { Track } from './Track';
 
 export class LocalDataTrack extends Track {
--- a/tsdef/twilio-video-tests.ts
+++ b/tsdef/twilio-video-tests.ts
@@ -31,7 +31,8 @@ function getDataTrack(track: Video.LocalDataTrack) {
 }
 
 function makeDataTrack() {
-  const localDataTrack: Video.LocalDataTrack = new Video.LocalDataTrack({ name: 'coolroom', logLevel: 'off' });
+  const localTrackOptions: Video.LocalDataTrackOptions = { name: 'coolroom', logLevel: 'off' };
+  const localDataTrack: Video.LocalDataTrack = new Video.LocalDataTrack(localTrackOptions);
   localDataTrack.send('hello world');
 }
 
--- a/tsdef/twilio-video-tests.ts
+++ b/tsdef/twilio-video-tests.ts
@@ -236,7 +236,7 @@ async function initRoom() {
     maxVideoBitrate: 200,
     bandwidthProfile: {
       video: {
-        subscribedTrackSwitchOffMode: 'auto',
+        clientTrackSwitchOffControl: 'auto',
         contentPreferencesMode: 'auto',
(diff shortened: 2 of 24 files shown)


charliesantos (approved):

makarandp0:
@PikaJoyce,  I am wondering if removing the file would break any existing usage ? The [github issue](https://github.com/twilio/twilio-video.js/issues/1458) that prompted this change, [proposed the fix](https://github.com/twilio/twilio-video.js/pull/1459)  in file you are now deleting. Does that mean the user was getting the definition for the now deleted file? Would it break his usage  ? 
      
Also please add test case to cover the case pointed by the github issue.

makarandp0 (changes requested):
Please add test to verify that this addresses the issue.

tsdef/LocalDataTrack.d.ts
```diff
@@ -1,4 +1,4 @@
-import { LocalDataTrackOptions } from './LocalDataTrackOptions';
+import { LocalDataTrackOptions } from './types';
```
manjeshbhargav, line 1: This will form a circular import. LocalDataTrackOptions is imported from `types` to which `LocalDataTrack` is exported to, which imports LocalDataTrackOptions from `types`, and so on.

What we need to do is:

* Create `LocalTrackOptions.d.ts`
* Retain `LocalDataTrackOptions.d.ts`
* Derive LocalAudioTrackOptions, LocalVideoTrackOptions and LocalDataTrackOptions from `LocalTrackOptions.d.ts`
* In `types.d.ts`, export LocalTrackOptions as follows:
```ts
export { LocalTrackOptions } from './LocalTrackOptions';
```

manjeshbhargav (changes requested):
@PikaJoyce Please add tests to verify the fix.

tsdef/types.d.ts
```diff
@@ -13,6 +14,7 @@ import { RemoteVideoTrackPublication } from './RemoteVideoTrackPublication';
 import { Track } from './Track';
 import { VideoTrack } from './VideoTrack';
 
+export { LocalTrackOptions } from './LocalTrackOptions';
```
makarandp0, line 17: looks like this file does not use `LocalTrackOptions` anymore. Do we still need to import and export it from this file?

tsdef/twilio-video-tests.ts
```diff
@@ -30,6 +30,11 @@ function getDataTrack(track: Video.LocalDataTrack) {
   localDataTrack.send('hello world');
 }
 
+function makeDataTrack() {
```
makarandp0, line 33: since this change involves `LocalDataTrackOptions`, this test should create this type on separate line:
perhaps something like
```
const localTrackOptions: LocalDataTrackOptions = { name: 'coolroom', logLevel: 'off' };
const localDataTrack: Video.LocalDataTrack = new Video.LocalDataTrack(localTrackOptions);
```

makarandp0 (changes requested):

tsdef/types.d.ts
manjeshbhargav: @makarandp0 But `LocalTrackOptions` still need to be exported since it is a publicly available API.

makarandp0: yes, and I see that it is already exported from index.d.ts (just like other not part of types.d.ts for example `LocalDataTrack`).  Looks like we need to export types form types.d.ts only if they were defined in that file.

PikaJoyce: @makarandp0 we still have other types that use the `LocalTrackOptions` such as `MediaStreamTrackPublishOptions`.

makarandp0: @PikaJoyce, you are right. I missed that reference. I guess importing `LocalTrackOptions` is needed. But does this file need to re-export it?

PikaJoyce: @makarandp0 Ah, sorry! I see what you mean! I need to import into types but will move the export to `index.d.ts`.

makarandp0 (approved):

PR #1474, opened 2021-05-14
VIDEO-4071 | TS Tooling

charliesantos:
TS Support for:
- New code - Only applies to new files.
- Type definitions - Definitions file is automatically generated for the ts file after transpiling. No need to manually create one.
- Lint - Uses existing eslint rules, but turned off unwanted typescript lint rules.
- JSDoc - API documentation is autogenerated for both TS and JS files. Just continue writing the docs in JSDoc format.
- Unit testing - Tests are run separately for TS and JS files. Ideally, we would use ts-node to run both, however, ts-node has limitations to our existing code base even with latest ts-node version. Supporting ts-node requires updating majority of our tests which is out of scope. Personally, it has little to no benefit IMO.
- Integration testing - Tests are all written in JS. Supporting both TS and JS to run in the same test runner is causing an issue between karma-typescript and browserify preprocessors. And using karma-typescript as the sole preprocessor for both TS and JS also causes the same issue as our unit tests because ts-node is used under the hood. Once we convert all our public APIs to TS, we should be able to completely switch to writing integration tests in TS.


makarandp0 (commented):
Its looking really good @charliesantos. And thank you for adding `preflight.ts` as a demo!
I have just one question - I see that you have both build.js and buld.ts I am wondering if we can use `tsc` as a compiler for both ts and js files.

charliesantos:
>I see that you have both build.js and buld.ts I am wondering if we can use tsc as a compiler for both ts and js files.

Unfortunately, I was having issues using tsc for our js files. Same issues I'm having right now for running unit tests. Making tsc work for our existing js files is taking so much time (already spent half day 😢 ) So to prevent unnecessary risk and time spent, I'm keeping the existing js build and just adding ts build on top for simplicity/less risk. This way, it's also easy to migrate to ts each file one by one.

lib/index.js
```diff
@@ -61,6 +61,10 @@ Object.defineProperties(Video, {
     enumerable: true,
     value: require('./vendor/loglevel')
   },
+  runPreflight: {
+    enumerable: true,
+    value: require('../es5/preflight').runPreflight
```
makarandp0, line 66: @charliesantos, I see that here we are 1st compiling `.ts` files into es5 and then using it from here. I do not like this dependency on generated code from the non-generated code. I am afraid that as we write more and more typescript code, our local workflow would become cumbersome as it would require `build` process. 
I see that this is needed because this file is `.js` and thus cannot include `.ts`. Would converting this file into `.ts` solve this problem or is there more to it?

package.json
```diff
@@ -87,11 +95,13 @@
...
-    "lint:ts": "eslint ./tsdef/*.ts",
+    "lint:ts": "eslint ./tsdef/*.ts ./lib/**/*.ts",
     "lint": "npm-run-all lint:js lint:ts",
-    "test:unit": "mocha ./test/unit/index.js",
+    "test:unit": "npm-run-all test:unit:js test:unit:ts",
+    "test:unit:js": "nyc --report-dir=./coverage/js --extension=.js --include=lib/**/* --reporter=html --reporter=lcov --reporter=text mocha ./test/unit/index.js",
+    "test:unit:ts": "nyc --report-dir=./coverage/ts --extension=.ts --include=lib/**/* --reporter=html --reporter=lcov --reporter=text mocha -r ts-node/register ./test/unit/index.ts",
     "test:integration:adapter": "node ./scripts/karma.js karma/integration.adapter.conf.js",
-    "test:integration": "node ./scripts/karma.js karma/integration.conf.js",
+    "test:integration": "npm run build:es5 && node ./scripts/karma.js karma/integration.conf.js",
```
makarandp0, line 104: did you look into [karma-typescript](https://www.npmjs.com/package/karma-typescript) to run our integration tests?

makarandp0 (commented):
Great work @charliesantos,  I had just couple of  questions. I will try this out locally soon, and may have few more questions.

lib/index.js
charliesantos: Hey @makarandp0 , good question. Actually, this was originally using the ts file. The problem is that this causes an issue with our integration tests. karma-typescript or browserify is having an issue requiring both ts and js. See PR description regarding the problem on integration tests.

package.json
charliesantos: Yes. See PR description that describes the issue of using karma-typescript in our codebase.

charliesantos:
> Its looking really good @charliesantos. And thank you for adding `preflight.ts` as a demo!
> I have just one question - I see that you have both build.js and buld.ts I am wondering if we can use `tsc` as a compiler for both ts and js files.

Great questions @makarandp0 . I actually went to this route in my initial implementation - all using ts compiler and using karma-typescript. However, to summarize the issues I have in the PR description, using the ts compiler and karma-typescript causes issues in our current code base related to mixing and matching ts/js files.

This is the same old issue we had on voice which we never found the fix for. How we fixed it is, we converted all public APIs to TS which helped fix the karma-typescript issue. Once karma-typescript issue was fixed, we were able to do this `require('./preflight').runPreflight` (without the es5), instead of `require('../es5/preflight').runPreflight;`. We can talk more live if needed.

PikaJoyce (approved):
This is awesome Charlie!! 🚢

lib/index.js
manjeshbhargav: @charliesantos What if we convert the whole `index.js` to `index.ts`. The JS modules can be imported as:
```ts
import connect = require('./connect');
```
and the TS modules (only preflight for now) can be imported as:
```ts
import { runPreflight } from './preflight';
```

charliesantos:
Per our conversation, we'll merge this to the feature branch and @makarandp0 will create a separate PR for the next action items.

PR #1482, opened 2021-05-26
VIDEO-4419 | Update isSupported to match browser matrix support

charliesantos:
Previously, we don't check for mobile browser specifically, we only test for WebRTC support. Now that mobile browsers are starting to support WebRTC, isSupported is returning true for these browsers. We want to still return false for these until QE officially support them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Version 1.x will End of Life on September 8th, 2021.** Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) to plan your migration to the latest 2.x version. Support for the 1.x version ended on December 4th, 2020.
 
+2.14.1 (In Progress)
+=====================
+
+Bug Fixes
+---------
+
+- Fixed a bug where `isSupported` is returning `true` on certain unsupported mobile browsers. With this release, `isSupported` should now match our list of [supported browsers](https://www.twilio.com/docs/video/javascript#supported-browsers).
+
 2.14.0 (May 11, 2021)
 =====================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 Bug Fixes
 ---------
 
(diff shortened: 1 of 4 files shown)


lib/util/support.js
```diff
@@ -33,6 +42,16 @@ function getParenthesizedSubstrings(string) {
...
 }
 
+/**
+ * Check whether the current browser is a mobile browser
+ * @param {string} userAgent
+ * @returns {boolean}
+ */
+function isMobile(userAgent) {
+  userAgent = userAgent || navigator.userAgent;
+  return /Mobi/.test(userAgent);
```
manjeshbhargav, line 52: Any reason why we are using `Mobi` instead of `Mobile`?

manjeshbhargav (approved):
@charliesantos LGTM. One nitpick: can you remove the trailing commas? We don't use them in this project, so I want to keep it consistent.

lib/util/support.js
charliesantos: Best practice, as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent). Scroll all the way to the bottom. But to summarize, all browsers uses `Mobile`. But opera/presto uses `Mobi`

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Version 1.x will End of Life on September 8th, 2021.** Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) to plan your migration to the latest 2.x version. Support for the 1.x version ended on December 4th, 2020.
 
+2.14.1 (In Progress)
+=====================
+
+Bug Fixes
+---------
+
+- Fixed a bug where `isSupported` is returning `true` on certain unsupported mobile browsers. With this release, `isSupported` should now match our list of [supported browsers](https://www.twilio.com/docs/video/javascript#supported-browsers).
```
manjeshbhargav, line 11:
```suggestion
- Fixed a bug where `isSupported` was returning `true` on certain unsupported mobile browsers. With this release, `isSupported` should now return true only for the [browsers supported by twilio-video.js](https://www.twilio.com/docs/video/javascript#supported-browsers).
```

manjeshbhargav (approved):

PR #1497, opened 2021-06-15
WIP: preflight + ts tooling update

makarandp0:
This change introduces preflight lite api + update to typescript tooling to simplify the workflow

TS Tooling Updates
===============
- use typescript compiler to compile both ts and js files. Removes intermittent `build:ts` step.
- type definitions will still be authored manually (this will change when we migrate index.js to typescript)

Preflight Update:
===============
Initial implementation of preflight-lite api is ready
Still todo:
- [x] calculate quality score
- [x] add more integration tests
- [x] Fix incoming/outgoing bitrate computations
  

**Contributing to Twilio**


lib/index.js
```diff
@@ -63,7 +63,7 @@ Object.defineProperties(Video, {
   },
   runPreflight: {
     enumerable: true,
-    value: require('../es5/preflight').runPreflight
+    value: require('../es5/preflight/preflighttest').runPreflight
```
charliesantos, line 66: With the new TS Tooling changes, we don't need to reference es5 folder here.

test/integration/spec/preflight.js
```diff
@@ -1,15 +1,128 @@
-/* eslint-disable no-console, no-undefined */
+/* eslint-disable no-undefined */
 'use strict';
 
 const assert = require('assert');
-const { runPreflight } = require('../../../lib');
+// const testPreflight = require('../../../lib/preflight/preflight');
+const runPreflight = require('../../../es5/index').runPreflight;
```
charliesantos, line 6: We also probably don't need to reference es5 here?

charliesantos (commented):
Good stuff @makarandp0 ! I left a few comments on the TS related things as a first pass. I think we need to do the following as well:
- delete .eslintignore
- remove tsbuild from .gitignore and .npmignore
- delete scripts/build-ts.js

lib/index.js
makarandp0: yep, Thanks, I will remove `es5`

test/integration/spec/preflight.js
makarandp0: I think we still need this, as we are not compiling test `js` code. it takes dependency on generated (es5) lib code.

charliesantos: Ok got it.

lib/index.js
makarandp0: looks like we have to keep this `es5` for now - integration tests load this version (and not es5) version - so it needs to go into es5 for the generated file. We could modify all integration tests to load the es5 version directly, but we can keep it for the next iteration when we update this file to be `.ts`

test/unit/spec/preflight.ts
```diff
@@ -1,9 +1,9 @@
...
-import { Preflight } from '../../../lib/preflight';
+import { runPreflight } from '../../../lib/preflight/preflighttest';
 
 describe('Preflight', () => {
   it('should return a Preflight object', () => {
-    const preflight = new Preflight();
+    const preflight = runPreflight('foo', {});
     assert(!!preflight);
   });
 });
```
charliesantos, line 9: Do we have plans to add more unit tests?

charliesantos (changes requested):
This looks great @makarandp0 ! Docs looks good too. We might need to ask QE/Sean to review the docs though. To make sure there are no gaps.

lib/preflight/TimeMeasurementImpl.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-undefined */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { TimeMeasurement } from '../../tsdef/PreflightTypes';
```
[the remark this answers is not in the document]
makarandp0: removing type also works, but I think `type` makes it more obvious.

lib/preflight/TimeMeasurementImpl.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class TimeMeasurementImpl {
```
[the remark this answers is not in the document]
makarandp0: sounds good. I will rename it as 'timer`.

lib/preflight/makeStat.ts
```diff
@@ -0,0 +1,18 @@
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { Stats } from '../../tsdef/PreflightTypes';
```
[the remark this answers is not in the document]
makarandp0: I removed the eslint-disable, but keeping `type`

test/unit/spec/preflight.ts
makarandp0: no because we have very good coverage with integration tests.

makarandp0: Now I have added more tests for new ts files.

lib/preflight/preflighttest.ts
```diff
@@ -0,0 +1,640 @@
...
+        if (remoteVideoTrack.packetsReceived !== null) {
+          packetsReceived += remoteVideoTrack.packetsReceived;
+        }
+      }
+      collectedStats.packetLoss.push(packetsReceived ? packetsLost * 100 / packetsReceived : 0);
+    }
+  }
+
+  private async collectRTCStatsForDuration({ duration, collectedStats, senderPC, receiverPC }:
+    { duration: number; collectedStats: PreflightStats; senderPC: RTCPeerConnection; receiverPC: RTCPeerConnection; }): Promise<PreflightStats> {
```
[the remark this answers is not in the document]
makarandp0: I will update this to not take a type, but a simple parameter list.

lib/preflight/preflighttest.ts
```diff
@@ -0,0 +1,640 @@
...
+const SECOND = 1000;
+const DEFAULT_TEST_DURATION = 10 * SECOND;
+
+
+/**
+ * progress values that are sent by {@link PreflightTest#event:progress}
+ * @enum {string}
+ */
+// eslint-disable-next-line
+const PreflightProgress = {
```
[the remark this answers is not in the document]
makarandp0: I removed the eslint-disable now. About enum v/s object - Rest of such types in our codebase are const objects 's and I wanted to keep it consistant.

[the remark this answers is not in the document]
charliesantos: But if we're using TS, should we use TS way? This doesn't break anything by the way.

lib/preflight/preflighttest.ts
```diff
@@ -0,0 +1,640 @@
...
+  /**
+   * established ICE connection. This is measured from ICE connection `checking` to `connected` state.
+   */
+  iceConnected: 'iceConnected'
+
+};
+
+declare interface InternalPreflightTestReport extends PreflightTestReport {
+  localAudio?: PreflightReportTrackStats,
+  localVideo?: PreflightReportTrackStats
```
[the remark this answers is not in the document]
makarandp0: No, I don't thing avoiding an internal type is worth adding new fields to the public type.

lib/preflight/preflighttest.ts
```diff
@@ -0,0 +1,640 @@
...
+    options = Object.assign(options, {
+      preferredVideoCodecs: [{ codec: 'VP8', simulcast: true }]
+    });
+
+    this._testTiming.start();
+    this.runPreflightTest(token, options);
+  }
+
+  toString(): string {
+    return '[Preflight]';
```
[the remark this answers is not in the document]
makarandp0: yes, I think it would be useful to follow same pattern. I will update.

lib/preflight/preflighttest.ts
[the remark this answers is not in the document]
makarandp0: okay, in this case I think its okay to use enum, sine we are not publicly exposing this type. We are just exposing the parameter as string.

CHANGELOG.md
```diff
@@ -2,6 +2,69 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Version 1.x will End of Life on September 8th, 2021.** Check [this guide](https://www.twilio.com/docs/video/migrating-1x-2x) to plan your migration to the latest 2.x version. Support for the 1.x version ended on December 4th, 2020.
 
+2.17.0 (In Progress)
+====================
+**Preflight API Pilot**
```
charliesantos, line 7: We already have this in beta in an older version right? Should this be beta instead? cc @seancoleman2

CHANGELOG.md
```diff
@@ -2,6 +2,69 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
+-----------------------
+
+- This release adds a new experimental API `runPreflight` to help test the connectivity with twilio servers.
```
charliesantos, line 10: We should mention that this is an update of the old preflight API and it's not a new concept.

CHANGELOG.md
```diff
@@ -2,6 +2,69 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
+The quality score can be one of:
+
+| Value                    | Meaning
+| -------------------------|--------------------------------|
+| 0                        | Failed
+| 1                        | Degraded
+| 2                        | Fair
+| 3                        | Good
+| 4                        | Great
+| 5                        | Excellent
```
charliesantos, line 42: Not sure if I missed it, but is the report documented in our API docs?

charliesantos (approved):
Great work @makarandp0 ! Just left a few comments which I don't think should block us from cutting an RC.

CHANGELOG.md
makarandp0: yes. here is a link to the doc: https://46634-46595751-gh.circle-artifacts.com/0/dist/docs/global.html#PreflightTestReport

CHANGELOG.md
makarandp0: yeah, I think this need some rewording and update, but that we can do it before going GA

CHANGELOG.md
makarandp0: I do not think we want to release this as "beta" release - since there is additional cost of keeping it from the mainline, and then updating it everytime mainline progresses. But we can decide on those details closer to releasing to public.

PR #1513, opened 2021-07-01
video 5988 - Use pc stats directly 

makarandp0:
Preflight test was using webrtc module's getStat api, which returns stats [only for 1st nominated](https://github.com/twilio/twilio-webrtc.js/blob/master/lib/getstats.js#L108) active-candidate-pair, which can result in bytesSent/Received to be reported as 0.

Also the api was found to be not returning round trip times for safari.
Fixed this by refactoring the code to use PeerConnection.getStats directly, and just retrieving the stats needed for preflight.
 

**Contributing to Twilio**


lib/preflight/getCombinedConnectionStats.ts
```diff
@@ -0,0 +1,141 @@
...
+//   const stats = await pc.getStats();
+//   stats.forEach(stat => {
+//     if (['outbound-rtp', 'remote-inbound-rtp', 'inbound-rtp', 'remote-inbound-rtp', 'candidate-pair'].includes(stat.type)) {
+//       if (typeof stats[statName] === 'undefined') {
+//         foundInReport.add(stat.type + '-' + stat.kind);
+//       }
+//     }
+//   });
+//   return foundInReport;
+// }
```
charliesantos, line 46: Should we remove this?

lib/preflight/getCombinedConnectionStats.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    return report;
+  }, {});
+}
+
+function extractSelectedActiveCandidatePair(stats: RTCStatsReport) : SelectedIceCandidatePairStats | null {
+  let activeCandidatePairStatsFound: RTCIceCandidatePairStats|null = null;
+  stats.forEach(statPair => {
+    if (statPair.type === 'candidate-pair' && statPair.nominated) {
```
charliesantos, line 118: I think `nominated` is not supported on some browsers. Here's a cross browser solution I did for voice. Let me know if this works or if this is applicable to video. See whole method [here](https://github.com/twilio/twilio-client.js/blob/master/lib/twilio/rtc/stats.js#L56).

```js
const selectedCandidatePairReport = candidatePairs.find(pair =>
      // Firefox
      pair.selected ||
      // Spec-compliant way
      (transport && pair.id === transport.selectedCandidatePairId)
    );
```

lib/preflight/getCombinedConnectionStats.ts
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { RTCIceCandidateStats, SelectedIceCandidatePairStats,  } from '../../tsdef/PreflightTypes';
+import type { RTCIceCandidatePairStats } from './rtctypes';
+
+interface RTCStatsReport {
+  [x: string]: any;
+  forEach(callbackfn: (value: any, key: string, parent: RTCStatsReport) => void, thisArg?: any): void;
```
charliesantos, line 7: This is basically `any` type. If we're not declaring all properties, how about doing `type RTCStatsReport = any` instead?

charliesantos (changes requested):

lib/preflight/getCombinedConnectionStats.ts
makarandp0: Thanks @charliesantos - I like your solution better, I will update.

lib/preflight/getCombinedConnectionStats.ts
makarandp0: somehow with with the declaration as you suggested, I see an error:
Parameter 'stat' implicitly has an 'any' type. when I use forEach later...with the forEach explicitely declared I do not see the same error

charliesantos (approved):
Nice work @makarandp0 !
I noticed there's a stable test failing. Can't tell if it's related to this PR https://app.circleci.com/pipelines/github/twilio/twilio-video.js/4357/workflows/e1f2db5e-f980-4c85-9bf6-85b25fb596d7/jobs/47325

lib/preflight/getCombinedConnectionStats.ts
charliesantos: Ok

makarandp0:
> Nice work @makarandp0 !
> I noticed there's a stable test failing. Can't tell if it's related to this PR https://app.circleci.com/pipelines/github/twilio/twilio-video.js/4357/workflows/e1f2db5e-f980-4c85-9bf6-85b25fb596d7/jobs/47325

yeah, I have seen those tests fail before and those failures are unrelated.

PR #1515, opened 2021-07-03
VIDEO-5921- Adding workarounds for chrome iOS support

PikaJoyce:
JIRA: [VIDEO-5021](https://issues.corp.twilio.com/browse/VIDEO-5921)

Applying the workaround for iOS Chrome as we do for Safari. @charliesantos recommended doing it this way which I think is a much better idea than my previous one! 

**Contributing to Twilio**


lib/util/support.js
```diff
@@ -152,4 +158,9 @@ function isSupported() {
     && (!isMobile() || supportedMobileBrowsers.includes(browser));
 }
 
-module.exports = isSupported;
+module.exports = {
+  isSupported,
+  // eslint-disable-next-line no-warning-comments
+  // TODO(joyce): Move browser detection to a different util
+  isChromeIOS
```
makarandp0, line 165: we should expose this as a function instead of variable - that way `guessBrowser` wont get called just be loading twilio-video.

lib/media/track/localmediatrack.js
```diff
@@ -32,7 +32,7 @@ function mixinLocalMediaTrack(AudioOrVideoTrack) {
       // NOTE(mpatwardhan): by default workaround for WebKitBug1208516 will be enabled on Safari browsers
       // although the bug is seen  mainly on iOS devices, we do not have a reliable way to tell iOS from MacOs
       // userAgent on iOS pretends its macOs if Safari is set to request desktop pages.
-      const workaroundWebKitBug1208516 = guessBrowser() === 'safari'
+      const workaroundWebKitBug1208516 = guessBrowser() === 'safari' || isChromeIOS
```
makarandp0, line 35: please make isChromeIOS a function not a variable.

lib/util/support.js
```diff
@@ -126,6 +126,12 @@ function mobileWebKitBrowser(browser) {
   }) || null;
 }
 
+// Browser detection to check for whether or not the browser is Chrome iOS.
+// eslint-disable-next-line no-warning-comments
+// TODO(joyce): Move browser detection to a different util
+const isAndroid = /android/.test(navigator.userAgent);
+const isChromeIOS = guessBrowser() === 'chrome' && isMobile() && !isAndroid;
```
makarandp0, line 133: is there anything in chrome ios useragent that we can use to see if its on iOS...that is instead of assuming its 
`iOS = !isAndroid`

charliesantos: I think we can do `/iPad|iPhone|iPod/`

lib/util/support.js
```diff
@@ -126,6 +126,12 @@ function mobileWebKitBrowser(browser) {
   }) || null;
 }
 
+// Browser detection to check for whether or not the browser is Chrome iOS.
+// eslint-disable-next-line no-warning-comments
+// TODO(joyce): Move browser detection to a different util
+const isAndroid = /android/.test(navigator.userAgent);
```
charliesantos, line 132: This is duplicated in the isSupported function. We should create `isAndroid` function and reuse that.

charliesantos (changes requested):
Looks great @PikaJoyce . I agree with @makarandp0 's comments. To summarize, I think we should:

1) create isIOS function
```js
function isIOS() {
  return /iPad|iPhone|iPod/.test(navigator.userAgent);
}
```

2) create isChromeIOS function and export it
```js
function isChromeIOS() {
  return guessBrowser() === 'chrome' && isMobile() && isIOS();
}
```

3) add serverrender test https://github.com/twilio/twilio-video.js/tree/master/test/serversiderender

lib/util/support.js
charliesantos: Please add unit tests for isChromeIOS once we make it a function.

PikaJoyce:
Thanks @charliesantos and @makarandp0 for the suggestions. I'll get right to working on all of that :)

lib/util/support.js
PikaJoyce: Sounds good, I'll go ahead and make that change. 

@charliesantos @makarandp0 I'll make another unit test file just for these browser device detection utils since we're going to be moving them out of the support.js file. How does that sound?

charliesantos: Sounds good to me.

test/unit/spec/util/browserdetection.js
```diff
@@ -0,0 +1,65 @@
...
+    [
+      'Firefox on iPhone',
+      'Mozilla/5.0 (iPhone; CPU OS 14_4_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/33.1 Mobile/15E148 Safari/605.1.15'
+    ],
+    [
+      'Brave on iPhone',
+      'Mozilla/5.0 (iPhone; CPU iPhone OS 14_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.2 Mobile/15E148 Safari/604.1',
+      null,
+      {}
+    ]
```
charliesantos, line 52: Can we also add negative tests for:
Desktop: chrome, safari, firefox, edge
iOS: edge
Android: chrome, firefox, edge

lib/util/support.js
```diff
@@ -152,4 +167,9 @@ function isSupported() {
     && (!isMobile() || supportedMobileBrowsers.includes(browser));
 }
 
-module.exports = isSupported;
+module.exports = {
+  isSupported,
```
charliesantos, line 171: Since we're updating how we're exporting this, we should update other places. Can you grep the codebase and update where we are using this? I don't memorize them all but I think it includes the following:
- lib/index.js
- some integration tests
- umd tests
- serversiderender tests

PikaJoyce: @charliesantos Do you think it's worthwhile to do this since I'd have to change it back again in the next PR? Or should I just make a new `browserdetection.js` file in this PR and we wouldn't have to adjust all the places that are importing `isSupported`?

charliesantos: @PikaJoyce I'll leave it up to you. As long as we don't have a broken build when you cut an RC 😄

lib/util/browserdetection.js
```diff
@@ -0,0 +1,45 @@
...
+  return /Android/.test(navigator.userAgent);
+}
+
+function isIOS() {
+  return /iPad|iPhone|iPod/.test(navigator.userAgent);
+}
+
+function isChromeIOS() {
+  return guessBrowser() === 'chrome' && isMobile() && isIOS();
+}
```
charliesantos, line 16: Can we add docs to these new functions?
isAndroid
isIOS
isChromeIOS

lib/util/browserdetection.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Check whether the current browser is a mobile browser
+ * @param {string} userAgent
+ * @returns {boolean}
+ */
+function isMobile(userAgent) {
+  userAgent = userAgent || navigator.userAgent;
```
charliesantos, line 24: I think we can remove passing userAgent and this line, since we are not following this pattern anyway on the other similar methods.

lib/util/browserdetection.js
charliesantos: Hmmm, I just remembered, do we know if iPad is considered mobile? If not, isChromeIOS might return false. Also, is the workaround we're trying to do applicable for both iPad and iPhone? @makarandp0 @PikaJoyce

charliesantos (changes requested):
Looks great overall @PikaJoyce ! Just left a few minor comments.

lib/util/browserdetection.js
PikaJoyce: We can additionally check for `CriOS` in the regex as well. I believe that iPad falls under that category. We currently have a test that runs this which passes. The second question is definitely one that @makarandp0 knows more about!

charliesantos: It seems iPad falls under mobile since the user agent contains the `mobile` string. So we're good there. Base on the existing workarounds, it seems they are getting applied for both ipad and iphone. So I think your current code works for chrome. No need to change anything here.

charliesantos (approved):
Great work 🎉

PR #1520, opened 2021-07-12
video 6079: Packet loss had wrong values. 

makarandp0:
This fixes an issue were packet loss fractions were getting rounded off - as `MovingAverageDelta` get was rounding the values to nearest integer before returning.
Also updated to report packet loss %age instead of packet  loss fraction.  

**Contributing to Twilio**


lib/preflight/preflighttest.ts
```diff
@@ -315,12 +314,13 @@ export class PreflightTest extends EventEmitter {
     this._sentBytesMovingAverage.putSample(bytesSent, timestamp);
     this._receivedBytesMovingAverage.putSample(bytesReceived, timestamp);
     this._packetLossMovingAverage.putSample(packetsLost, packets);
-
     if (hasLastData) {
       collectedStats.outgoingBitrate.push(this._sentBytesMovingAverage.get());
       collectedStats.incomingBitrate.push(this._receivedBytesMovingAverage.get());
       const fractionPacketLost = this._packetLossMovingAverage.get();
-      collectedStats.packetLoss.push(fractionPacketLost);
+      const percentPacketsLost = Math.min(100, fractionPacketLost * 100);
```
charliesantos, line 321: When will fractionPacketLost ever be over 1? I'm assuming that is the reason for the Math.min?

makarandp0: it shouldn't be over 1, but since we report this value to customers, I added Math.min just in case

charliesantos (approved):

PR #1521, opened 2021-07-13
convert our index.js into index.ts

makarandp0:
This change converts top level - index.js into index.ts - so that we do not have to take dependency on generated files.
We are still not generating the types from the index.ts, and using types from tsdef/index.d.ts - and that because I couldn't figure out how to export LocalAudioTrack/LocalVideoTrack as a type as well as top level function export.

This necessitated few more changes in test files.
- Our js tests now must use generated files. This is good because we really want to test generated change.
- Fixed an issue with typescript generated code  where classes derived from Error/Array objects did not return correct instance of. More about the issue @ https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
 
**Contributing to Twilio**


charliesantos (changes requested):
Nice one @makarandp0 ! Can we add tests for this?

makarandp0:
Hey @charliesantos - I have now fixed the tests, it needed some changes more about them in PR description. Please take another look!

package.json
```diff
@@ -97,7 +97,7 @@
     "lint:js": "eslint ./lib ./test/*.js ./docker/**/*.js ./test/framework/*.js ./test/lib/*.js ./test/integration/** ./test/unit/** ",
     "lint:ts": "eslint ./tsdef/*.ts ./lib/**/*.ts",
     "lint": "npm-run-all lint:js lint:ts",
-    "test:unit": "npm-run-all test:unit:js test:unit:ts",
+    "test:unit": "npm-run-all build:es5 test:unit:js test:unit:ts",
```
charliesantos, line 100: Why do we need to build es5 for unit tests?

charliesantos (approved):
Looks good! Just had one question.

package.json
makarandp0: That is because our integration and unit tests share some shared libraries. Any library used by integration test cannot use "source" files - It must `require es5` version. And same libraries are used by unit tests, we must build es5 before unit tests.

Issue #1528, opened 2021-07-23
Chrome 92 limits the creation of MediaPlayers


arunpbk-stayqrious:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

Connect to a twilio room, and ask the participants to turn their video on and off for ~ 75 times (in total).

**Expected behavior:**

Should work as expected

**Actual behavior:**

Video stops playing with the error
`[Intervention] Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27`

**Software versions:**

 - [x] Browser(s): Chrome 92
 - [x] Operating System: Mac / Windows
 - [x] twilio-video.js: Probably all versions
 - [x] Third-party libraries (e.g., Angular, React, etc.): React

The culprit seems to be the allocations for dummy media elements created for tracking media track dimensions in mediatrack.js 
https://github.com/twilio/twilio-video.js/blob/master/lib/media/track/mediatrack.js#L233

charliesantos:
Thanks @arunpbk-stayqrious for submitting. We are currently working on this issue.

fippo:
@anna-vasilko girts got a note about this

anna-vasilko:
Despite the upcoming improvement in this SDK (PR #1530), the new limit of MediaPlayers in Chrome 92 may still impact group room applications where participants subscribe to many media tracks. Find more in this chrome ticket 
https://bugs.chromium.org/p/chromium/issues/detail?id=1232649

matt-hensley:
Seeing this issue with automatic track switching enabled with a handful of participants in a room.

m-andrew-albright:
[...]

Issue #1367, opened 2021-02-01
Question about detached MediaStreamTracks, HTMLAudioElements & HTMLVideoElements pilling up


bwillner:
Hi,

when memory profiling my react implementation of twilio-video app I noticed that there are detached MediaStreamTracks, HTMLAudioElements and HTMLVideoElements pilling up after a call has ended and the room was disconnected. Around 4 per call. 
Is this expected behavior?

Please let me know if you need any additional logs etc.

Thanks in advance

charliesantos:
Hey @bwillner , thanks for reporting. Can you please export a heaptimeline using chrome dev tools? Also, how many participants are in the room? Can you please provide some steps that I can try on my end?

Thank you,
Charlie

bwillner:
Hi @charliesantos ,
There are two participants in the room. Room type is group room. I can send you a heaptimeline of the call. The corresponding Room ID is: RM2a800e21d54bd4f12d7d0745b7bd06e3.

How should I send you the Heaptimeline? I'd rather not post it here.

If it is any help for you I probably could give you access to our testing server.

Thanks,

charliesantos:
Hi @bwillner , 
Thank you for this additional information. You can send the heaptimeline to csantos@twilio.com if you prefer. No need to give me access to your testing server for now. I'll let you know if I need more information after inspecting the heaptimeline.

Thank you,
Charlie

bwillner:
Have just sent you the Heaptimeline via Email.

charliesantos:
Hi @bwillner , thank you for sending the heaptimeline. I'm also able to reproduce it. We already have an internal tracker for this and will keep you posted once we have a fix.
https://issues.corp.twilio.com/browse/JSDK-3120

alucio-echen:
Has there been any updates to this issue? 

I believe we are seeing something similar in chrome 92 (Chrome Beta) where we get the following error in the console:  
```[Intervention] Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27```

When this happens new video / audio tracks don't play. 

Our react app is roughly based on the sample react video app and switches video publications on and off frequently during meetings (20 participants /w video & audio). It seems like each time a publication is stopped and re-started a new media player is being allocated and they are never freed. 

We are calling detach whenever the track is switched off (like in the sample app) and I've tried different combinations of stopping the playback of the element, setting the src on the video element to null, etc. but the issue remains.

[...]

PR #1530, opened 2021-07-24
video-6336: media element leak causes error on chrome 92 

makarandp0:
chrome 92 started [limiting](https://chromium-review.googlesource.com/c/chromium/src/+/2816118) the WebMediaPlayers. This [breaks](https://bugs.chromium.org/p/chromium/issues/detail?id=1232649) lot of webrtc applications. This fix ensures that 1) we do not leak media elements created internally and 2) we clear the `srcObject` before removing media elements.

The application that use `RemoteAudioTrack.detach`() method need to ensure that they clear srcObject on the returned elements as well.

related github issues: https://github.com/twilio/twilio-video.js/issues/1367 & https://github.com/twilio/twilio-video.js/issues/1528
  
**Contributing to Twilio**

### DIFF
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 'use strict';
 
 const { guessBrowser } = require('@twilio/webrtc/lib/util');
@@ -145,8 +146,11 @@ class MediaTrack extends Track {
   _end() {
     this._log.debug('Ended');
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
+      this._detachElementInternal(this._dummyEl);
+      this._dummyEl.remove();
+      this._dummyEl.srcObject = null;
       this._dummyEl.oncanplay = null;
+      this._dummyEl = null;
     }
   }
 
@@ -259,6 +263,13 @@ class MediaTrack extends Track {
     return elements.map(this._detachElement.bind(this));
   }
 
+  _detachElementInternal(el) {
+    const mediaStream = el.srcObject;
+    if (mediaStream instanceof this._MediaStream) {
+      mediaStream.removeTrack(this.processedTrack || this.mediaStreamTrack);
+    }
+  }
+
   /**
    * @private
    */
@@ -266,12 +277,7 @@ class MediaTrack extends Track {
     if (!this._attachments.has(el)) {
       return el;
     }
-
-    const mediaStream = el.srcObject;
-    if (mediaStream instanceof this._MediaStream) {
-      mediaStream.removeTrack(this.processedTrack || this.mediaStreamTrack);
-    }
-
+    this._detachElementInternal(el);
     this._attachments.delete(el);
 
     if (this._shouldShimAttachedElements && this._elShims.has(el)) {
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -119,24 +119,29 @@ class MediaTrack extends Track {
    * @private
    */
   _initialize() {
-    const self = this;
+    try {
+      const self = this;
 
-    this._log.debug('Initializing');
-    this._dummyEl = this._createElement();
+      this._log.debug('Initializing');
+      this._dummyEl = this._createElement();
 
-    this.mediaStreamTrack.addEventListener('ended', function onended() {
-      self._end();
-      self.mediaStreamTrack.removeEventListener('ended', onended);
-    });
+      this.mediaStreamTrack.addEventListener('ended', function onended() {
+        self._end();
+        self.mediaStreamTrack.removeEventListener('ended', onended);
+      });
 
-    if (this._dummyEl) {
-      this._dummyEl.muted = true;
-      this._dummyEl.oncanplay = this._start.bind(this, this._dummyEl);
+      if (this._dummyEl) {
+        this._dummyEl.muted = true;
+        this._dummyEl.oncanplay = this._start.bind(this, this._dummyEl);
 
-      // NOTE(csantos): We always want to attach the original mediaStreamTrack for dummyEl
-      this._attach(this._dummyEl, this.mediaStreamTrack);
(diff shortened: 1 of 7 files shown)


lib/media/track/mediatrack.js
```diff
@@ -118,24 +119,29 @@ class MediaTrack extends Track {
    * @private
    */
   _initialize() {
-    const self = this;
+    try {
```
charliesantos, line 122: Do we still need to keep this try catch block?

lib/media/track/mediatrack.js
```diff
@@ -145,27 +151,35 @@ class MediaTrack extends Track {
   _end() {
     this._log.debug('Ended');
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
+      this._detachElementInternal(this._dummyEl);
+      this._dummyEl.remove();
+      this._dummyEl.srcObject = null;
       this._dummyEl.oncanplay = null;
+      this._dummyEl = null;
     }
```
charliesantos, line 159: Nice one @makarandp0 . So we are only detaching elements, not properly disposing them. I've learned that we also need to pause the element first, then, call load after setting the srcObject to null. Something like

```js
this._detachElementInternal(this._dummyEl);
this._dummyEl.pause();
this._dummyEl.remove();
this._dummyEl.srcObject = null;
this._dummyEl.oncanplay = null;
this._dummyEl.load();
this._dummyEl = null;
```

lib/media/track/mediatrack.js
```diff
@@ -145,27 +151,35 @@ class MediaTrack extends Track {
...
-    }
-    this._log.debug('Attempting to attach to element:', el);
-    el = this._attach(el);
-
-    if (this._shouldShimAttachedElements && !this._elShims.has(el)) {
-      const onUnintentionallyPaused = this._playPausedElementsIfNotBackgrounded
-        ? () => playIfPausedAndNotBackgrounded(el, this._log)
-        : null;
-      this._elShims.set(el, shimMediaElement(el, onUnintentionallyPaused));
+    try {
```
charliesantos, line 163: Is this try catch block for debugging? Should we remove this?

lib/media/track/mediatrack.js
```diff
@@ -145,27 +151,35 @@ class MediaTrack extends Track {
   _end() {
     this._log.debug('Ended');
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
+      this._detachElementInternal(this._dummyEl);
```
charliesantos, line 154: Why do we need to create a separate `_detachElementInternal` method? So on `_end`, calling `_detachElementInternal` will not update internal properties such as `_attachments` and `_elShims` right?

charliesantos (changes requested):

lib/media/track/mediatrack.js
makarandp0: we need seperate function because `_detachElement` is no-op for `_dummyEl` as its not added to `this._attachments`
It was a bug that caused us to never detach  `_dummyEl`

charliesantos: Ah good one.

lib/media/track/mediatrack.js
makarandp0: I think so. This fix ensures that we clean up the dummy element. but customers would still see the problem if they manage to create 75+ elements (possible on large rooms). In such cases its good to log the error for further confirmation and of the issue

lib/media/track/mediatrack.js
makarandp0: okay, I will update and test per your suggestion.

lib/media/track/mediatrack.js
makarandp0: I would keep this block, since the issue will still happen when 75+ elements are created.

lib/media/track/mediatrack.js
charliesantos: Should we throw a TwilioError instead? That way, it follows our convention when an error happens.

lib/media/track/mediatrack.js
charliesantos: Should we throw a TwilioError instead? That way, it follows our convention when an error happens.

makarandp0: TwilioErrors are heavy - require associated documentation and such. I think this error is fine for now. I am also hoping that this issue will get fixed by chrome as lot of apps are affected by it. https://bugs.chromium.org/p/chromium/issues/detail?id=1232649

lib/media/track/mediatrack.js
makarandp0: After experimenting a bit, I found that just setting srcObject to null is sufficient to avoid this issue. I will keep this simple and just reset `srcObject` as [suggested](https://bugs.chromium.org/p/chromium/issues/detail?id=1232649#c8) as a workaround:

lib/media/track/mediatrack.js
makarandp0: Update: I found that the try block is not very useful as the error happens async and not caught by the block :(. I will remove the try block.

lib/media/track/mediatrack.js
```diff
@@ -145,8 +145,10 @@ class MediaTrack extends Track {
   _end() {
     this._log.debug('Ended');
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
+      this._dummyEl.remove();
+      this._dummyEl.srcObject = null;
```
charliesantos, line 149: Just want to double check. Since we're not calling detachElement anymore, is setting srcObject to null enough, and we don't need to call `removeTrack`?
```js
    const mediaStream = this._dummyEl.srcObject;
    if (mediaStream instanceof this._MediaStream) {
      mediaStream.removeTrack(this.processedTrack || this.mediaStreamTrack);
    }
```

makarandp0: thats correct, removeTrack applies only for elements attached by application - since they may already have a mediaStream, in that case we add/remove track from it.

charliesantos (approved):

PR #1533, opened 2021-07-26
Small iOS Chrome Fixes

PikaJoyce:
This a relatively small PR:
1. Adding `isChromeIOS` to some other places I missed in the last PR to apply a few extra safari workarounds. 
2. Adding in an additional check to make sure that the browser is iOS Chrome 14.3+ since WebRTC WKWebView support didn't happen until iOS OS v14.3, and it does not seem that they will be fixing previous versions [as seen here](https://bugs.chromium.org/p/chromium/issues/detail?id=752458).
3. Failing Unit test cases added 

**Contributing to Twilio**

### DIFF
--- a/lib/util/support.js
+++ b/lib/util/support.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { guessBrowser, support: isWebRTCSupported } = require('@twilio/webrtc/lib/util');
-const { isAndroid, isMobile, isNonChromiumEdge, rebrandedChromeBrowser, mobileWebKitBrowser } = require('./browserdetection');
+const { isAndroid, isChromeIOS, isChromeVersionSupported, isMobile, isNonChromiumEdge, rebrandedChromeBrowser, mobileWebKitBrowser } = require('./browserdetection');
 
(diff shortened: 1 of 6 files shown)


lib/util/support.js
```diff
@@ -34,6 +34,13 @@ function isSupported() {
     return false;
   }
 
+  // NOTE (joma): Return if iOS chrome version is not 14.3 and above.
+  if (isChromeIOS()) {
+    if (!isChromeVersionSupported()) {
+      return false;
+    }
```
charliesantos, line 41: So we don't support chrome on ios 14.3 and above?

PikaJoyce: We do support chrome (more specifically WKWebView) on iOS 14.3 and above. If `isChromeVersionSupported` is false, then we will return false. But `getUserMedia` is not available in iOS versions below 14.3. But good point, maybe I should rename this util to something else if it helps make it clearer? How does something like `isiOSVersionSupported` sound? 🤔

PikaJoyce:
Had a chat with @charliesantos on this. I'll hold off on this making this PR for the future until we get some more tests in to confirm that iOS versions 14.3 will not work.

PR #1534, opened 2021-07-27
VIDEO-6336 - Cleanup dummy audio element after start

makarandp0:
We create dummy element for audio only to generate 'started' event. Once started we were calling `_detachElement`, but that call is no-op for dummy element - Causing the audio element active until track ended. Fixed this by cleaning up dummy element specifically.

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/audiotrack.js
+++ b/lib/media/track/audiotrack.js
@@ -35,9 +35,7 @@ class AudioTrack extends MediaTrack {
     super._start();
     if (this._dummyEl) {
       // once started let go of dummy element
-      this._dummyEl.remove();
       this._dummyEl.srcObject = null;
-      this._dummyEl.oncanplay = null;
       this._dummyEl = null;
     }
   }
--- a/lib/media/track/audiotrack.js
+++ b/lib/media/track/audiotrack.js
@@ -34,7 +34,11 @@ class AudioTrack extends MediaTrack {
   _start() {
     super._start();
     if (this._dummyEl) {
-      this._detachElement(this._dummyEl);
+      // once started let go of dummy element
+      this._dummyEl.remove();
+      this._dummyEl.srcObject = null;
+      this._dummyEl.oncanplay = null;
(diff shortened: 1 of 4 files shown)


charliesantos (approved):
Do we need to add any tests for this?

pethog77:
FYI....We have been testing this with the 2.3.13-dev build and found that the Video elements also need to have this applied to them. We dropped it in on mediatrack.js as well to remove the secondary placeholder which was not being unloaded:
`  /**
   * @private
   */
  _start() {
    this._log.debug('Started');
    this._isStarted = true;
    if (this._dummyEl) {
      this._dummyEl.remove();
      this._dummyEl.srcObject = null;
      this._dummyEl.oncanplay = null;
      this._dummyEl = null;
    }
    // eslint-disable-next-line no-use-before-define
    this.emit('started', this);
  }`

makarandp0:
> FYI....We have been testing this with the 2.15.3-dev build and found that the Video elements also need to have this applied to them. We dropped it in on mediatrack.js as well to remove the secondary placeholder which was not being unloaded:
> 
> ```
>    * @private
>    */
>   _start() {
>     this._log.debug('Started');
>     this._isStarted = true;
>     if (this._dummyEl) {
>       this._dummyEl.remove();
>       this._dummyEl.srcObject = null;
>       this._dummyEl.oncanplay = null;
>       this._dummyEl = null;
>     }
>     // eslint-disable-next-line no-use-before-define
>     this.emit('started', this);
>   }
> ```

@pethog77, besides 'started' - dummy video elements are also used to generate [dimensionsChanged event](https://github.com/twilio/twilio-video.js/blob/master/lib/media/track/videotrack.js#L206),  so we cannot clean those up after track is started. We do clean them up when track ends tho.

makarandp0:
@charliesantos, I have added a unit test for this now. Can you please take another look?

charliesantos (approved):

PR #1537, opened 2021-07-29
use audio jitter only

makarandp0:
although chrome reports jitter values for video tracks, they does not seem correct - they are too large to be in seconds. 
For now just use jitter from audio tracks.

**Contributing to Twilio**


charliesantos:
Is this something we are planning to investigate to understand why?

makarandp0:
Its seems like chrome issue - and since safari does not provide jitter for video tracks anyways, I think it's safer to just look at audio.

charliesantos:
Do we need to update any docs?

makarandp0:
> Do we need to update any docs?

Looks like our doc is not very specific which track we use for jitter - But I will update to make the units specific. While at it also noticed that bitrates were not in correct units. Fixed now.

charliesantos (approved):

PR #1538, opened 2021-07-30
edited changelog for preflight release

seancoleman2:
**Contributing to Twilio**


CHANGELOG.md
```diff
@@ -4,43 +4,16 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
+This release includes the **Preflight API Public Beta** (`runPreflight`) to help test connectivity with Twilio servers and estimate media quality for a Participant. It can be used to detect issues prior to joining a Video Room or as part of a troubleshooting page.  
 
-**Few things to note:**
-- This function uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API"> web  audio API's.</a>
-Browser's  autoplay policies sometimes require user action before accessing these APIs. Please ensure that this API
-is called in response to user action like a button click.
-- `preflightTest` emits `failed` event to indicate test failures. Please refer to
-[this guide](https://www.twilio.com/docs/video/build-js-video-application-recommendations-and-best-practices#connection-errors)
-for interpreting common errors.
+The API connects two peer connections using Twilio's TURN servers. It publishes synthetic audio and video tracks from one participant and ensures that other participant receives media on those tracks. After successfully verifying connectivity, it generates a report with statistics on the connection.
```
donaltoomey, line 12: Is it a peer connection to itself or two separate peer connections to itself? Seems like it is a single peer connection to itself. 
You should also mention that it connects to Twilio signaling servers.

CHANGELOG.md
```diff
@@ -64,6 +37,34 @@ preflightTest.on('completed', (report: PreflightTestReport) => {
...
+| 2                        | Fair
+| 3                        | Good
+| 4                        | Great
+| 5                        | Excellent
+
+
+A few things to note:
+
+- This function uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API"> web audio API's.</a> Browser's autoplay policies sometimes require user action before accessing these APIs. Please ensure that this API is called in response to user action like a button click.
+- `preflightTest` emits a `failed` event to indicate test failures. You can use the Preflight `progress` events to better understand where the test failed and refer to
```
seancoleman2, line 64: @makarandp0 I wanted to link to documentation outlining the progress events and noticed that the possible values are not documented here: https://sdk.stage.twilio.com/js/video/releases/2.17.0-rc2/docs/global.html#PreflightProgress

CHANGELOG.md
```diff
@@ -64,6 +37,34 @@ preflightTest.on('completed', (report: PreflightTestReport) => {
...
+| jitter                   | Packets delay variation on audio tracks | < 30ms
+| packetLoss               | Packet loss as a percent of total packets sent | < 1%
+| rtt                      | Round trip time in milliseconds. | < 200ms
+
+The API also calculates an overall `qualityScore` using jitter, packet loss, and rtt values observed. The quality score can be one of:
+
+| Value                    | Meaning
+| -------------------------|--------------------------------|
+| 0                        | Failed
+| 1                        | Degraded
```
donaltoomey, line 54: Is "poor" a better word than degraded?

CHANGELOG.md
```diff
@@ -64,6 +37,34 @@ preflightTest.on('completed', (report: PreflightTestReport) => {
...
+| 4                        | Great
+| 5                        | Excellent
+
+
+A few things to note:
+
+- This function uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API"> web audio API's.</a> Browser's autoplay policies sometimes require user action before accessing these APIs. Please ensure that this API is called in response to user action like a button click.
+- `preflightTest` emits a `failed` event to indicate test failures. You can use the Preflight `progress` events to better understand where the test failed and refer to
+[this guide](https://www.twilio.com/docs/video/build-js-video-application-recommendations-and-best-practices#connection-errors) for interpreting common errors.
+- Stats are based on the P2P connection. This means that the `outgoingBitrate` and `incomingBitrate` returned in the `stats` object are not a measure of uplink / downlink bandwidth.
```
donaltoomey, line 66: This bullet is a little confusing. For this loopback test, should the outgoingBitrate equal the incomingBitrate?

CHANGELOG.md
```diff
@@ -64,6 +37,34 @@ preflightTest.on('completed', (report: PreflightTestReport) => {
...
 ```
+
+The [PreflightTestReport](https://sdk.twilio.com/js/video/releases/2.17.0/docs/global.html#PreflightTestReport) generated by `preflightTest` on `completed` provides statistics that can be used to determine the quality of the connection. Some of the useful stats and their optimal values are listed below.
+
+| Stat                     |  Description | Optimal values
+| -------------------------|--------------|------------------|
+| jitter                   | Packets delay variation on audio tracks | < 30ms
+| packetLoss               | Packet loss as a percent of total packets sent | < 1%
+| rtt                      | Round trip time in milliseconds. | < 200ms
+
```
donaltoomey, line 48: Should there be a bitrate stat too?

CHANGELOG.md
makarandp0: its two peer connections. +1 on mentioning Twilio signaling and turn servers

CHANGELOG.md
makarandp0: we only use jitter/packetLoss and rtt to compute quality score - and they also have an "optimal value" associated with them. Although the generated report has bitrates - they are not used for any scoring, as we are not using full available bandwidth.

CHANGELOG.md
makarandp0: good point, I will add those in this PR.

CHANGELOG.md
makarandp0: yes, outgoing and incoming will  be mostly equal (excluding some packet loss).  @donaltoomey  Are you suggesting we just report overall bitrate instead of these two separate values?

CHANGELOG.md
donaltoomey: ok, what way do we use the peer connections? Are we sending video/audio on both?

CHANGELOG.md
donaltoomey: yeah, I think it is tied to my other question above about the 2 peer connections. At the end of the day we need to provide some indication of what the upstream bitrate is going to be like and what the downstream bitrate is going to be like when the participant connects to an actual Room. For example, if they are able to get say 1 Mbps upstream and downstream then they should be able to have a good Video call.

CHANGELOG.md
donaltoomey: ok, fair enough

CHANGELOG.md
makarandp0: we have to use two peerConnections to establish connectivity - one for Alice another for Bob - Alice's PC publishes media, Bob's PC subscribes to it.

CHANGELOG.md
makarandp0: Preflight test  publishes one audio and video track. This does not give good indication of bandwidth availability. I think customer needs separate bandwidth test. I believe we already have such test in Diagnostic SDK that uses data channels - Or am I missing something @charliesantos ? )

CHANGELOG.md
makarandp0: I have fixed this in https://github.com/twilio/twilio-video.js/pull/1538/commits/2b2cf0e436612378748f73ec6cb3d796545cd3e1 
Basically it turned out that jsdocs does not like typescript enum - as those are stripped from the generated code. Converted it to const like rest of the code does. With this you should see those in generated doc [here](https://49575-46595751-gh.circle-artifacts.com/0/dist/docs/global.html#PreflightProgress)

CHANGELOG.md
donaltoomey: ok, we may need a call at some stage to discuss. I think what we really want to tell end users is what kind of bitrate they can expect when connected to the Twilio Cloud. Speed tests likely have edge caches and use TCP so are not really representative of what video/audio bitrate you are going to get.

charliesantos: Right, we have bitrate tests that uses data channels (between to peer connections) while connected to Twilio Turn Servers.

lib/preflight/preflighttest.ts
```diff
@@ -21,48 +21,42 @@ const DEFAULT_TEST_DURATION = 10 * SECOND;
  * progress values that are sent by {@link PreflightTest#event:progress}
  * @enum {string}
  */
-enum PreflightProgress {
+const PreflightProgress = {
   /**
-   * Preflight test {@link PreflightTest} has successfully acquired media
+   * {@link PreflightTest} has successfully generated synthetic tracks
```
charliesantos, line 26: The name of the event is 'mediaAcquired', I think it's ok to mention `...has successfully acquired media` right? Mentioning `generated synthetic tracks` might be more confusing to the customer.

CHANGELOG.md
```diff
@@ -4,43 +4,16 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
-**Few things to note:**
-- This function uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API"> web  audio API's.</a>
-Browser's  autoplay policies sometimes require user action before accessing these APIs. Please ensure that this API
-is called in response to user action like a button click.
-- `preflightTest` emits `failed` event to indicate test failures. Please refer to
-[this guide](https://www.twilio.com/docs/video/build-js-video-application-recommendations-and-best-practices#connection-errors)
-for interpreting common errors.
+The API connects two peer connections using Twilio's TURN servers. It publishes synthetic audio and video tracks from one participant and ensures that other participant receives media on those tracks. After successfully verifying connectivity, it generates a report with statistics on the connection.
 
+`runPreflight` was originally introduced as an experimental API in `2.8.0-beta1` and has been updated based on feedback. In short, usage of the API will now be free of charge.
```
charliesantos, line 14: Is using TURN servers for video customers free? Or do they pay after exceeding a certain threshold?

donaltoomey: it is free for Programmable Video customers

donaltoomey (approved):
LGTM

makarandp0 (approved):

PR #1551, opened 2021-08-16
iOS Chrome Fixes

PikaJoyce:
This is a WIP PR, however I think there are some various changes that may be worth reviewing a bit earlier for some context.

This PR addresses some of our current large bugs [VIDEO-6073](https://issues.corp.twilio.com/browse/VIDEO-6073) and [VIDEO-5884](https://issues.corp.twilio.com/browse/VIDEO-5884). Safari is not supporting legacy WebRTC, therefore, I added `.addTransceivers` prior to making an offer in order to produce a full SDP from iOS Chrome. This fixes iOS Chrome participants inability to subscribe if they are not publishing.

@DenissSazanovs has also confirmed that gUM is not available to devices on iOS 14.2 and below. There is an added util to catch this that has been added to the `isSupported` flag as well as the unit tests. I've also added some future proofing to allow for some leeway for future versions to also return `true` within the regex.

Some to-dos:

- [ ]  Write some unit tests for added `addTransceivers`
- [ ] Investigate why Safari works but not iOS Chrome

**Contributing to Twilio**


lib/util/browserdetection.js
```diff
@@ -27,6 +27,15 @@ function isChromeIOS() {
...
 }
 
+/**
+ * Check whether the current OS version is 14_3 and above,
+ * used in conjunction with isChromeIOS and isMobile.
+ * @returns {boolean}
+ */
+function isChromeVersionSupported() {
+  return /14_([3-9]|[1-9]\d+)|(1[5-9]|[2-9]\d+)_\d+/.test(navigator.userAgent);
+}
```
manjeshbhargav, line 37: @PikaJoyce Do we need this check? Since WebRTC APIs are not available in Chrome on iOS 14.2 and below, doesn't `isWebRTCSupported()` already return false?

lib/util/support.js
```diff
@@ -34,6 +34,13 @@ function isSupported() {
     return false;
   }
 
+  // NOTE (joma): Return if iOS chrome version is not 14.3 and above.
+  if (isChromeIOS()) {
+    if (!isChromeVersionSupported()) {
+      return false;
+    }
+  }
```
manjeshbhargav, line 42: @PikaJoyce Do we need this check?

lib/signaling/v2/peerconnection.js
```diff
@@ -830,7 +832,7 @@ class PeerConnectionV2 extends StateMachine {
         // support, we have to use the same hacky solution as Safari. Revisit
         // this when RTCRtpTransceivers and MIDs land. We should be able to use
         // the same technique as Firefox.
-        : isSafari || this._isUnifiedPlan ? new OrderedTrackMatcher() : new IdentityTrackMatcher();
+        : isSafari || isiOSChrome || this._isUnifiedPlan ? new OrderedTrackMatcher() : new IdentityTrackMatcher();
```
manjeshbhargav, line 835:
```suggestion
        : isSafari || isChromeIOS || this._isUnifiedPlan ? new OrderedTrackMatcher() : new IdentityTrackMatcher();
```

manjeshbhargav (changes requested):
@PikaJoyce I have left some comments. Regarding adding the transceivers, please do so in the `twilio-webrtc.js` project. You can refer to the [SafariRTCPeerConnection.createOffer](https://github.com/twilio/twilio-webrtc.js/blob/master/lib/rtcpeerconnection/safari.js#L158) shim in order to figure out how to do it.

lib/util/support.js
PikaJoyce: Went to double check, it seems that we don't! Thanks Manjesh!

lib/signaling/v2/peerconnection.js
```diff
@@ -323,7 +325,7 @@ class PeerConnectionV2 extends StateMachine {
           || preferredCodecs.audio.some(({ codec, dtx }) => codec === 'opus' && dtx)
       },
       _shouldApplySimulcast: {
-        value: (isChrome || isSafari) && preferredCodecs.video.some(
+        value: (isChrome || isSafari || isChromeIOS) && preferredCodecs.video.some(
```
manjeshbhargav, line 327: I think `isChrome` will cover iOS chrome as well here:
```suggestion
        value: (isChrome || isSafari) && preferredCodecs.video.some(
```

lib/signaling/v2/peerconnection.js
```diff
@@ -59,9 +59,11 @@ const IdentityTrackMatcher = require('../../util/sdp/trackmatcher/identity');
 const OrderedTrackMatcher = require('../../util/sdp/trackmatcher/ordered');
 const MIDTrackMatcher = require('../../util/sdp/trackmatcher/mid');
 const workaroundIssue8329 = require('../../util/sdp/issue8329');
+const { isChromeIOS } = require('../../util/browserdetection');
 
 const guess = guessBrowser();
 const platform = getPlatform();
+const isiOSChrome = isChromeIOS();
```
manjeshbhargav:
```suggestion
```

lib/signaling/v2/peerconnection.js
manjeshbhargav: @PikaJoyce `isUnifiedPlan` will cover the iOS chrome case, so we don't need the extra check.
```suggestion
        : isSafari || this._isUnifiedPlan ? new OrderedTrackMatcher() : new IdentityTrackMatcher();
```

lib/util/browserdetection.js
manjeshbhargav:
```suggestion
/**
 * Check whether the current iOS version is 14_3 and above,
 * used in conjunction with isChromeIOS and isMobile.
 * @returns {boolean}
 */
function isIOSVersionSupported() {
  return /14_([3-9]|[1-9]\d+)|(1[5-9]|[2-9]\d+)_\d+/.test(navigator.userAgent);
}
```

lib/util/browserdetection.js
```diff
@@ -134,6 +143,7 @@ module.exports = {
   isAndroid,
   isIOS,
   isChromeIOS,
+  isChromeVersionSupported,
```
manjeshbhargav:
```suggestion
  isIOSVersionSupported,
```

lib/util/support.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const { guessBrowser, support: isWebRTCSupported } = require('@twilio/webrtc/lib/util');
-const { isAndroid, isMobile, isNonChromiumEdge, rebrandedChromeBrowser, mobileWebKitBrowser } = require('./browserdetection');
+const { isAndroid, isChromeIOS, isChromeVersionSupported, isMobile, isNonChromiumEdge, rebrandedChromeBrowser, mobileWebKitBrowser } = require('./browserdetection');
```
manjeshbhargav:
```suggestion
const { isAndroid, isChromeIOS, isIOSVersionSupported, isMobile, isNonChromiumEdge, rebrandedChromeBrowser, mobileWebKitBrowser } = require('./browserdetection');
```

lib/util/support.js
```diff
@@ -34,6 +34,13 @@ function isSupported() {
     return false;
   }
 
+  // NOTE (joma): Return if iOS chrome version is not 14.3 and above.
+  if (isChromeIOS()) {
+    if (!isChromeVersionSupported()) {
```
manjeshbhargav:
```suggestion
    if (!isIOSVersionSupported()) {
```

manjeshbhargav (changes requested):

lib/util/support.js
PikaJoyce: @manjeshbhargav Hm, I've removed this change in a recent commit because you mentioned that `isWebRTCSupported` will catch this. Do you think we should add an `iOSVersionSupported` check still?

manjeshbhargav (commented):
Please use `isIOSChrome` from `@twilio/webrtc` as follows:
```js
const { isIOSChrome } = require('@twilio/webrtc/util');
```
Then, you can get rid of `isChromeIOS`.

PR #1624, opened 2021-11-02
VIDEO-6192 / 6188 - Rollbar error fixes

PikaJoyce:
JIRA Tickets:
[VIDEO-6188](https://issues.corp.twilio.com/browse/VIDEO-6188)
[VIDEO-6192](https://issues.corp.twilio.com/browse/VIDEO-6192)

Addresses some of the errors we're seeing in rollbar. Some types (name/roomName) that were meant to be a string came in as integers and others (maxAudioBitrate/maxVideoBitrate) that were meant to be integers could come in as strings. Also added in some extra unit tests to capture and verify these changes.

**Contributing to Twilio**

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -588,11 +588,23 @@ function createRoomConnectEventPayload(connectOptions) {
     payload[eventProp] = boolToString(!!connectOptions[prop]);
   });
 
-  // numbers and string properties.
-  [['maxVideoBitrate'], ['maxAudioBitrate'], ['iceTransportPolicy'], ['region'], ['name', 'roomName']].forEach(([prop, eventProp]) => {
+  // numbers properties.
+  [['maxVideoBitrate'], ['maxAudioBitrate']].forEach(([prop, eventProp]) => {
     eventProp = eventProp || prop;
-    if (typeof connectOptions[prop] === 'number' || typeof connectOptions[prop] === 'string') {
+    if (typeof connectOptions[prop] === 'number') {
       payload[eventProp] = connectOptions[prop];
+    } else if (typeof connectOptions[prop] === 'string') {
+      payload[eventProp] = Number(connectOptions[prop]);
+    }
+  });
+
+  // string properties.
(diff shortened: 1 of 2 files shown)


lib/util/index.js
```diff
@@ -588,11 +588,23 @@ function createRoomConnectEventPayload(connectOptions) {
...
-  [['maxVideoBitrate'], ['maxAudioBitrate'], ['iceTransportPolicy'], ['region'], ['name', 'roomName']].forEach(([prop, eventProp]) => {
+  // numbers properties.
+  [['maxVideoBitrate'], ['maxAudioBitrate']].forEach(([prop, eventProp]) => {
     eventProp = eventProp || prop;
-    if (typeof connectOptions[prop] === 'number' || typeof connectOptions[prop] === 'string') {
+    if (typeof connectOptions[prop] === 'number') {
       payload[eventProp] = connectOptions[prop];
+    } else if (typeof connectOptions[prop] === 'string') {
+      payload[eventProp] = Number(connectOptions[prop]);
+    }
```
makarandp0, line 598: I think we need to ensure that we do not send 'NaN' values.
```suggestion
    } else if (typeof Number(connectOptions[prop]) === 'number') {
      payload[eventProp] = Number(connectOptions[prop]);
    }
```

test/unit/spec/util/index.js
```diff
@@ -74,25 +74,40 @@ describe('util', () => {
...
+        connectOptions: { name: 1234 },
+        expectedPayload: { roomName: '1234' },
+      },
       {
         testCase: 'region specified',
         connectOptions: { region: 'in1' },
         expectedPayload: { region: 'in1' },
       },
       {
-        testCase: 'maxVideoBitrate specified',
```
makarandp0, line 87: we should add a test case where bitrate is specified as non-number string -  say "foo"

makarandp0 (commented):
Joyce, looks like we accept bitrates as string values...I wonder what happens when they result in `NaN` do they get sent to gateway ? does the connect succeed?

makarandp0 (approved):
LGTM, you might still want to hold off on unit test fixes before merging.

PR #1639, opened 2021-11-15
Video-6677 consider PIP windows for render hints.

makarandp0:
SDK sends hints about video elements considering their visibility and dimensions to SFU. Those hints were not considering picture-in-picture windows. So if PIP window is open and a tab is switched or video element becomes invisible, the track would get switched off. This PIP handling fixes that. 
**Contributing to Twilio**

### DIFF
--- a/lib/media/track/remotevideotrack.js
+++ b/lib/media/track/remotevideotrack.js
@@ -171,7 +171,7 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
     return this;
   }
 
-  unObserverPIP(el) {
+  _unObservePip(el) {
     const pipCallbacks = this._elToPipCallbacks.get(el);
     if (pipCallbacks) {
       el.removeEventListener('enterpictureinpicture', pipCallbacks.onEnterPip);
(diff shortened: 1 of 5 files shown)


charliesantos (commented):
Implementation looks good. But have we considered checking if there's a pictureInPictureElement on the document object instead? It might be less expensive since we are not attaching events and it might make this simpler. Thoughts?

xtianjohns:
I've verified that this fix works for me locally, does not pause PIP video content when I change tabs.

makarandp0:
> Implementation looks good. But have we considered checking if there's a pictureInPictureElement on the document object instead? It might be less expensive since we are not attaching events and it might make this simpler. Thoughts?

@charliesantos I considered that approach, but ruled it out because [the spec](https://w3c.github.io/picture-in-picture/#one-pip-window) leaves room for multiple windows for PIP: 
> Whether only one window is allowed in Picture-in-Picture mode will be left to the implementation and the platform. However, because of the one Picture-in-Picture window limitation, the specification assumes that a given Document can only have one Picture-in-Picture window.

Yeah, registering callbacks adds to code, but they all get cleaned up.

lib/media/track/remotevideotrack.js
```diff
@@ -165,6 +171,50 @@ class RemoteVideoTrack extends RemoteMediaVideoTrack {
     return this;
   }
 
+  unObserverPIP(el) {
```
charliesantos, line 174: nit: how about `unObservePip`? That way it's consistent with how you name other Pip related things in this PR. Same for it's counterpart below, `observePip`.

charliesantos (changes requested):

charliesantos:
> > Implementation looks good. But have we considered checking if there's a pictureInPictureElement on the document object instead? It might be less expensive since we are not attaching events and it might make this simpler. Thoughts?
> 
> @charliesantos I considered that approach, but ruled it out because [the spec](https://w3c.github.io/picture-in-picture/#one-pip-window) leaves room for multiple windows for PIP:
> 
> > Whether only one window is allowed in Picture-in-Picture mode will be left to the implementation and the platform. However, because of the one Picture-in-Picture window limitation, the specification assumes that a given Document can only have one Picture-in-Picture window.
> 
> Yeah, registering callbacks adds to code, but they all get cleaned up.

Ok, lgtm. Just left one comment.

lib/media/track/remotevideotrack.js
makarandp0: Thanks @charliesantos, I have now renamed the functions + added some basic unit test and added a changelog entry.

charliesantos (approved):

PR #1560, opened 2021-08-31
VIDEO-6263: low fps from safari with simulcast enabled. 

makarandp0:
The default simulcast layers chosen by safari have same dimensions. This causes low framerate for tracks from safari. This 
change configures the layers with with `scaleResolutionDownBy` set to 1, 2, and 4. This is still pending design task: VIDEO-6730
  
**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

PR #1641, opened 2021-11-15
VIDEO-7790: safari replacetrack should reset layers per track dimensions

makarandp0:
We re-enable simulcast layers (and let SFU know about it) when a track is replaced. 
However since we configure [simulcast layers  on safari depending on track dimensions](https://github.com/twilio/twilio-video.js/pull/1560) - the reset encodings need to reconfigure the layers based on replaced Tracks dimensions. This change ensures that replaceTrack does not enable layers that are were not supposed to be enabled to begin with.  

**Contributing to Twilio**

### DIFF
--- a/lib/media/track/sender.js
+++ b/lib/media/track/sender.js
@@ -93,6 +93,11 @@ class MediaTrackSender extends MediaTrackTransceiver {
     return this;
   }
 
+  /**
+   * applies given encodings, or resets encodings if none specified.
+   * @param {Array<{enabled: boolean, layer_index: number}>|null} encodings
+   * @returns {Promise<string>}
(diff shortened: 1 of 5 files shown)


lib/media/track/sender.js
```diff
@@ -93,6 +93,11 @@ class MediaTrackSender extends MediaTrackTransceiver {
     return this;
   }
 
+  /**
+   * applies given encodings, or resets encodings if none specified.
```
manjeshbhargav, line 97:
```suggestion
   * Applies given encodings, or resets encodings if none specified.
```

manjeshbhargav (approved):

PR #1644, opened 2021-11-17
VIDEO-7799 Remove cached access token in TwilioConnectionTransport

manjeshbhargav:
**Contributing to Twilio**

### DIFF
--- a/lib/signaling/v2/twilioconnectiontransport.js
+++ b/lib/signaling/v2/twilioconnectiontransport.js
@@ -94,13 +94,10 @@ class TwilioConnectionTransport extends StateMachine {
       renderHints: true,
       userAgent: getUserAgent()
     }, options);
+
     super('connecting', states);
 
-
     Object.defineProperties(this, {
-      _accessToken: {
-        value: accessToken
-      },
       _automaticSubscription: {
         value: options.automaticSubscription
       },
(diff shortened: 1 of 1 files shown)


lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -628,4 +552,109 @@ function setupTransport(transport) {
...
+      _adaptiveSimulcast,
+      _renderHints);
+
+    message.subscribe = createSubscribePayload(_automaticSubscription);
+
+    const { sdpFormat } = _options;
+    if (sdpFormat) {
+      message.format = sdpFormat;
+    }
+    message.token = accessToken;
```
makarandp0, line 634: @manjeshbhargav, Can you give us some context on the purpose of this change. If it's about hiding access token from the user, wouldn't they be able to see it from the network tab anyways?

manjeshbhargav: @makarandp0 Since we are not charging for the Media Extension Participant minutes, there is a danger that the Custom Media Extension code, which is managed by the developer, can access this token and use it to join Rooms for free.

manjeshbhargav: @makarandp0 The network tab thing won't be a problem here since the media extension is running in a CEF environment on Twilio's cloud infrastructure.

charliesantos: Do we have safeguards on the backend in case the token is used outside of the media extension participant?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -179,93 +176,20 @@ class TwilioConnectionTransport extends StateMachine {
...
-
-      const sdpFormat = this._options.sdpFormat;
-      if (sdpFormat) {
-        message.format = sdpFormat;
+    // NOTE(mmalavalli): In order to hide the AccessToken from being available
+    // to the application code, we define those member functions that make use
+    // of it here.
+    Object.defineProperties(this, {
+      _createConnectOrSyncOrDisconnectMessage: {
+        value: () => createConnectOrSyncOrDisconnectMessage(this, accessToken)
```
charliesantos, line 184: How about
```js
value: createConnectOrSyncOrDisconnectMessage.bind(this, accessToken)
```

or
```js
value: () => createConnectOrSyncOrDisconnectMessage.call(this, accessToken)
```

This way, you don't need to pass the transport object and don't have to do the big destructuring in line 564

charliesantos (changes requested):

lib/signaling/v2/twilioconnectiontransport.js
manjeshbhargav: @charliesantos I'd like to avoid using `bind`, and be more explicit in defining the arguments required for a particular function. Otherwise, it is a bit confusing to read the function, where `this` is being used even though it is not a member of any class.

lib/signaling/v2/twilioconnectiontransport.js
makarandp0: how does the Media Extension Participant get this access token in the 1st place - That is how do they call connect?

manjeshbhargav: @charliesantos @makarandp0 I believe the token is internally acquired by the MediaProcessor service and passed to the CEF context, where the token is used by an internal JS function to join a Room and then the Room object is passed to the application code. That is why we don't want any reference to the token in the Room object.

charliesantos (approved):

Issue #1647, opened 2021-11-18
The `connect` method does not return a `Promise` despite Typescript type definitions saying so


Mange:
- [X] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [X] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [X] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [X] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

```js
connect("access-token-here", {name: "room-name"}).finally(() => alert("Finally"));
```

**Expected behavior:**

An alert should be shown after the connection is fulfilled or errors. Typescript type definitions say that `connect` should return a `Promise`.

**Actual behavior:**

Uncaught `TypeError` because `finally` is not a method on `CancelablePromise`, which is returned instead of a `Promise`.

Either declaring the types of `connect` correctly or implementing `finally` is a good solution to me, but these kinds of issues will keep happening if the types indicate a `Promise` and the library doesn't return a `Promise` (or subclass of it), so I think that solution is more correct.
(I wouldn't mind a `finally` too, of course! :smile: )

PikaJoyce:
Hi @Mange,

Thank you for submitting this issue. You're right in that `connect` is supposed to return a `CancelablePromise` instead of a `Promise`. I've created an internal ticket to fix this (VIDEO-7831). I'll keep this issue open and keep it updated as we progress with this ticket.

Thanks again!
Joyce

PikaJoyce:
Hi @Mange !

The fix for this issue has been merged into develop so I'll be closing this issue now. Let me know if you've got any further questions or issues. And please don't hesitate to open another issue with us if that's the case :) 

Thanks again!
Joyce

Mange:
Thank you. Very nice work! :heart:

PR #1648, opened 2021-11-18
VIDEO-7831 Fix Connect to return CancelablePromise on Type Definitions

PikaJoyce:
This PR fixes an [issue](https://github.com/twilio/twilio-video.js/issues/1647) that was reported by a user. Currently our `connect` method returns a `Promise`, but it's supposed to return a `CancelablePromise`.

**Contributing to Twilio**

### DIFF
--- a/lib/util/cancelablepromise.js
+++ b/lib/util/cancelablepromise.js
@@ -110,6 +110,18 @@ class CancelablePromise {
       promise.then(...args).then(resolve, reject);
     }, this._onCancel);
   }
+
+  /**
+ * @param {?function} onResolved
+ * @returns {CancelablePromise}
+ */
+  finally() {
+    const args = [].slice.call(arguments);
+    const promise = this._promise;
+    return new CancelablePromise(function onCreate() {
+      promise.finally(...args);
+    }, this._onCancel);
+  }
 }
 
 module.exports = CancelablePromise;
--- a/lib/util/cancelablepromise.js
+++ b/lib/util/cancelablepromise.js
@@ -112,7 +112,7 @@ class CancelablePromise {
   }
 
   /**
- * @param {?function} onResolved
+ * @param {?function} onFinally
  * @returns {CancelablePromise}
  */
   finally() {
--- a/lib/util/cancelablepromise.js
+++ b/lib/util/cancelablepromise.js
@@ -118,8 +118,8 @@ class CancelablePromise {
   finally() {
     const args = [].slice.call(arguments);
     const promise = this._promise;
-    return new CancelablePromise(function onCreate() {
-      promise.finally(...args);
+    return new CancelablePromise(function onCreate(resolve, reject) {
+      promise.finally(...args).then(resolve, reject);
     }, this._onCancel);
   }
 }
--- a/lib/util/cancelablepromise.js
+++ b/lib/util/cancelablepromise.js
@@ -110,18 +110,6 @@ class CancelablePromise {
       promise.then(...args).then(resolve, reject);
     }, this._onCancel);
   }
-
-  /**
-   * @param {?function} onResolved
-   * @returns {CancelablePromise}
-   */
-  finally() {
-    const args = [].slice.call(arguments);
-    const promise = this._promise;
-    return new CancelablePromise(function onCreate() {
-      promise.finally(...args);
-    }, this._onCancel);
-  }
 }
 
 module.exports = CancelablePromise;
--- a/lib/util/cancelablepromise.js
+++ b/lib/util/cancelablepromise.js
@@ -110,6 +110,18 @@ class CancelablePromise {
       promise.then(...args).then(resolve, reject);
     }, this._onCancel);
   }
+
+  /**
+   * @param {?function} onResolved
+   * @returns {CancelablePromise}
+   */
+  finally() {
+    const args = [].slice.call(arguments);
+    const promise = this._promise;
+    return new CancelablePromise(function onCreate() {
+      promise.finally(...args);
+    }, this._onCancel);
+  }
 }
 
 module.exports = CancelablePromise;
--- a/test/integration/spec/connect.js
+++ b/test/integration/spec/connect.js
@@ -744,11 +744,10 @@ describe('connect', function() {
       let room;
       let errorThrown = null;
       try {
-        room = await cancelablePromise;
+        room = await cancelablePromise.finally(onFinally);
       } catch (error) {
         errorThrown = error;
       }
-      await cancelablePromise.finally(onFinally);
       sinon.assert.calledOnce(onFinally);
       assert(!errorThrown);
       assert(room instanceof Room);
@@ -756,13 +755,14 @@ describe('connect', function() {
 
     it('should reject and finally is called', async () => {
       const onFinally = sinon.stub();
+      let err;
       try {
-        await cancelablePromise;
+        await cancelablePromise.finally(onFinally);
         throw new Error('Connecting to room, but expecting to cancel');
       } catch (error) {
-        assert(error);
+        err = error;
       }
-      await cancelablePromise.finally(onFinally);
+      assert(err);
       sinon.assert.calledOnce(onFinally);
     });
   });
--- a/test/integration/spec/connect.js
+++ b/test/integration/spec/connect.js
@@ -720,6 +720,27 @@ describe('connect', function() {
     });
   });
 
+  describe('called with a Room name and', () => {
+    let sid;
+    let cancelablePromise;
+
+    before(async () => {
+      sid = await createRoom(randomName(), defaults.topology);
+      const options = Object.assign({ name: sid, tracks: [] }, defaults);
+      cancelablePromise = connect(getToken(randomName()), options);
+    });
+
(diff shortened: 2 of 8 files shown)


tsdef/index.d.ts
```diff
@@ -10,7 +10,7 @@ import { Room } from './Room';
 export const isSupported: boolean;
 export const version:string;
 export const Logger: Log.RootLogger;
-export function connect(token: string, options?: ConnectOptions): Promise<Room>;
+export function connect(token: string, options?: ConnectOptions): CancelablePromise<Room>;
```
makarandp0, line 13: is the definition of CancelablePromise public? Shouldn't it be imported in this file?

makarandp0: Please  include definition of cancellable promise like:
```
interface CancelablePromise<T> extends Promise<T> {
  cancel: () => void
}
```
also include the usage in tests.

makarandp0 (changes requested):

makarandp0:
@PikaJoyce, Looks like the issue you mentioned is also asking for `.finally` to work with `CancelablePromise` - if we extend from regular Promise, we also need to ensure that `finally` orks.

tsdef/index.d.ts
PikaJoyce: I was thinking of actually writing up the CancelablePromise type def but I wasn't sure since it's not public. I noticed that our CancelablePromise isn't actually extending the class `Promise` as well.

makarandp0: good point about `CancelablePromise` not extending from `Promise`.  In that case we can add its definition and exclude finally. But It would be better if we supported finally (and really extended it from Promise) - Maybe this should be part of convert `connect` to typescript work

tsdef/types.d.ts
```diff
@@ -258,3 +258,7 @@ export class StatsReport {
   remoteAudioTrackStats: RemoteAudioTrackStats[];
   remoteVideoTrackStats: RemoteVideoTrackStats[];
 }
+interface CancelablePromise<T> extends Promise<T> {
+  cancel: () => void;
+  finally(onfinally?: (() => void) | undefined | null): Promise<T>
```
makarandp0, line 263: is finally necessary here give that its already part of Promise?

tsdef/twilio-video-tests.ts
```diff
@@ -230,6 +230,15 @@ let room: Video.Room | null = null;
 let localVideoTrack: Video.LocalVideoTrack | null = null;
 let localAudioTrack: Video.LocalAudioTrack | null = null;
 
+// Testing finally method
+Video.connect('$TOKEN', {
```
makarandp0, line 234: this looks good! Please add an integration test as well.

makarandp0 (changes requested):

lib/util/cancelablepromise.js
```diff
@@ -110,6 +110,18 @@ class CancelablePromise {
       promise.then(...args).then(resolve, reject);
     }, this._onCancel);
   }
+
+  /**
+ * @param {?function} onResolved
```
charliesantos, line 115:
```suggestion
 * @param {?function} onFinally
```

test/integration/spec/connect.js
```diff
@@ -720,6 +720,58 @@ describe('connect', function() {
...
+      const options = Object.assign({ name: sid, tracks: [] }, defaults);
+      cancelablePromise = connect(getToken(randomName()), options);
+    });
+
+    after(() => completeRoom(sid));
+
+    it('should return a promise', async () => {
+      let finallyCalled = false;
+      await cancelablePromise.finally(() => {
+        finallyCalled = true;
```
charliesantos, line 738: Instead of using booleans, how about using stubs or spies? Something like

```js
const onFinally = sinon.stub();
await cancelablePromise.finally(onFinally);
sinon.assert.calledOnce(onFinally);
```

test/integration/spec/connect.js
```diff
@@ -720,6 +720,58 @@ describe('connect', function() {
...
+        room = await cancelablePromise;
+      } catch (error) {
+        errorThrown = error;
+      }
+      await cancelablePromise.finally(() => {
+        finallyCalled = true;
+      });
+      assert(finallyCalled);
+      assert(!errorThrown);
+      assert(room instanceof Room);
```
charliesantos, line 757: We should be using stubs or spies. Can you please check if this works?

```js
const onFinally = sinon.stub();
const onError = sinon.stub();
const room = cancelablePromise.catch(onError).finally(onFinally);

sinon.assert.calledOnce(onFinally);
sinon.assert.notCalled(onError);
assert(room instanceof Room);
```

test/integration/spec/connect.js
```diff
@@ -720,6 +720,58 @@ describe('connect', function() {
...
+        await cancelablePromise;
+        throw new Error('Connecting to room, but expecting to cancel');
+      } catch (error) {
+        assert(error);
+      }
+      await cancelablePromise.finally(() => {
+        finallyCalled = true;
+      });
+      assert(finallyCalled);
+    });
```
charliesantos, line 772: Same for this test. Please use stubs.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "checkJs": false,
     "declaration": false,
     "target": "es5",
-    "lib": ["es5", "es6", "es2016", "es2019", "es2018.promise", "dom"],
```
charliesantos, line 7: Why did we remove this? Please check our jsdocs still works after removing.

tsdef/twilio-video-tests.ts
```diff
@@ -230,6 +230,15 @@ let room: Video.Room | null = null;
...
+  name: 'room-name',
+  video: false,
+  audio: false
+}).finally(() => {
```
charliesantos, line 238: Do we have tests for `catch` and `then` in this file? It might be good to chain them here.

charliesantos (changes requested):

test/integration/spec/connect.js
PikaJoyce: @charliesantos I noticed that we weren't using sinons in any of our integration tests so I wasn't sure if I should add it for just this which is why I went the boolean route. I can absolutely use a spy though, do you think that'd be okay?

charliesantos: @PikaJoyce yes I think it's better. Check logger.js integration tests for reference.

makarandp0:
@PikaJoyce, I see that there are no unit tests for `CancelablePromise` - since we are updating the behavior, its good time to add some unit tests to ensure that `then`, `catch` `finally` all work as expected.

tsconfig.json
PikaJoyce: @charliesantos it was removed since I found that es2019 contains everything es2018 and so on and so forth. But how do you double check jsdocs are still working? I've ran `build:quick` and `docs` with no errors but I'm not sure if that's enough.

charliesantos: @PikaJoyce after running `build:quick` or `docs`, please check the generated docs under dist folder. Open the index.html file and make sure the docs are rendered properly. You need to check docs generated from a .ts file and .js file.

PikaJoyce: They rendered fine!

test/integration/spec/connect.js
```diff
@@ -720,6 +721,52 @@ describe('connect', function() {
...
+    it('should resolve with a room and finally is called', async () => {
+      const onFinally = sinon.stub();
+      let room;
+      let errorThrown = null;
+      try {
+        room = await cancelablePromise;
+      } catch (error) {
+        errorThrown = error;
+      }
+      await cancelablePromise.finally(onFinally);
```
charliesantos, line 751: should we chain this on line 747 instead?

test/integration/spec/connect.js
```diff
@@ -720,6 +721,52 @@ describe('connect', function() {
...
+
+    it('should reject and finally is called', async () => {
+      const onFinally = sinon.stub();
+      try {
+        await cancelablePromise;
+        throw new Error('Connecting to room, but expecting to cancel');
+      } catch (error) {
+        assert(error);
+      }
+      await cancelablePromise.finally(onFinally);
```
charliesantos, line 765: Same here, maybe chain the `finally` on line 760?

test/integration/spec/connect.js
```diff
@@ -720,6 +721,52 @@ describe('connect', function() {
...
+      assert(room instanceof Room);
+    });
+
+    it('should reject and finally is called', async () => {
+      const onFinally = sinon.stub();
+      try {
+        await cancelablePromise;
+        throw new Error('Connecting to room, but expecting to cancel');
+      } catch (error) {
+        assert(error);
```
charliesantos, line 763: I don't think we should perform assertions inside a catch block. There are chances that catch block is not executed, skipping your assertion. You can do the same pattern you did on the previous test above instead.

charliesantos (changes requested):

charliesantos (approved):

charliesantos:
@PikaJoyce I think we forgot to add a changelog

PR #1650, opened 2021-11-20
VIDEO-7800 Add function to connect as a media extension Participant.

manjeshbhargav:
**Contributing to Twilio**

### DIFF
--- a/lib/connectasmediaextension.js
+++ b/lib/connectasmediaextension.js
@@ -1,0 +1,23 @@
+/* globals context */
+'use strict';
+
+const CancelablePromise = require('./util/cancelablepromise');
+const { typeErrors: E } = require('./util/constants');
+
+/**
+ * Connect as a media extension {@link Participant}.
+ * @param {ConnectOptions} [options]
+ * @returns {CancelablePromise<Room>}
+ */
+function connectAsMediaExtension(options = {}) {
+  if (typeof context === 'object' && typeof context.connectAsMediaExtension === 'function') {
+    return connectAsMediaExtension(options);
+  }
+  // eslint-disable-next-line new-cap
+  return CancelablePromise.reject(E.ILLEGAL_INVOKE(
+    'connectAsMediaExtension',
+    'The context is not a media extension'
+  ));
+}
+
+module.exports = connectAsMediaExtension;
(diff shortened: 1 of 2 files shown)


lib/connectasmediaextension.js
```diff
@@ -0,0 +1,23 @@
...
+    return connectAsMediaExtension(options);
+  }
+  // eslint-disable-next-line new-cap
+  return CancelablePromise.reject(E.ILLEGAL_INVOKE(
+    'connectAsMediaExtension',
+    'The context is not a media extension'
+  ));
+}
+
+module.exports = connectAsMediaExtension;
```
charliesantos, line 23: Can we use ts for this whole file?

lib/connectasmediaextension.js
```diff
@@ -0,0 +1,23 @@
...
+const CancelablePromise = require('./util/cancelablepromise');
+const { typeErrors: E } = require('./util/constants');
+
+/**
+ * Connect as a media extension {@link Participant}.
+ * @param {ConnectOptions} [options]
+ * @returns {CancelablePromise<Room>}
+ */
+function connectAsMediaExtension(options = {}) {
+  if (typeof context === 'object' && typeof context.connectAsMediaExtension === 'function') {
```
charliesantos, line 13: Is this the whole implementation of this video sdk api? What is the benefit of adding this to the sdk if the application already have access to `context.connectAsMediaExtension`?

charliesantos (changes requested):

lib/connectasmediaextension.js
manjeshbhargav: @charliesantos Sure. This is just a POC. I will change this to TS once we begin full implementation.

lib/connectasmediaextension.js
manjeshbhargav: We discussed about who should own this function, and it was agreed that the SDK will own it, because we need some place where we can provide documentation and examples. The internal context-based function will be managed by the back-end team, but it will not be public.

charliesantos:
Ok approved. I thought we're close to prod and not realizing this is all still POC.

charliesantos (approved):

PR #1685, opened 2022-01-27
VIDEO-8413 resend network stats if NQ score is not received 

makarandp0:
**Contributing to Twilio**

### Description

On network quality MSP channel, client sends Network stats, and waits for server to send NQ score, once score is received the client sends next batch of stats. Server always sends response only when stats are received. 
We found that this protocol breaks sometimes with packet loss. With this fix, client will wait 5 seconds for NQ score from server after sending stats, if it does not receive the NQ score, it will send the stats again, with exponentially increasing time.

Spec: https://code.hq.twilio.com/client/room-signaling-protocol/pull/78

### Before review
* [x] TODO: Manually sanity tested running locally
* [x] Ready for review


lib/signaling/v2/networkqualitysignaling.js
```diff
@@ -96,6 +99,13 @@ class NetworkQualitySignaling extends MediaSignaling {
       _networkQualityInputs: {
         value: new AsyncVar()
       },
+      _resendTimer: {
+        value: new Timeout(() => {
+          // and schedule next timer for double the delay..
+          this._resendTimer.setDelay(this._resendTimer.delay * 2);
```
charliesantos, line 105: This seems too steep of a curve. Thoughts on using a lower multiplier? 1.3 maybe?

makarandp0: Okay, I have now proposed that we do it at multiplier of 1.5 (https://code.hq.twilio.com/client/room-signaling-protocol/pull/78) Once that PR is approved I will code with whatever we agree.

charliesantos (approved):

PR #1689, opened 2022-01-31
VIDEO-7728 | Better Preflight Errors

charliesantos:
**Contributing to Twilio**

### Description

See changelog.
I will update the tests once these changes are approved.

### Before review
* [ ] Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,8 @@ executors:
     environment:
       <<: *defaultEnv
   machine-executor:
-    machine: true
+    machine:
+      image: ubuntu-2004:202111-02
     environment:
       <<: *defaultEnv
 commands:
--- a/.circleci/images/docker-compose.yml
+++ b/.circleci/images/docker-compose.yml
@@ -1,11 +1,12 @@
 version: '3'
 # Note: This file is used to build run the integration tests.
 #   for running integration tests
-#   BROWSER=chrome BVER=stable docker-compose --file=./docker-compose.yml run integrationTests
+#   BROWSER=chrome BVER=stable docker-compose --file=.circleci/images/docker-compose.yml run integrationTests
 #   to build container with browsers that will be used by integrationTests:
-#   BROWSER=chrome BVER=stable docker-compose --file=./docker-compose.yml build browserContainer
+#   BROWSER=chrome BVER=stable docker-compose --file=.circleci/images/docker-compose.yml build browserContainer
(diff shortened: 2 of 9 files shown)


lib/preflight/preflighttest.ts
```diff
@@ -110,6 +115,7 @@ export class PreflightTest extends EventEmitter {
   private _connectTiming = new Timer();
   private _sentBytesMovingAverage = new MovingAverageDelta();
   private _packetLossMovingAverage = new MovingAverageDelta();
+  private _progressEvents: string[] = [];
```
makarandp0, line 118: How about changing _progressEvents to be `Map<time, event>` and also reporting it to insights.

lib/preflight/preflighttest.ts
```diff
@@ -347,19 +365,19 @@ export class PreflightTest extends EventEmitter {
         const updatedAnswer = answer;
```
makarandp0, line 365: looks like `updatedAnswer` is not needed

makarandp0 (approved):
@charliesantos, This is looking very good. I have just one suggestion about adding preflight progress events along with their timing to insights.

makarandp0:
also please hold off on merging to develop - we need to decide if we want to release `2.19.1` before this change.

lib/preflight/preflighttest.ts
```diff
@@ -163,13 +169,13 @@ export class PreflightTest extends EventEmitter {
       },
       selectedIceCandidatePairStats: collectedStats ? collectedStats.selectedIceCandidatePairStats : null,
       iceCandidateStats: collectedStats ? collectedStats.iceCandidateStats : [],
+      progressEvents: this._progressEvents,
```
makarandp0, line 172: I think insights does not like arrays. please check if this works

lib/preflight/preflighttest.ts
```diff
@@ -438,6 +456,11 @@ export class PreflightTest extends EventEmitter {
     }
     return collectedStats;
   }
+
+  private _updateProgress(name: string): void {
+    this._progressEvents.push({ timestamp: Date.now(), name });
```
makarandp0, line 461: instead of  `timestamp` we should use timeFromStart as it would be much useful that absolute time.

makarandp0 (commented):
looking good, but have couple more suggestions

makarandp0 (approved):

PR #1717, opened 2022-02-23
VIDEO-8868 Connect to a RSP v3 Room (Large Room)

manjeshbhargav:
**Contributing to Twilio**

### Description

Changes:

1. Set RSP version to 3.
2. Always request track_subscriptions MSP.

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/util/index.js
+++ b/lib/util/index.js
@@ -718,7 +718,9 @@ function createMediaSignalingPayload(dominantSpeaker, networkQuality, trackPrior
     trackSwitchOff
       // eslint-disable-next-line
       ? { track_switch_off: transports }
-      : {}
+      : {},
+    // eslint-disable-next-line
+    { track_subscriptions: transports }
   );
 }
 
--- a/test/unit/spec/signaling/v2/twilioconnectiontransport.js
+++ b/test/unit/spec/signaling/v2/twilioconnectiontransport.js
@@ -139,6 +139,7 @@ describe('TwilioConnectionTransport', () => {
           assert.equal(typeof message.format, 'string');
           assert.equal(message.ice_servers, iceServersStatus);
           assert.equal(message.name, test.name);
+          assert.deepEqual(message.media_signaling.track_subscriptions.transports, [{ type: 'data-channel' }]);
 
           if (networkQuality) {
             assert.deepEqual(message.media_signaling.network_quality.transports, [{ type: 'data-channel' }]);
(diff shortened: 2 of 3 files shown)


cmebarrow (approved):
LGTM. Just one point: will the code stay under lib/signaling/v2 or move under v3? I guess that'll be sorted out later.

manjeshbhargav:
@cmebarrow 
> LGTM. Just one point: will the code stay under lib/signaling/v2 or move under v3? I guess that'll be sorted out later.
Yes. This will happen in future PRs.

lib/util/index.js
```diff
@@ -718,7 +718,9 @@ function createMediaSignalingPayload(dominantSpeaker, networkQuality, trackPrior
     trackSwitchOff
       // eslint-disable-next-line
       ? { track_switch_off: transports }
-      : {}
+      : {},
+    // eslint-disable-next-line
+    { track_subscriptions: transports }
```
ceaglest, line 723: Does the v3 to v2 downgrade imply that trackSwitchOff is no longer optional? The seamless downgrade is one of the reasons I proposed to make bandwidth profiles default / mandatory in the next major version.

In C++ SDK the track_switch_off is only negotiated only when a bandwidth profile is set today. If we make the API change the implementation would take care of itself I guess.

manjeshbhargav: @ceaglest Sure, I agree with making bandwidth profile mandatory (plus removing `disabled` for `trackSwitchOffMode`). However, I want to keep the bandwidth profile related changes for large rooms out of scope for March 1. I'm sure the react app always enables bandwidth profile right now (which will be used for the demo).

ceaglest: Fair enough, as long as we don't expect the final behavior downgrade behavior in the pilot. 👍

test/unit/spec/signaling/v2/twilioconnectiontransport.js
```diff
@@ -182,7 +183,7 @@ describe('TwilioConnectionTransport', () => {
           assert.deepEqual(message.peer_connections, test.peerConnectionManager.getStates());
           assert.equal(message.token, test.accessToken);
           assert.equal(message.type, 'connect');
-          assert.equal(message.version, 2);
+          assert.equal(message.version, 3);
```
ceaglest, line 186: It looks like the Ice request below is for version 1 (line 200). If possible all the messages should reflect version 3. I don't think it's mandatory for this PR though.

ceaglest (approved):
Looks good based upon the requirements of the ticket. My comments can definitely be addressed later.

PR #1722, opened 2022-03-03
VIDEO-8954: Adding in extra links and details to the CHANGELOG entry

PikaJoyce:
JIRA Ticket: [VIDEO-8954](https://issues.corp.twilio.com/browse/VIDEO-8954)

Team, this is a relatively small ticket that simply adds a bit more detail into our CHANGELOG with regards to the WKWebView  release. Mostly some links and some shifting around. 

**Contributing to Twilio**

### Description

### Before review
* [ ] Ready for review


anna-vasilko (changes requested):
Few minor notes:
- Splitting into paragraphs would help, especially for the note on Safari 
- Please add a detail ** track of each kind** to "WKWebViews only support only one local media track at a time" since it's one for audio and one for video
- I'd remove a link to Safari page since it is for really old Safari versions, it does not carry much value today

sarahcstringer:
Looks good, I agree with @anna-vasilko's suggestions.

A few questions/thoughts
> As some of the logic behind the isSupported flag relies on the [User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) string, it should follow the correct format if your application is modifying the default value.
- This wasn't totally clear to me on first reading. Here's a slightly restructured sentence, does this capture what this is about?
"The [`isSupported` flag](https://sdk.twilio.com/js/video/releases/2.20.1/docs/module-twilio-video.html) relies partly on the [User-Agent header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) to determine if twilio-video.js officially supports the user's browser. If your application modifies the default value for the User-Agent heading, the string value should follow the [correct syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax)."

> for [iOS applications](https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_ios), your application will need to include the [camera usage description](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/plist/info/NSCameraUsageDescription), [microphone usage description](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW25) and [inline media playback](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/1614793-allowsinlinemediaplayback) in order for the SDK to work on WKWebView.
- I'm not an iOS developer, so this might just be me not knowing things :) But it's not clear to me where you would need to specify this? Is there a link you could add or a sample of what this would look like?

anna-vasilko (changes requested):
@PikaJoyce Re Known issues, could you mention that backgrounding / interruption issues being not unique to webviews but common to mobile browsers overall and point to our known issues page.

anna-vasilko (approved):

PR #1732, opened 2022-03-12
VIDEO-8668,VIDEO-8670,VIDEO-8671 [Large Rooms] RemoteAudioTrack subscription and switch on/off.

manjeshbhargav:
**Contributing to Twilio**

#### Summary

This pull request contains changes for the following tickets:
- [VIDEO-8668](https://issues.corp.twilio.com/browse/VIDEO-8668) - Switch on/off RemoteAudioTracks using `track_subscriptions` MSP to keep only the N-loudest Tracks on.
- [VIDEO-8670](https://issues.corp.twilio.com/browse/VIDEO-8670) - Add `.switchOffReason` property to RemoteAudioTrack.
- [VIDEO-8671](https://issues.corp.twilio.com/browse/VIDEO-8671) - Manage RemoteAudioTrack subscriptions using `track_subscriptions` MSP instead of the RSP `subscribed` payload.

##### RoomV3

- RemoteTracks' subscription and switch off states are received on the `track_subscriptions` MSP channel, as opposed to the `subscribed` RSP payload and `track_switch_off` MSP channels respectively in RoomV2.
- RemoteTracks are mapped to their respective MediaStreamTracks using the corresponding RTCRtpTransceivers' MIDs, as opposed to the MediaStreamTrack IDs in RoomV2.
- Cached switch off states for not-yet-known RemoteTracks contain both `isSwitchedOff` and `switchOffReason`, as opposed to only `isSwitchedOff` in RoomV2.
- MIDs for not-yet-known RemoteTracks are cached, unlike in RoomV2 where RemoteTracks were always known before their corresponding MediaStreamTrack IDs.
- RoomV3 constructs RemoteParticipantV3 instances, as opposed to RoomV2, which constructs RemoteParticipantV2 instances.

##### RemoteParticipantV3

- RemoteParticipantV3 constructs RemoteTrackPublicationV3 instances, as opposed to RemoteParticipantV2, which constructs RemoteTrackPublicationV2 instances.
- RemoteParticipantV3 makes sure that when it constructs a RemoteTrackPublicationV3 instance, it sets the corresponding MediaStreamTrack if its MID has been cached (MID was received before the RemoteTrack was known).

##### RemoteTrackPublicationV3

- Lack of association with a MediaStreamTrack does not mean that the RemoteTrack is unsubscribed from (it can be switched off). So, a separate boolean property tracks the subscription status.
- Switch off reason is also updated along with switch off state.
- The enabled state is updated based on switch off state and reason, as opposed to updating based on the `enabled` field in `participant.tracks` payload of an RSP `UPDATE` message.

##### RemoteTrackPublication

- Updates the corresponding RemoteTrack's switchOffReason.
- Updates the corresponding RemoteTrack's MediaStreamTrack.

##### RemoteAudioTrack

- Adds a new `.switchOffReason` property.
- When `.isSwitchedOff` remains `true` and the `.switchOffReason` changes, then `switchedOff` is emitted.

##### Room

- Room now accepts either a RoomV3 or a RoomV2 instance in its constructor.

##### createCancelableRoomSignalingPromise

- Depending on the `version` property in the RSP CONNECTED message, it resolves the returned Promise with either a RoomV3 or a RoomV2 instance.

#### Resources

 - [IDL changes](https://code.hq.twilio.com/client/video-sdk-api/pull/96)
 - [Track subscriptions MSP specification](https://code.hq.twilio.com/client/room-signaling-protocol/pull/82)

### Before review
* [x] Ready for review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Version 1.x reached End of Life on September 8th, 2021.** See the changelog entry [here](https://www.twilio.com/changelog/end-of-life-complete-for-unsupported-versions-of-the-programmable-video-sdk). Support for the 1.x version ended on December 4th, 2020.
 
+3.0.0-preview.1
+===============
+
+New Features
+------------
+
+**Large Rooms Pilot**
+
+twilio-video.js now allows you to create and join a [Large Room](TODO_doc_link), which supports a large number of Participants (> 50).
+
+- A RemoteAudioTrack of a Participant that is not active in the Room will now be switched off. (VIDEO-8668)
+- A RemoteAudioTrack will now have an additional property called [`switchOffReason`](TODO_doc_link), which describes
+  the reason for it being switched off. (VIDEO-8670)
+
 2.21.0 (In Progress)
 ==========================
 
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -727,6 +727,37 @@ const TrackSwitchOffMode = {
 };
 
 /**
+ * TrackSwitchOffReason describes why a {@link RemoteTrack} is switched off.
+ * @enum {string}
+ */
+const TrackSwitchOffReason = {
+  /**
+   * The publisher disabled the {@link RemoteTrack}.
+   */
+  'disabled-by-publisher': 'disabled-by-publisher',
+
+  /**
+   * The subscriber requested to stop receiving media for the {@link RemoteTrack}.
+   */
+  'disabled-by-subscriber': 'disabled-by-subscriber',
+
+  /**
+   * The remaining downlink bandwidth is not sufficient to receive media for the {@link RemoteTrack}.
+   */
+  'max-bandwidth-reached': 'max-bandwidth-reached',
+
+  /**
+   * The maximum number of {@link RemoteTrack}s of a {@link Track.Kind} that can be switch on has been reached.
+   */
+  'max-tracks-switched-on': 'max-tracks-switched-on',
+
+  /**
+   * Media for the {@link RemoteTrack} could not be received due to network congestion.
+   */
+  'network-congestion': 'network-congestion'
+};
+
+/**
  * {@link BandwidthProfileMode} specifies how {@link RemoteVideoTrack}s' {@link Track.Priority} values
  * are mapped to bandwidth allocation in Group Rooms.
  * @enum {string}
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -727,32 +727,46 @@ const TrackSwitchOffMode = {
 };
 
 /**
- * TrackSwitchOffReason describes why a {@link RemoteTrack} is switched off.
+ * {@link TrackSwitchOffReason} describes why a {@link RemoteTrack} is switched off. This reason
+ * accompanies the {@link RemoteTrack} event <code>switchedOff</code> and is used for
+ * the {@link RemoteTrack} property <code>switchOffReason</code>.
  * @enum {string}
  */
 const TrackSwitchOffReason = {
   /**
-   * The publisher disabled the {@link RemoteTrack}.
+   * The {@link RemoteTrack} was disabled by the publishing {@link Participant}.
+   * The media server does not send media to the subscribing {@link Participant}s
+   * for a disabled {@link Track}.
    */
   'disabled-by-publisher': 'disabled-by-publisher',
 
   /**
-   * The subscriber requested to stop receiving media for the {@link RemoteTrack}.
+   * The {@link RemoteVideoTrack} was disabled by the subscribing {@link Participant}
+   * and the media server stopped sending its media.
    */
   'disabled-by-subscriber': 'disabled-by-subscriber',
 
   /**
-   * The remaining downlink bandwidth is not sufficient to receive media for the {@link RemoteTrack}.
+   * The {@link RemoteVideoTrack} was switched off because the remaining
+   * downlink bandwidth is not sufficient to receive its media. The bandwidth
+   * limit is configured by specifying <code>maxSubscriptionBitrate</code>
+   * in {@link VideoBandwidthProfileOptions} or a default value is selected
+   * by the media server.
    */
   'max-bandwidth-reached': 'max-bandwidth-reached',
 
   /**
-   * The maximum number of {@link RemoteTrack}s of a {@link Track.Kind} that can be switch on has been reached.
+   * The {@link RemoteTrack} was switched off because the number of switched on
+   * {@link Track}s reached the limit specified by the <code>maxSwitchedOnTracks</code>
+   * property in {@link AudioBandwidthProfile} or {@link VideoBandwidthProfileOptions}.
    */
   'max-tracks-switched-on': 'max-tracks-switched-on',
 
   /**
-   * Media for the {@link RemoteTrack} could not be received due to network congestion.
+   * The {@link RemoteVideoTrack} was switched off because network congestion
+   * was detected or predicted. The <code>trackSwitchOffMode</code> property
+   * on {@link VideoBandwidthProfileOptions} allows you to specify how the
+   * switch off is managed.
    */
   'network-congestion': 'network-congestion'
 };
--- a/lib/media/track/remoteaudiotrack.js
+++ b/lib/media/track/remoteaudiotrack.js
@@ -11,6 +11,7 @@ const RemoteMediaAudioTrack = mixinRemoteMediaTrack(AudioTrack);
  * @extends AudioTrack
  * @property {boolean} isEnabled - Whether the {@link RemoteAudioTrack} is enabled
  * @property {boolean} isSwitchedOff - Whether the {@link RemoteAudioTrack} is switched off
+ * @property {?TrackSwitchOffReason} switchOffReason - The reason for the {@link RemoteMediaTrack} being switched off
  * @property {Track.SID} sid - The {@link RemoteAudioTrack}'s SID
  * @property {?Track.Priority} priority - The subscribe priority of the {@link RemoteAudioTrack}
  * @emits RemoteAudioTrack#disabled
@@ -23,16 +24,17 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
   /**
    * Construct a {@link RemoteAudioTrack}.
    * @param {Track.SID} sid - The {@link RemoteAudioTrack}'s SID
-   * @param {MediaTrackReceiver} mediaTrackReceiver - An audio MediaStreamTrack container
+   * @param {?MediaTrackReceiver} mediaTrackReceiver - An audio MediaStreamTrack container
    * @param {boolean} isEnabled - Whether the {@link RemoteAudioTrack} is enabled
    * @param {boolean} isSwitchedOff - Whether the {@link RemoteAudioTrack} is switched off
+   * @param {?string} switchOffReason - The reason the {@link RemoteAudioTrack} is switched off
    * @param {function(?Track.Priority): void} setPriority - Set or clear the subscribe
    *  {@link Track.Priority} of the {@link RemoteAudioTrack}
    * @param {function(ClientRenderHint): void} setRenderHint - Set render hints.
-   * @param {{log: Log}} options - The {@link RemoteTrack} options
+   * @param {{log: Log, name: string}} options - The {@link RemoteTrack} options
    */
-  constructor(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options) {
-    super(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options);
+  constructor(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options) {
+    super('audio', sid, mediaTrackReceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options);
   }
 
   toString() {
@@ -76,6 +78,8 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
  * A {@link RemoteAudioTrack} was switched off.
  * @param {RemoteAudioTrack} track - The {@link RemoteAudioTrack} that was
  *   switched off
+ * @param {?TrackSwitchOffReason} switchOffReason - The reason the {@link RemoteAudioTrack}
+ *   was switched off
  * @event RemoteAudioTrack#switchedOff
  */
 
--- a/lib/media/track/remoteaudiotrack.js
+++ b/lib/media/track/remoteaudiotrack.js
@@ -11,7 +11,9 @@ const RemoteMediaAudioTrack = mixinRemoteMediaTrack(AudioTrack);
  * @extends AudioTrack
  * @property {boolean} isEnabled - Whether the {@link RemoteAudioTrack} is enabled
  * @property {boolean} isSwitchedOff - Whether the {@link RemoteAudioTrack} is switched off
- * @property {?TrackSwitchOffReason} switchOffReason - The reason for the {@link RemoteMediaTrack} being switched off
+ * @property {?TrackSwitchOffReason} switchOffReason - The reason for the {@link RemoteMediaTrack} being switched off;
+ *   If switched on, it is set to <code>null</code>; The {@link RemoteMediaTrack} is initially switched off with this
+ *   property set to <code>disabled-by-subscriber</code>
  * @property {Track.SID} sid - The {@link RemoteAudioTrack}'s SID
  * @property {?Track.Priority} priority - The subscribe priority of the {@link RemoteAudioTrack}
  * @emits RemoteAudioTrack#disabled
@@ -75,7 +77,11 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
  */
 
 /**
- * A {@link RemoteAudioTrack} was switched off.
+ * A {@link RemoteAudioTrack} was switched off. The media server stops sending media for the
+ * {@link RemoteAudioTrack} until it is switched back on. Just before the event is raised,
+ * <code>isSwitchedOff</code> is set to <code>true</code> and <code>switchOffReason</code> is
+ * set to a {@link TrackSwitchOffReason}. Also, the <code>mediaStreamTrack</code> property is
+ * set to <code>null</code>.
  * @param {RemoteAudioTrack} track - The {@link RemoteAudioTrack} that was
  *   switched off
  * @param {?TrackSwitchOffReason} switchOffReason - The reason the {@link RemoteAudioTrack}
@@ -84,7 +90,11 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
  */
 
 /**
- * A {@link RemoteAudioTrack} was switched on.
+ * A {@link RemoteAudioTrack} was switched on. The media server starts sending media for the
+ * {@link RemoteAudioTrack} until it is switched off. Just before the event is raised,
+ * <code>isSwitchedOff</code> is set to <code>false</code> and <code>switchOffReason</code>
+ * is set to <code>null</code>. Also, the <code>mediaStreamTrack</code> property is set to a
+ * MediaStreamTrack that is the source of the {@link RemoteAudioTrack}'s media.
  * @param {RemoteAudioTrack} track - The {@link RemoteAudioTrack} that was
  *   switched on
  * @event RemoteAudioTrack#switchedOn
--- a/lib/media/track/remotemediatrack.js
+++ b/lib/media/track/remotemediatrack.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { typeErrors: E, trackPriority } = require('../../util/constants');
+const { typeErrors: E, trackPriority, trackSwitchOffReason } = require('../../util/constants');
 const { guessBrowser, isIOSChrome } = require('@twilio/webrtc/lib/util');
 const documentVisibilityMonitor = require('../../util/documentvisibilitymonitor.js');
 
@@ -10,6 +10,7 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
    * {@link Room} by a {@link RemoteParticipant}.
    * @property {boolean} isEnabled - Whether the {@link RemoteMediaTrack} is enabled
    * @property {boolean} isSwitchedOff - Whether the {@link RemoteMediaTrack} is switched off
+   * @property {?TrackSwitchOffReason} switchOffReason - The reason for the {@link RemoteMediaTrack} being switched off
    * @property {Track.SID} sid - The SID assigned to the {@link RemoteMediaTrack}
    * @property {?Track.Priority} priority - The subscribe priority of the {@link RemoteMediaTrack}
    * @emits RemoteMediaTrack#disabled
@@ -20,16 +21,18 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
   return class RemoteMediaTrack extends AudioOrVideoTrack {
     /**
      * Construct a {@link RemoteMediaTrack}.
+     * @param {Track.Kind} kind
      * @param {Track.SID} sid
-     * @param {MediaTrackReceiver} mediaTrackReceiver
+     * @param {?MediaTrackReceiver} mediaTrackReceiver
      * @param {boolean} isEnabled
-      @param {boolean} isSwitchedOff
+     * @param {boolean} isSwitchedOff
+     * @param {?string} switchOffReason
      * @param {function(?Track.Priority): void} setPriority - Set or clear the subscribe
      *  {@link Track.Priority} of the {@link RemoteMediaTrack}
      * @param {function(ClientRenderHint): void} setRenderHint - Set render hints.
-     * @param {{log: Log, name: ?string}} options
+     * @param {{log: Log, name: string}} options
      */
-    constructor(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options) {
+    constructor(kind, sid, mediaTrackReceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options) {
       options = Object.assign({
         // NOTE(mpatwardhan): WebKit bug: 212780 sometimes causes the audio/video elements to stay paused when safari
         // regains foreground. To workaround it, when safari gains foreground - we will play any elements that were
@@ -40,7 +43,7 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
           && typeof document.visibilityState === 'string'
       }, options);
 
-      super(mediaTrackReceiver, options);
+      super(kind, mediaTrackReceiver, options);
 
       Object.defineProperties(this, {
         _isEnabled: {
@@ -64,6 +67,17 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
             setRenderHint(renderHint);
           }
         },
+        _switchOffReason: {
+          value: switchOffReason,
+          writable: true
+        },
+        _workaroundWebKitBug212780: {
+          value: options.workaroundWebKitBug212780
+        },
+        _workaroundWebKitBug212780Cleanup: {
+          value: null,
+          writable: true
+        },
         isEnabled: {
           enumerable: true,
           get() {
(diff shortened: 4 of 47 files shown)


lib/media/track/remotetrackpublication.js
```diff
@@ -94,14 +98,26 @@ class RemoteTrackPublication extends TrackPublication {
...
-        this._log.debug(`${this.trackSid}: ${isSwitchedOff ? 'OFF' : 'ON'} => ${signaling.isSwitchedOff ? 'OFF' : 'ON'}`);
+      const newSwitchOffReason = signaling.switchOffReason || null;
+      if (isSwitchedOff !== signaling.isSwitchedOff || switchOffReason !== newSwitchOffReason) {
+        log.debug(`${this.trackSid}: ${isSwitchedOff ? 'OFF' : 'ON'} => ${signaling.isSwitchedOff ? 'OFF' : 'ON'}`);
+        log.debug(`${this.trackSid} off_reason: ${switchOffReason} => ${newSwitchOffReason}`);
         isSwitchedOff = signaling.isSwitchedOff;
+        switchOffReason = newSwitchOffReason;
+        if (this.track) {
+          this.track._setSwitchedOff(signaling.isSwitchedOff, switchOffReason);
+          this.emit(`trackSwitched${isSwitchedOff ? 'Off' : 'On'}`,  this.track, ...(isSwitchedOff ? [this.track.switchOffReason] : []));
```
makarandp0, line 109: NIT. breaking strings make it harder to search for code references. 
```suggestion
          this.emit(isSwitchedOff ? 'trackSwitchedOff' : 'trackSwitchedOn',  this.track, ...(isSwitchedOff ? [this.track.switchOffReason] : []));
```

lib/signaling/v2/remoteparticipant.js
```diff
@@ -21,16 +21,17 @@ class RemoteParticipantV2 extends RemoteParticipantSignaling {
     super(participantState.sid, participantState.identity);
 
     options = Object.assign({
-      RemoteTrackPublicationV2
+      RemoteTrackPublicationSignaling: RemoteTrackPublicationV2,
+      otherProperties: {}
```
makarandp0, line 25: curious - why `otherProperties` are not folded into options?

lib/signaling/v2/remoteparticipant.js
```diff
@@ -52,14 +53,18 @@ class RemoteParticipantV2 extends RemoteParticipantSignaling {
       }
     });
 
+    Object.entries(options.otherProperties).forEach(([key, value]) => {
+      Object.defineProperty(this, key, { value });
```
makarandp0, line 57: I am not really understanding why we want some runtime properties on this type. If they are really known at runtime only, Could we keep them in a single Map property instead.
something like 
```
  otherProperties:  options.otherProperties
});
```

I am afraid generating them as object properties would make it hard to convert to typescript.

lib/signaling/v2/room.js
```diff
@@ -142,6 +145,13 @@ class RoomV2 extends RoomSignaling {
     this._initNetworkQualityMonitorSignaling();
     this._initPublisherHintSignaling();
 
+    this._otherMediaSignalings.forEach(({ MediaSignaling, name, initMethod }) => {
```
makarandp0, line 148: same comment here. I prefer not making these runtime properties, and instead keep them in separate map.

lib/signaling/v2/room.js
```diff
@@ -190,14 +200,14 @@ class RoomV2 extends RoomSignaling {
   /**
    * @private
    */
-  _getOrCreateTrackReceiverDeferred(id) {
+  _getOrCreateTrackReceiverDeferred(id, idOrMid = 'id') {
```
makarandp0, line 203: NIT: now that id could be mid, could we rename this variable to propertyValue 

```suggestion
  _getOrCreateTrackReceiverDeferred(propertyValue, propertyName = 'id') {
```

lib/signaling/v2/room.js
```diff
@@ -190,14 +200,14 @@ class RoomV2 extends RoomSignaling {
...
     const deferred = this._trackReceiverDeferreds.get(id) || defer();
```
makarandp0, line 204: is id going to be unique even when its `mid`? I suggest prepend it when maintaining those.
```suggestion
    const deferred = this._trackReceiverDeferreds.get(propertyName + ":" + propertyValue) || defer();
```

lib/signaling/v2/room.js
```diff
@@ -220,6 +230,20 @@ class RoomV2 extends RoomSignaling {
     return this;
   }
 
+  /**
+   * @private
+   */
+  _createRemoteParticipant(participantState) {
+    const { _RemoteParticipantSignaling: RemoteParticipantV2 } = this;
```
makarandp0, line 237: _RemoteParticipantSignaling can now be RemoteParticipantV3 or RemoteParticipantV2 right? should this be:
```suggestion
    const { _RemoteParticipantSignaling: RemoteParticipantSignalingClassType } = this;
```

lib/signaling/v2/room.js
```diff
@@ -241,8 +265,8 @@ class RoomV2 extends RoomSignaling {
   /**
    * @private
    */
-  _getTrackReceiver(id) {
-    return this._getOrCreateTrackReceiverDeferred(id).promise.then(trackReceiver => {
+  _getTrackReceiver(id, idOrMid = 'id') {
```
makarandp0, line 268: similar to earlier suggestion
```suggestion
  _getTrackReceiver(propertyValue, propertyName = 'id') {
```

lib/signaling/v3/remoteparticipant.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const RemoteParticipantV2 = require('../v2/remoteparticipant');
```
makarandp0, line 3: Was there something blocking new code to be written in typescript? was it because of the existing `js` requires?

lib/signaling/v3/remoteparticipant.js
```diff
@@ -0,0 +1,66 @@
...
+    getPendingTrackReceiver,
+    getInitialTrackSwitchOffState,
+    setPriority,
+    setRenderHint,
+    clearTrackHint,
+    options
+  ) {
+    options = Object.assign({
+      RemoteTrackPublicationSignaling: RemoteTrackPublicationV3,
+      otherProperties: {
```
makarandp0, line 31: curious why otherProperties is a separate object.

lib/signaling/v3/remoteparticipant.js
```diff
@@ -0,0 +1,66 @@
...
+      options
+    );
+  }
+
+  /**
+   * @private
+   */
+  _getOrCreateTrack(trackState) {
+    const {
+      _RemoteTrackPublicationSignaling: RemoteTrackPublicationV3,
```
makarandp0, line 51: maybe it's worth keeping both `RemoteTrackPublicationV3` and `RemoteTrackPublicationV2` in options object (instead of conditional _RemoteTrackPublicationSignaling), so that we do not need to rename it here.

lib/signaling/v3/room.js
```diff
@@ -0,0 +1,164 @@
...
+  _createRemoteParticipant(participantState) {
+    const { _RemoteParticipantSignaling: RemoteParticipantV3 } = this;
+    return new RemoteParticipantV3(
+      participantState,
+      trackSid => this._getPendingTrackReceiver(trackSid),
+      trackSid => this._getInitialTrackSwitchOffState(trackSid),
+      (trackSid, priority) => this._trackPrioritySignaling.sendTrackPriorityUpdate(trackSid, 'subscribe', priority),
+      (trackSid, hint) => this._renderHintsSignaling.setTrackHint(trackSid, hint),
+      trackSid => this._renderHintsSignaling.clearTrackHint(trackSid)
+    );
```
makarandp0, line 69: this parameter list is getting too long, and making this caller code to hard to read. How about we pass these params as an object. This will make it more readable for unit tests as well.
```suggestion
    return new RemoteParticipantV3({
      participantState: participantState,
      getReceiver: trackSid => this._getPendingTrackReceiver(trackSid),
      getSwitchOfState: trackSid => this._getInitialTrackSwitchOffState(trackSid),
      setPriority: (trackSid, priority) => this._trackPrioritySignaling.sendTrackPriorityUpdate(trackSid, 'subscribe', priority),
      setRenderHint: (trackSid, hint) => this._renderHintsSignaling.setTrackHint(trackSid, hint),
      clearRenderHint: trackSid => this._renderHintsSignaling.clearTrackHint(trackSid)
    });
```

makarandp0 (changes requested):
It's looking good, I made one pass, and have few suggestion. I am wondering is there anything we can do to be able to unblock us from writing new code in typescript.

lib/signaling/v2/room.js
makarandp0: never mind this. I understand that it will always be `RemoteParticipantV2` as its v2 version. Ignore my suggestions Looks good, its easier to search this way.

mmartikainen (commented):
Just a few comments/suggestions/questions.

lib/connect.js
```diff
@@ -726,6 +726,37 @@ const TrackSwitchOffMode = {
...
+ * @enum {string}
+ */
+const TrackSwitchOffReason = {
+  /**
+   * The publisher disabled the {@link RemoteTrack}.
+   */
+  'disabled-by-publisher': 'disabled-by-publisher',
+
+  /**
+   * The subscriber requested to stop receiving media for the {@link RemoteTrack}.
```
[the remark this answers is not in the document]
manjeshbhargav: I think "subscriber" is fine to use here, as we already have public API events (`trackSubscribed` and `trackSubscriptionFailed`) and switch off reason (`disabled-by-subscriber`) that mention subscriptions.

dipankadas (changes requested):

lib/media/track/remoteaudiotrack.js
```diff
@@ -23,16 +24,17 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
...
+   * @param {?string} switchOffReason - The reason the {@link RemoteAudioTrack} is switched off
    * @param {function(?Track.Priority): void} setPriority - Set or clear the subscribe
    *  {@link Track.Priority} of the {@link RemoteAudioTrack}
    * @param {function(ClientRenderHint): void} setRenderHint - Set render hints.
-   * @param {{log: Log}} options - The {@link RemoteTrack} options
+   * @param {{log: Log, name: string}} options - The {@link RemoteTrack} options
    */
-  constructor(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options) {
-    super(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options);
+  constructor(sid, mediaTrackReceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options) {
```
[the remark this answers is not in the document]
manjeshbhargav: Currently, we specify required parameters as function arguments and optional parameters as properties of the options object. I think we should keep this approach.

lib/signaling/v2/room.js
```diff
@@ -122,6 +122,9 @@ class RoomV2 extends RoomSignaling {
       _trackSwitchOffSignaling: {
         value: new options.TrackSwitchOffSignaling(getTrackReceiver, { log }),
       },
+      _otherMediaSignalings: {
```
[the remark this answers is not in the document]
manjeshbhargav: We have used this nomenclature since the beginning, for example TrackSignaling, PeerConnectionSignaling etc. So, I would like to retain the naming convention.

lib/signaling/v2/room.js
manjeshbhargav: I think the current argument names are more descriptive.

lib/signaling/v2/room.js
manjeshbhargav: MIDs will be unique.

lib/signaling/v2/room.js
manjeshbhargav: See my earlier [response](https://github.com/twilio/twilio-video.js/pull/1732/files#r828499850).

lib/signaling/v3/remoteparticipant.js
manjeshbhargav: I want the typescript conversion to happen top-down, with public classes being converted first. I don't want to be inn a situation where some public classes are in ts, some are in js, some internal classes are in ts, some are in js.

lib/signaling/v3/remoteparticipant.js
manjeshbhargav: This should be resolved in the next commit.

lib/signaling/v3/remoteparticipant.js
manjeshbhargav: I don't think we should do that. In principle, if there is not much difference between v2 and v3, then at the base class level, a more generic name is much more descriptive of what is going on.

lib/signaling/v3/remotetrackpublication.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * Set the {@link MediaTrackReceiver} on the {@link RemoteTrackPublicationV3}.
+   * @override
+   * @param {MediaTrackReceiver} trackReceiver
+   * @param {boolean} isSubscribed
+   * @returns {this}
+   */
+  setTrackTransceiver(trackReceiver, isSubscribed) {
```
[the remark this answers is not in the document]
manjeshbhargav: We are overriding a base class function `setTrackTransceiver` in a derived class which deals with receivers only. So I think this is fine.

lib/signaling/v3/remotetrackpublication.js
```diff
@@ -0,0 +1,126 @@
...
+    return this;
+  }
+
+  /**
+   * Compare the {@link RemoteTrackPublicationV3} to a
+   * {@link RemoteTrackPublicationV2#Representation} of itself and perform any
+   * updates necessary.
+   * @param {RemoteTrackPublicationV3#Representation} track
+   * @returns {this}
+   * @fires TrackSignaling#updated
```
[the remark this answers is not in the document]
manjeshbhargav: It is still correct, although the only track property updated through RSP is the priority, since enabled is removed.

lib/signaling/v3/remotetrackpublication.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * The Room Signaling Protocol (RSP) representation of a {@link RemoteTrackPublicationV3}.
+ * @typedef {object} RemoteTrackPublicationV3#Representation
+ * @property {Track.Kind} kind
+ * @property {string} name
+ * @priority {Track.Priority} priority
+ * @property {Track.SID} sid
+ */
```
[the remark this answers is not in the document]
manjeshbhargav: What am I missing exactly here?

lib/signaling/v3/room.js
manjeshbhargav: See my previous [response](https://github.com/twilio/twilio-video.js/pull/1732/files#r828481248).

lib/signaling/v3/room.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * @private
+   * @override
+   */
+  _getInitialTrackSwitchOffState(trackSid) {
+    const switchOffState = this._pendingSwitchOffStates.get(trackSid)
+      || { state: 'ON', switchOffReason: null };
```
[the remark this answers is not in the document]
manjeshbhargav: Got it. Let me change it to:
```suggestion
    const switchOffState = this._pendingSwitchOffStates.get(trackSid)
      || { state: 'OFF', switchOffReason: 'DISABLED_BY_SUBSCRIBER' };
```

lib/media/track/remoteaudiotrack.js
```diff
@@ -23,16 +24,17 @@ class RemoteAudioTrack extends RemoteMediaAudioTrack {
   /**
    * Construct a {@link RemoteAudioTrack}.
    * @param {Track.SID} sid - The {@link RemoteAudioTrack}'s SID
-   * @param {MediaTrackReceiver} mediaTrackReceiver - An audio MediaStreamTrack container
+   * @param {?MediaTrackReceiver} mediaTrackReceiver - An audio MediaStreamTrack container
    * @param {boolean} isEnabled - Whether the {@link RemoteAudioTrack} is enabled
    * @param {boolean} isSwitchedOff - Whether the {@link RemoteAudioTrack} is switched off
+   * @param {?string} switchOffReason - The reason the {@link RemoteAudioTrack} is switched off
```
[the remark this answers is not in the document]
manjeshbhargav: isEnabled will be deprecated in a future PR.

lib/connect.js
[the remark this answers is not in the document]
mmartikainen: Do we refer to the SDK user/local participant as 'the subscriber' elsewhere in the doc? (In this context that term feels like it is referring to some other party, not "me".)

lib/signaling/v3/remotetrackpublication.js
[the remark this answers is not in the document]
mmartikainen: OK, It's your code :-)

lib/signaling/v3/remotetrackpublication.js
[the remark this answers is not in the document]
mmartikainen: The doc comment is not attached to anything, i.e. there is nothing after it.

lib/connect.js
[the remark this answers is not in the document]
manjeshbhargav: We already use it while documenting bandwidth profile options like `maxSubscriptionBitrate` and subscriber `priority`.

lib/signaling/v3/remotetrackpublication.js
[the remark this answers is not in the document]
manjeshbhargav: In JSDoc, this is the way to document JSON objects that has specific properties. It is like an object typedef.

makarandp0 (approved):

mmartikainen (approved):

dipankadas (approved):

lib/signaling/v2/room.js
```diff
@@ -220,6 +229,20 @@ class RoomV2 extends RoomSignaling {
...
   }
 
+  /**
+   * @private
+   */
+  _createRemoteParticipant(participantState) {
+    const { _RemoteParticipantSignaling: RemoteParticipantV2 } = this;
+    return new RemoteParticipantV2(
+      participantState,
+      trackSid => this._getInitialTrackSwitchOffState(trackSid),
```
ceaglest, line 239: Tracks are initially on in v2. If they've since been switched off when the RemoteParticipant is created (MSP beats RSP) will this logic suppress the initial off event? I don't think its desirable to do this, but I think it was pre-existing behavior.

lib/signaling/v3/remoteparticipant.js
ceaglest: This is definitely a discussion for the JS team to have, but I think the time to have it is before tickets are cut and code is written. I think we missed the boat to transition internal classes to .ts files this time.

lib/signaling/v3/room.js
```diff
@@ -0,0 +1,158 @@
...
+    });
+  }
+
+  /**
+   * @private
+   * @override
+   */
+  _updateSubscribed(roomState) {
+    /* Do nothing since RSP v3 messages will not contain the "subscribed" property. */
+  }
```
ceaglest, line 155: Maybe subclassing will prevent you from adopting future RSP v3 changes due to dependency on RSP v2. Just a thought anyways..

lib/signaling/v3/room.js
```diff
@@ -0,0 +1,158 @@
...
+          // NOTE(mmalavalli): If a RemoteTrackPublicationV3's MID changes, then we need to unsubscribe
+          // from the RemoteTrack before subscribing to it again with the MediaTrackReceiver associated with the new
+          // MID. If a RemoteTrackPublicationV3's RemoteTrack is switched off, then we should still be subscribed
+          // to it, even though it no longer has an MID associated with it.
+          trackSignaling.setTrackTransceiver(null, isSwitchedOff);
+        }
+        if (!isSwitchedOff) {
+          this._getTrackReceiver(mid, 'mid').then(trackReceiver => trackSignaling.setTrackTransceiver(trackReceiver, true));
+        }
+        trackSignaling.setSwitchedOff(isSwitchedOff);
```
ceaglest, line 128: I think you are coupling the `state` and transceiver management together slightly by disallowing an `'OFF'` state with an assigned MID. I know that the server should not do this, and the protocol spec says it should not be done but I don't know if the client should also enforce the same restriction at the room level without even raising warnings to the developer.

lib/signaling/v3/tracksubscriptionssignaling.js
```diff
@@ -0,0 +1,53 @@
...
+class TrackSubscriptionsSignaling extends MediaSignaling {
+  /**
+   * Construct a {@link TrackSubscriptionsSignaling}.
+   */
+  constructor(getReceiver, options) {
+    super(getReceiver, 'track_subscriptions', options);
+
+    Object.defineProperties(this, {
+      _currentRevision: {
+        value: 0,
```
ceaglest, line 14: Can the initial value be unset? Using zero is ambiguous since the schema specifies the type as `uint32` and doesn't prevent the server from selecting `0` as the first revision.

lib/signaling/v3/tracksubscriptionssignaling.js
```diff
@@ -0,0 +1,53 @@
...
+   */
+  _handleIncomingMessage(message) {
+    const { _log: log, _currentRevision: currentRevision } = this;
+    const { errors = {}, revision, subscribed = {} } = message;
+
+    if (currentRevision >= revision) {
+      log.warn(`Ignoring incoming ${this.channel} message as ${currentRevision} (current revision) >= ${revision} (incoming revision)`);
+      log.debug(`Ignored incoming ${this.channel} message:`, message);
+      return;
+    }
```
ceaglest, line 46: Check for an unset value in order to unambiguously handle the `revision=0` case.

ceaglest (commented):
This is an extensive PR with awesome work! I did my best to review it, but had to skip the testing code due to the time commitment required and my lack of relevant language experience.

Regarding the API documentation, I thought we had general agreement in the video-sdk-api repo about the content, and the JavaScript docs in this PR would match it (except platform differences, like links to classes). Are the docs in this PR considered temporary, to be updated later?

I definitely do not want to debate API documentation in a 4000+ line implementation PR, as it is not the right place. I didn't merge PR 96 due to an open issue regarding `maxSwitchedOnTracks`, but otherwise consider it complete based upon feedback from the group.

Regarding the implementation, it looks like Track Subscriptions is partially supported in v2 Rooms. This wasn't expected based upon discussions of RSP v3 with the team. If its a temporary requirement for testing in dev can we make lifting that requirement a blocker in order to merge this open PR?

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 
 **Version 1.x reached End of Life on September 8th, 2021.** See the changelog entry [here](https://www.twilio.com/changelog/end-of-life-complete-for-unsupported-versions-of-the-programmable-video-sdk). Support for the 1.x version ended on December 4th, 2020.
 
+3.0.0-preview.1
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest The earlier 2.x beta strings were a mistake. According to `semver`, there should be a `.` between `preview` and `1`.

lib/connect.js
```diff
@@ -726,6 +726,37 @@ const TrackSwitchOffMode = {
   disabled: 'disabled'
 };
 
+/**
+ * TrackSwitchOffReason describes why a {@link RemoteTrack} is switched off.
+ * @enum {string}
+ */
+const TrackSwitchOffReason = {
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest Thanks for pointing it out. It was an oversight on my part that I forgot to expand the description for each enum.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
...
+===============
+
+New Features
+------------
+
+**Large Rooms Pilot**
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest As far as I understand it, a pilot is a feature-related term and preview is an SDK related term. So, in the first 3.x preview release, we are introducing a large rooms feature on a pilot basis.

lib/media/track/remotemediatrack.js
```diff
@@ -128,20 +141,34 @@ function mixinRemoteMediaTrack(AudioOrVideoTrack) {
...
+    }
+
     /**
      * @private
      * @param {boolean} isSwitchedOff
+     * @param {?string} [switchOffReason=null]
      */
-    _setSwitchedOff(isSwitchedOff) {
-      if (this._isSwitchedOff !== isSwitchedOff) {
+    _setSwitchedOff(isSwitchedOff, switchOffReason = null) {
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest The assumption is the SFU will always send a `switchOffReason`. We could enforce this in our integration tests, where we can fail the test if the track is switched off but the reason is null. WDYT?

lib/media/track/remotetrackpublication.js
```diff
@@ -173,6 +189,8 @@ class RemoteTrackPublication extends TrackPublication {
 /**
  * The {@link RemoteTrack} was switched off.
  * @param {RemoteTrack} track - the {@link RemoteTrack} that was switched off
+ * @param {?TrackSwitchOffReason} switchOffReason - the reason the {@link RemoteTrack}
+ *   was switched off
  * @event RemoteTrackPublication#trackSwitchedOff
  */
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest I will update the documentation to match the IDL. For deprecations, I have a separate ticket, which may or may not be done for the pilot (depending on the time remaining).

lib/participant.js
```diff
@@ -346,7 +346,7 @@ class Participant extends EventEmitter {
       };
       const track = kind === 'data'
         ? new RemoteTrack(sid, trackTransceiver, options)
-        : new RemoteTrack(sid, trackTransceiver, isEnabled, isSwitchedOff, setPriority, setRenderHint, options);
+        : new RemoteTrack(sid, trackTransceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options);
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest Right now, DataTracks cannot be switched off. `isSwitchedOff` is always `false` and `switchOffReason` is always `null`.

lib/signaling/v2/room.js
```diff
@@ -119,6 +119,11 @@ class RoomV2 extends RoomSignaling {
       _trackPrioritySignaling: {
         value: new options.TrackPrioritySignaling(getTrackReceiver, { log }),
       },
+      _trackSubscriptionsSignaling: {
+        value: options.TrackSubscriptionsSignaling
+          ? new options.TrackSubscriptionsSignaling(getTrackReceiver, { log })
+          : null
+      },
```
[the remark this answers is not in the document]
manjeshbhargav: @ceaglest It will be instantiated only for RoomV3, because this code only runs if RoomV2 is acting as a base class for RoomV3. Otherwise, this property is `null`.

lib/signaling/v2/room.js
manjeshbhargav: @ceaglest I don't think this will suppress the initial switch off event, since the SFU will not turn it off until the SDK detects that there are no attached visible video elements in the subscription event handler, and sends an off request.
The only time what you are saying may happen is if the SFU switches off the track due to congestion just before the track is created.

lib/signaling/v3/room.js
manjeshbhargav: @ceaglest Yes, I agree. Most of the code shared between RoomV3 and RoomV2 should actually be in the RoomSignaling base class. However, it is a very complex undertaking to move all that code from RoomV2 to RoomSignaling, so I will do so when it comes time to get rid of RoomV2 altogether.

lib/signaling/v3/room.js
manjeshbhargav: @ceaglest I think the code is written with the server behavior as described in the blueprint /  MSP spec. What you are saying is only useful if the SFU introduces a bug into prof where an MID is assigned to a switched off track. In my view the SFU code should have a test where it makes sure that this doesn't happen.

PR #1735, opened 2022-03-15
noise cancellation using rnnoise

makarandp0:
implements API described in this [idl](https://code.hq.twilio.com/client/video-sdk-api/pull/98/files) PR
**Contributing to Twilio**

### Description

### Before review
* [ ] Ready for review


lib/createlocaltracks.ts
```diff
@@ -149,6 +150,12 @@ export async function createLocalTracks(options?: CreateLocalTracksOptions): Pro
     extraLocalTrackOptions.audio.workaroundWebKitBug1208516 = fullOptions.audio.workaroundWebKitBug1208516;
   }
 
+  let noiseCancellationOptions: NoiseCancellationOptions|undefined;
+  if (typeof fullOptions.audio === 'object' && 'noiseCancellationOptions' in fullOptions.audio) {
+    noiseCancellationOptions = fullOptions.audio.noiseCancellationOptions;
+    delete fullOptions.audio.noiseCancellationOptions;
+  }
+
```
charliesantos, line 158: We just kind of adapted this pattern in this file when we were converting it to TS. But now that we're building on top of it, do you think there's a better way to pass around options?

public/rnnoise/rnnoise_sdk.mjs
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
charliesantos, line 1: All these new files under public folder are krisp/rnnoise sdk files correct? If so, I think we need to put them in an assets folder, organized by sdk name and version each file to allow cache-busting on the consumer side. We should also add a doc (before release) to recommend copying the files over and serve them in their web app. It can be done using post install scripts.
For folder structure, maybe something like
```
lib/assets/<sdk name>/<filename>-<version>.<extension>
```

tsdef/types.d.ts
```diff
@@ -164,8 +164,17 @@ export interface CreateLocalTrackOptions extends MediaTrackConstraints {
   workaroundWebKitBug1208516?: boolean;
 }
 
+export interface NoiseCancellationOptions {
```
charliesantos, line 167: We should export this in index.d.ts

tsdef/types.d.ts
```diff
@@ -164,8 +164,17 @@ export interface CreateLocalTrackOptions extends MediaTrackConstraints {
...
+  sdkAssetsPath: string;
+  sdkFile: string;
+}
+
+export interface LocalAudioTrackOptions extends CreateLocalTrackOptions {
```
charliesantos, line 172: Same here, we should export this in index.d.ts

charliesantos (changes requested):
I did a quick pass and left some comments Mak.

manjeshbhargav (changes requested):
@makarandp0 Can you add a detailed PR description with links to the blueprint and IDL changes? If there is no IDL PR, I suggest that we should create and review that before the implementation.

charliesantos (commented):
I like the solution in general since we really don't have another alternative for dynamic imports. And I'm not sure if using script tags is better. Having said that, I have a concern. Does it make sense for dynamicImport.js to live under lib folder? Will this get confusing if we start adding more files that are not compiled using ts?

lib/noisecancellationadapter.ts
```diff
@@ -38,9 +38,16 @@ class NoiseCancellationAdapter  {
...
-      const module = await import(/* webpackIgnore: true */ `${sdkRootPath}/${sdkFile}`);
-      this._log.debug('Loaded noise cancellation sdk:', module);
-      const sdkAPI = module.default as NoiseCancellationSDK;
+
+      // NOTE(mpatwardhan): Typescript has an issue - when module set to "commonjs"
+      // typescript converts dynamic import calls to require :(
+      // https://github.com/microsoft/TypeScript/issues/43329
+      // const dynamicModule = await import(/* webpackIgnore: true */ `${sdkRootPath}/${sdkFile}`);
+      // to workaround this issue we move import call into seperate modue (dynamicImport) that is not
+      // compiled with typescript.
```
charliesantos, line 47: I don't think we need to add this comment block here. Can we just keep it inside the dynamicImport  module?

lib/createlocaltracks.ts
makarandp0: yes, I followed the same pattern, but yeah, I will look into cleaning up this and existing code around this.

lib/preflight/preflighttest.ts
```diff
@@ -9,7 +9,7 @@ import { getTurnCredentials } from './getturncredentials';
 import { makeStat } from './makestat';
 import { syntheticAudio } from './syntheticaudio';
 import { syntheticVideo } from './syntheticvideo';
-import { waitForSometime } from '../util';
+import { waitForSometime } from '../util/index';
```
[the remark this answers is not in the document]
makarandp0: I had to do this when I changed to `module="esnext"` but now that we are keeping commonjs, this change is not necessary.

public/rnnoise/rnnoise_sdk.mjs
makarandp0: Thanks @charliesantos, I will address this before release.

tsdef/types.d.ts
```diff
@@ -164,8 +164,17 @@ export interface CreateLocalTrackOptions extends MediaTrackConstraints {
   workaroundWebKitBug1208516?: boolean;
 }
 
+export interface NoiseCancellationOptions {
+  sdkAssetsPath: string;
+  sdkFile: string;
```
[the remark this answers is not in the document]
makarandp0: yes, `sdkFile` is included under `sdkAssetsPath`. I will remove this before merge.

tsdef/types.d.ts
makarandp0: can you clarify, it is already exported from `index.d.ts`

tsdef/types.d.ts
makarandp0: can you clarify -  I am not sure what need to change.

tsdef/types.d.ts
charliesantos: I should be able to do this
```ts
import { NoiseCancellationOptions } from 'twilio-video';

const options: NoiseCancellationOptions = {foo};
```

But in order to do this, we need to export NoiseCancellationOptions from tsdef/index.d.ts. Currently, I don't see it there.

tsdef/types.d.ts
charliesantos: I should be able to do this
```ts
import { LocalAudioTrackOptions } from 'twilio-video';

const options: LocalAudioTrackOptions = {foo};
```
But in order to do this, we need to export LocalAudioTrackOptions from tsdef/index.d.ts. Currently, I don't see it there.
By the way, were already doing the same pattern for `CreateLocalTrackOptions`. See this [line](https://github.com/twilio/twilio-video.js/blob/audio_processor/tsdef/index.d.ts#L54).

tsdef/types.d.ts
makarandp0: got it! I will update, thanks, I do have a test app, that uses these, but it did not initialize these types specifically, so I missed this part.

tsdef/types.d.ts
makarandp0: got it! I will update, thanks, I do have a test app, that uses these, but it did not initialize these types specifically, so I missed this part.

public/rnnoise/rnnoise_sdk.mjs
makarandp0: now moved the asset files to `assets/<sdk-name>/<version>/files`

lib/createlocaltracks.ts
makarandp0: This PR is already getting pretty big, and this change is following existing pattern. I need more thinking on changing the patter, but I will leave that for future PRs,

makarandp0:
reopened as https://github.com/twilio/twilio-video.js/pull/1766

Issue #455, opened 2018-10-31
“global is not defined” when attempting to consume in Angular/Webpack (using Angular CLI)


poke:
I am getting a *“global is not defined”* reference error when attempting to integrate the project into an Angular application. The error is caused by the use of `global` in 
[`insightspublisher/index.js`](https://github.com/twilio/twilio-video.js/blob/1.14.1/lib/util/insightspublisher/index.js#L11) but I would expect the same error to also occur from the [`sip.js`](https://github.com/twilio/twilio-video.js/blob/1.14.1/lib/sip.js#L3) file.

The way the sources expect a global `global` makes this difficult to use the library in certain build environments that do not provide this as part of the bundling process. Webpack in particular does not do this. Instead, Webpack puts the compiled modules into a closure that looks approximately like this:

```js
function (module, exports, __webpack_require__) {
   // … module content
}
```

As you can see, there’s no `global` being provided.

### Steps to reproduce:

1. `ng new` with a current Angular CLI (7.0.3)
2. `npm i --save twilio-video` to install the dependency
3. Add the following to `app.component.ts`:
    ```ts
    import * as twilio from 'twilio-video';
    console.log(twilio);
    ```
4. `ng serve`
5. Open in browser, and observe the error in the console

### Workaround

A possible workaround would be to import the bundle instead from `twilio-video/dist/twilio-video`. This changes the import name though, prevents [`@types/twilio-video`](http://npmjs.com/package/@types/twilio-video) from matching, and is generally less efficient as this will now bundle a bundle.

poke:
Hey @manjeshbhargav, did you have a chance to look into this problem yet?

manjeshbhargav:
Hi @poke ,

Sorry for the delay. According to [this comment](https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814), they have stopped polyfilling Node "globals" in Angular. However, you can do it yourself by adding this line at the end of your `src/polyfills.ts`:

```typescript
(window as any).global = window;
```

Please let me know if this works for you.

Thanks,

Manjesh Malavalli
JSDK Team

poke:
Yeah, that workaround works for me. Since originally posting the issue, I also have been able to work around this with a TypeScript path mapping to the `dist` version.

Are there any plans to fix this within the library though? There is no `globals` in a web environment, so it doesn’t really make much sense to me to rely on that within this library.

manjeshbhargav:
@poke ,

[...]

PR #1752, opened 2022-04-04
VIDEO-9282 - Migrate SDK from using node.js util.

PikaJoyce:
**Contributing to Twilio**

## Pull Request Details
In this PR, we are redefining the way we grab the global context.
`global.window` to `window`

as well as 

migrating the `util.inherits` method into the SDK

### Description
This PR partially addresses [VIDEO-9282](https://issues.corp.twilio.com/browse/VIDEO-9282). 

We have migrated the SDK dependencies off of the node.js `util` module as well as fixing the issue mentioned on [this issue](https://github.com/twilio/twilio-video.js/issues/455).

However, there is a larger effort and a ticket in place to address our dependencies that rely on `util` still which will be addressed in the future.

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/vendor/inherits.js
+++ b/lib/vendor/inherits.js
@@ -1,0 +1,24 @@
+module.exports = function inherits(ctor, superCtor) {
+  if (ctor && superCtor) {
+    ctor.super_ = superCtor;
+    if (typeof Object.create === 'function') {
+      // implementation from standard node.js 'util' module
+      ctor.prototype = Object.create(superCtor.prototype, {
+        constructor: {
+          value: ctor,
+          enumerable: false,
+          writable: true,
+          configurable: true
+        }
+      });
+    } else {
+      // old school shim for old browsers
+      class TempCtor {
+        constructor() { }
+      }
+      TempCtor.prototype = superCtor.prototype;
+      ctor.prototype = new TempCtor();
+      ctor.prototype.constructor = ctor;
+    }
+  }
+};
--- a/lib/media/track/es5/localaudiotrack.js
+++ b/lib/media/track/es5/localaudiotrack.js
@@ -3,8 +3,7 @@
 // only in place so that we can support ES6 classes without requiring `new`.
 'use strict';
 
-const { inherits } = require('util');
-
+const inherits = require('../../../vendor/inherits');
 const LocalAudioTrackClass = require('../localaudiotrack');
 
 function LocalAudioTrack(mediaStreamTrack, options) {
--- a/lib/media/track/es5/localdatatrack.js
+++ b/lib/media/track/es5/localdatatrack.js
@@ -3,7 +3,7 @@
 // only in place so that we can support ES6 classes without requiring `new`.
 'use strict';
 
-const { inherits } = require('util');
+const inherits = require('../../../vendor/inherits');
 
 const LocalDataTrackClass = require('../localdatatrack');
 
--- a/lib/media/track/es5/localvideotrack.js
+++ b/lib/media/track/es5/localvideotrack.js
@@ -3,7 +3,7 @@
 // only in place so that we can support ES6 classes without requiring `new`.
 'use strict';
 
-const { inherits } = require('util');
+const inherits = require('../../../vendor/inherits');
 
 const LocalVideoTrackClass = require('../localvideotrack');
 
--- a/lib/twilioconnection.js
+++ b/lib/twilioconnection.js
@@ -61,7 +61,7 @@ const WS_CLOSE_BUSY_WAIT = 3005;
 const WS_CLOSE_SERVER_BUSY = 3006;
 const WS_CLOSE_OPEN_TIMEOUT = 3007;
 
-const toplevel = global.window || global;
+const toplevel = window || global;
 const WebSocket = toplevel.WebSocket ? toplevel.WebSocket : require('ws');
 
 const CloseReason = {
--- a/lib/util/insightspublisher/index.js
+++ b/lib/util/insightspublisher/index.js
@@ -8,7 +8,7 @@ const MAX_RECONNECT_ATTEMPTS = 5;
 const RECONNECT_INTERVAL_MS = 50;
 const WS_CLOSE_NORMAL = 1000;
 
-const toplevel = global.window || global;
+const toplevel = window || global;
 const WebSocket = toplevel.WebSocket ? toplevel.WebSocket : require('ws');
 const util = require('../../util');
 
--- a/lib/webrtc/rtcpeerconnection/chrome.js
+++ b/lib/webrtc/rtcpeerconnection/chrome.js
@@ -3,7 +3,7 @@
 
 var ChromeRTCSessionDescription = require('../rtcsessiondescription/chrome');
 var EventTarget = require('../util/eventtarget');
-var inherits = require('util').inherits;
+var inherits = require('../../vendor/inherits');
 var Latch = require('../util/latch');
 var MediaStream = require('../mediastream');
 var RTCRtpSenderShim = require('../rtcrtpsender');
--- a/lib/webrtc/rtcpeerconnection/firefox.js
+++ b/lib/webrtc/rtcpeerconnection/firefox.js
@@ -3,7 +3,7 @@
 
 var EventTarget = require('../util/eventtarget');
 var FirefoxRTCSessionDescription = require('../rtcsessiondescription/firefox');
-var inherits = require('util').inherits;
+var inherits = require('../../vendor/inherits');
 var updateTracksToSSRCs = require('../util/sdp').updateUnifiedPlanTrackIdsToSSRCs;
 var util = require('../util');
 
--- a/lib/webrtc/rtcpeerconnection/safari.js
+++ b/lib/webrtc/rtcpeerconnection/safari.js
@@ -2,7 +2,7 @@
 'use strict';
 
 var EventTarget = require('../util/eventtarget');
-var inherits = require('util').inherits;
+var inherits = require('../../vendor/inherits');
 var Latch = require('../util/latch');
 var sdpUtils = require('../util/sdp');
 var util = require('../util');
--- a/test/lib/webrtc/fakemediastream.js
+++ b/test/lib/webrtc/fakemediastream.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var inherits = require('util').inherits;
+var inherits = require('../../../lib/vendor/inherits');
 var randomName = require('./util').randomName;
 var EventTarget = require('../../../lib/webrtc/util/eventtarget');
 


lib/vendor/inherits.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function inherits(ctor, superCtor) {
```
charliesantos, line 1: Please add a comment at the top of this file explaining why we are pulling in this. Also add the license header, grab it from the author https://github.com/isaacs/inherits/blob/main/LICENSE. See vendor/loglevel.js for reference.

charliesantos (changes requested):
Can you please add a changelog?

lib/signaling/v2/peerconnection.js
```diff
@@ -959,7 +959,9 @@ class PeerConnectionV2 extends StateMachine {
     }
     this._log.warn('An ICE restart has been scheduled');
     this._isIceRestartBackoffInProgress = true;
-    this._iceRestartBackoff.backoff();
+    setTimeout(() => {
```
makarandp0, line 962: It does not look like new backoff is implementing the functionality that is assumed here. 

here `setTimeout` is going to fire no matter what. We need this timer to get cancelled when `this._iceRestartBackoff.reset();` is called.  The new implementation of reset does not do such thing.

makarandp0 (changes requested):
@PikaJoyce, I took a quick look at new `Backoff` implementation and it does not look like it implements the same contract. specifically there is no way to cancel the timer started by backoff.

I have not looked at inherit implementation, but I think it would best if you can add unit tests for both backoff and inherits before replacing them, so that you can see them passing before and after your changes. that would help you figure out if new backoff and inherit implementation are doing what they are supposed to do.

lib/signaling/v2/peerconnection.js
PikaJoyce: I see, I actually added a `clearTimeout` yesterday right where we called `reset` just to try some things out but removed it. Do you think that would be the proper step to take to clean up the `setTimeout`?

makarandp0: Yes we do need `clearTimeout` somewhere. Basically we need it be cleared when backoff.reset is called

I suggest that you start with some unit tests against the existing Backoff module, One of the test should execute this code path that  clears the timer, Once you have those test in place, you can replace Backoff with your implementation and ensure that the tests pass again.

PikaJoyce: Sounds good. I think it will look a bit different since our current back off module has a different API compared to our old one. Previously the timer clears the timers whereas this backoff module simply generates the numbers to be passed into a `setTimeout`. But I'll create a test on another branch with the old backoff module, and a test on this pr with similar numbers to ensure that they're indeed generating similar numbers.

lib/util/backoff.js
```diff
@@ -0,0 +1,102 @@
...
+      },
+      _factor: {
+        value: options.factor || 2
+      },
+      _attempts: {
+        value: 0,
+        writable: true
+      },
+      _duration: {
+        enumerable: true,
```
makarandp0, line 35: enumerable should be `false`

lib/util/backoff.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  /**
+  * Start the backoff operation.
+  *
+  * @return {void}
+  * @api public
+  */
+  backoff(fn) {
```
makarandp0, line 77: i didn't understand why we need `fn` parameter, once we have it in constructor.  We can talk more about it offline

lib/signaling/v2/peerconnection.js
```diff
@@ -960,7 +959,7 @@ class PeerConnectionV2 extends StateMachine {
     }
     this._log.warn('An ICE restart has been scheduled');
     this._isIceRestartBackoffInProgress = true;
-    this._iceRestartBackoff.backoff();
+    this._iceRestartBackoff.backoff(this._initiateIceRestart.bind(this));
```
makarandp0, line 962:
```suggestion
    this._iceRestartBackoff.backoff(() => this._initiateIceRestart());
```

lib/signaling/v2/peerconnection.js
```diff
@@ -191,7 +190,8 @@ class PeerConnectionV2 extends StateMachine {
           false)
       },
       _iceRestartBackoff: {
-        value: options.Backoff.exponential(iceRestartBackoffConfig)
+        // eslint-disable-next-line new-cap
+        value: new options.Backoff(null, iceRestartBackoffConfig)
```
makarandp0, line 194: why is 1st parameter null? doesn't the new function just take one parameter?

lib/signaling/v2/twilioconnectiontransport.js
```diff
@@ -147,7 +147,7 @@ class TwilioConnectionTransport extends StateMachine {
         writable: true
       },
       _reconnectBackoff: {
-        value: options.Backoff.exponential(reconnectBackoffConfig)
+        value: new options.Backoff(null, reconnectBackoffConfig)
```
makarandp0, line 150: why is 1st parameter null? doesn't the new function just take one parameter?

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,123 @@
...
+      backoff.backoff(fn);
+      fakeTimer.tick(210);
+      backoff.backoff(fn);
+      fakeTimer.tick(410);
+      sinon.assert.calledThrice(fn);
+    });
+
+    it('should reset the duration', () => {
+      backoff.backoff(fn);
+      fakeTimer.tick(110);
```
makarandp0, line 43:
```suggestion
      fakeTimer.tick(110);
      sinon.assert.calledOnce(fn);
```

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,123 @@
...
+      backoff.backoff(fn);
+      fakeTimer.tick(100);
+      backoff.backoff(fn);
+      fakeTimer.tick(200);
+      backoff.backoff(fn);
+      fakeTimer.tick(400);
+      backoff.backoff(fn);
+      fakeTimer.tick(800);
+      sinon.assert.callCount(fn, 4);
+      backoff.reset();
```
makarandp0, line 91: looks like this test isn't really verifying 'max' property since you call reset here.

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,123 @@
...
+      sinon.assert.calledThrice(fn);
+    });
+
+    it('should reset the duration', () => {
+      backoff.backoff(fn);
+      fakeTimer.tick(110);
+      backoff.reset();
+      fakeTimer.tick(210);
+      fakeTimer.tick(410);
+      sinon.assert.calledOnce(fn);
```
makarandp0, line 47: why is it getting called just once here? shouldn't it get called thrice?

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,123 @@
...
+    it('min', () => {
+      const options = {
+        min: 10,
+      };
+      const backoff = new DefaultBackoff(options);
+
+      backoff.backoff(fn);
+      fakeTimer.tick(10);
+      backoff.reset();
+      sinon.assert.calledOnce(fn);
```
makarandp0, line 73:
```suggestion
      sinon.assert.calledOnce(fn);
      backoff.reset();
```

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,123 @@
...
+      fakeTimer.tick(300);
+      backoff.backoff(fn);
+      fakeTimer.tick(900);
+      sinon.assert.calledTwice(fn);
+      backoff.reset();
+    });
+  });
+
+  describe('Callback passed into backoff method', () => {
+    it('should set the private member and call upon the function', () => {
```
makarandp0, line 111: not sure what this test is doing different from earlier tests.

test/unit/spec/util/backoff.js
PikaJoyce: Hm, since we changed how our backoff behaves, I'm thinking it's going to be hard since we don't call a function automatically (onTimer) until the max duration. We're now using max as a variable for our setTimeout duration calculation so that we don't get numbers beyond the max that was set. 

So you're correct, but I'm thinking if there's a better way to test "max" since it's no longer being used to determine if we should stop retrying, and used more simply as a means to calculate a new duration once `backoff` is called. But `_duration` is a private member of this class so it doesn't seem right to check the duration? WDYT?

test/unit/spec/util/backoff.js
PikaJoyce: No, because we only called `backoff` once. 

Previously though, it also would have been one because it should not have called `fn` after we call reset unless we start `backoff` again.

test/unit/spec/util/backoff.js
makarandp0: I think way you could verify max is to write a test that would tick at 10, 20, 40, 80, 160, 
and verify that at tickCount = 140 you get 4 calls without max, and 5 calls with max set to 40.

PikaJoyce: We currently have backoff implemented so that we only start the operation when we call `backoff`. It's no longer calling on another function to continuously retry the function. 

So for example this test:
`    it('max', () => {
      const options = {
        max: 40,
      };
      const backoff = new DefaultBackoff(options);

      backoff.backoff(fn);
      fakeTimer.tick(10);
      fakeTimer.tick(20);
      fakeTimer.tick(40);
      fakeTimer.tick(160);
      sinon.assert.callCount(fn, 1);
      backoff.reset();
    });`
    
    This will pass because we only called `backoff` once. Basically, the max duration is no longer being used to determine our attempts at calling a function anymore. It's only used to determine the max duration for our setTimeout

makarandp0 (changes requested):
@PikaJoyce - This is looking good. I have  just couple of small change requests. Also please make sure that tests pass after the changes.

test/unit/spec/util/backoff.js
```diff
@@ -0,0 +1,143 @@
...
+      fn = sinon.spy();
+      fakeTimer = sinon.useFakeTimers();
+    });
+
+    afterEach(() => {
+      fn.resetHistory();
+      fakeTimer.restore();
+    });
+
+    it('min', () => {
```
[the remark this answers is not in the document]
PikaJoyce: Oops forgot to remove this one, we already have a test for min so I'll remove that one. Thank you!

test/integration/spec/connect.js
```diff
@@ -831,7 +831,7 @@ describe('connect', function() {
   // eslint-disable-next-line no-warning-comments
   // TODO: The following tests verifies bitrates using default codec, OPUS.
   // We should also verify other codecs like ISAC, PCMU and PCMA.
-  describe('called with EncodingParameters', () => {
+  describe('called with EncodingParameters (@unstable: VIDEO-9969)', () => {
     const minAudioBitrate = 6000;
```
makarandp0, line 835: This should be marked unstable only for firefox, make sure `isFireFox` is defined
```suggestion
  describe(`called with EncodingParameters ${isFirefox ? ' - @unstable: VIDEO-9969' : ''}`, () => {
    const minAudioBitrate = 6000;
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Bug Fixes
 
 - Fixed an issue where some extraneous errors were logged to console when a video track was stopped. (VIDEO-9511)
 - Fixed an issue where the `dimensionsChanged` event was not firing when the track dimensions first became available. (VIDEO-3576)
+- Fixed issue where `global` was undefined causing run time issues in Angular. Additionally, migrated SDK usage of util.inherits to using [inherits](https://www.npmjs.com/package/inherits) to remove references to node dependencies. (VIDEO-9282)
```
makarandp0, line 13:
```suggestion
- Removed node dependancies that build errors on some platforms (VIDEO-9282)
```

makarandp0 (changes requested):
need couple more changes.

makarandp0 (approved):
Thanks @PikaJoyce - Please cut an RC!

PR #1760, opened 2022-04-21
VIDEO-8817 Integration with RSP and SFU changes for large rooms.

manjeshbhargav:
**Contributing to Twilio**

### Description

@makarandp0 @PikaJoyce 

 - [x] Changes based on integration with Room Service (RSP v3)
   - Tracks specified in ConnectOptions are now published after the connect Promise is resolved
 - [x] Changes based on integration with SFU changes
   - RemoteTracks are keyed off of Track SIDs in internal maps instead of Track IDs, since they can be associated with more than one MediaStreamTrack in large Rooms
   - Keep existing subscription behavior foe DataTracks

### Before review
* [x] Ready for review

### DIFF
--- a/lib/participant.js
+++ b/lib/participant.js
@@ -207,26 +207,26 @@ class Participant extends EventEmitter {
 
   /**
    * @private
-   * @param {RemoteTrack} track
-   * @param {Track.ID} id
-   * @returns {?RemoteTrack}
+   * @param {Track} track
+   * @param {key} key
+   * @returns {?Track}
    */
-  _addTrack(track, id) {
+  _addTrack(track, key) {
     const log = this._log;
-    if (this._tracks.has(id)) {
+    if (this._tracks.has(key)) {
       return null;
     }
-    this._tracks.set(id, track);
+    this._tracks.set(key, track);
 
     const tracksByKind = {
       audio: this._audioTracks,
       video: this._videoTracks,
       data: this._dataTracks
     }[track.kind];
-    tracksByKind.set(id, track);
-    reemitTrackEvents(this, track, id);
+    tracksByKind.set(key, track);
+    reemitTrackEvents(this, track, key);
 
-    log.info(`Added a new ${util.trackClass(track)}:`, id);
+    log.info(`Added a new ${util.trackClass(track)}:`, key);
     log.debug(`${util.trackClass(track)}:`, track);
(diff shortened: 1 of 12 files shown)


lib/participant.js
```diff
@@ -207,26 +207,26 @@ class Participant extends EventEmitter {
 
   /**
    * @private
-   * @param {RemoteTrack} track
-   * @param {Track.ID} id
-   * @returns {?RemoteTrack}
+   * @param {Track} track
+   * @param {key} key
```
makarandp0, line 211:
```suggestion
   * @param {Track} track
   * @param {string} key
```

lib/participant.js
```diff
@@ -348,14 +340,14 @@ class Participant extends EventEmitter {
         ? new RemoteTrack(sid, trackTransceiver, options)
         : new RemoteTrack(sid, trackTransceiver, isEnabled, isSwitchedOff, switchOffReason, setPriority, setRenderHint, options);
 
-      self._addTrack(track, publication, trackTransceiver.id);
+      self._addTrack(track, publication);
```
makarandp0, line 343: This is confusing. `_addTrack` method on this class takes `track` and `key` as parameters, but looks like derived class takes `track` and `publication` as parameter's.  
By invoking `self._addTrack` here with `publication` as parameter means this class knows about the derived class. I suggest we add different method to distinguish these two.
`_addTrackWithPublication`, and `_addTrackWithKey`

makarandp0 (changes requested):
looks good, but have couple of comments

lib/participant.js
manjeshbhargav: @makarandp0 You're right. This is a known issue, since the actual solution is to move all this code to the RemoteParticipant class. But, I don't want to do that in this PR.

PikaJoyce (commented):
Awesome work on the v3 stuff!! So excited to see this go out. I have a few questions though. In some places, we replace `id` with `key` but there are some places where we call it a `sid`.  I guess I might be a little lost about what the differences are between the three? Otherwise lookin' really good!

lib/participant.js
manjeshbhargav: @makarandp0 I'm going to merge this in order to cut an RC today. Thanks for the review!

PR #1768, opened 2022-05-02
VIDEO-9511 - do not configure encodings for stopped tracks

makarandp0:
For adaptive simulcast we configure simulcast layers depending on the track dimensions. This code encounters an exception and emits a console error for tracks that are stopped.  Thats because stopped track return `NaN` values for width and height [here](https://github.com/twilio/twilio-video.js/blob/82995cd8515ed384fe82acce23aa9340c2eb2f98/lib/signaling/v2/peerconnection.js#L521)

This change fixes this by not attempting to change encodings for stopped tracks. 

**Contributing to Twilio**

### Description

### Before review
* [x] Included screenshot as PR comment (if needed)
* [x] Ready for review

### DIFF
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -469,12 +469,7 @@ class PeerConnectionV2 extends StateMachine {
    * @returns {boolean} true if encodings were updated.
    */
   _maybeUpdateEncodings(track, encodings, trackReplaced = false) {
-    if (track.kind !== 'video') {
-      return false;
-    }
-
-    if (track.readyState === 'ended') {
-      // do not attempt to configure stopped tracks.
+    if (track.kind !== 'video' || track.readyState === 'ended') {
       return false;
     }
 
--- a/lib/signaling/v2/peerconnection.js
+++ b/lib/signaling/v2/peerconnection.js
@@ -472,6 +472,12 @@ class PeerConnectionV2 extends StateMachine {
     if (track.kind !== 'video') {
       return false;
     }
+
+    if (track.readyState === 'ended') {
+      // do not attempt to configure stopped tracks.
+      return false;
+    }
+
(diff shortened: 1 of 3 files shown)


test/unit/spec/signaling/v2/peerconnection.js
```diff
@@ -185,6 +185,14 @@ describe('PeerConnectionV2', () => {
         encodings: [{}, {}, {}],
         preferredCodecs: { audio: [], video: [{ codec: 'vp8', simulcast: true }] }
       },
+      {
+        browser: 'chrome',
+        testName: 'does not update encodings when track is stopped',
+        readyState: 'ended',
```
charliesantos, line 191: Can we update some of the positive tests to include a readyState with a value of 'live'? To make sure tests are still passing.

lib/signaling/v2/peerconnection.js
```diff
@@ -472,6 +472,12 @@ class PeerConnectionV2 extends StateMachine {
     if (track.kind !== 'video') {
       return false;
     }
+
+    if (track.readyState === 'ended') {
+      // do not attempt to configure stopped tracks.
+      return false;
+    }
```
charliesantos, line 479: Want to combine this check in the line above?
```js
    if (track.kind !== 'video' || track.readyState === 'ended') {
      return false;
    }
```

charliesantos (approved):
Left a couple of comments. But looks good 👍

test/unit/spec/signaling/v2/peerconnection.js
makarandp0: when readyState is not explicitly specified, test takes it as 'live'. see line 240

makarandp0: but let me change one test to explicitely pass it anyways

PR #1778, opened 2022-05-18
VIDEO-9966 - use no media_region as a signal to revert to unicast

makarandp0:
when connected with `preferredVideoCodecs="auto"` we enable adaptive simulcast. That mean we default to  simulcast and if joined a p2p room revert back to unicast. Previously we used the media_signaling payload in  `connected` response from room service, as a signal to determine if we need to revert to unicast or not. We reverted to unicast if publisher_hint MSP channel was not part of the connected response.

With large rooms, connected response  does not contain information about media_signaling. This caused us to always revert to unicast for large rooms. To fix this issue this code change now looks at `media_region` in connected response as a clue to decide on reverting to unicast. `media_region` is available only when connecting to group rooms - so If media_region is not part of connected response we will revert to unicast.

**Contributing to Twilio**

### Description

### Before review
* [x] Included screenshot as PR comment (if needed)
* [x] Ready for review

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -313,9 +313,11 @@ class RoomV2 extends RoomSignaling {
 
     this._update(initialState);
 
-    // NOTE(mpatwardhan) after initial state we know if publisher_hints are enabled or not
-    // if they are not enabled. we need to undo simulcast that if it was enabled with initial offer.
-    this._peerConnectionManager.setEffectiveAdaptiveSimulcast(this._publisherHintsSignaling.isSetup);
+    // NOTE(mpatwardhan) after initial state we know if we are connected to group room or not.
+    // if we are not connected to group room turn off adaptive simulcast.
+    if (!this.mediaRegion) {
+      this._peerConnectionManager.setEffectiveAdaptiveSimulcast(false);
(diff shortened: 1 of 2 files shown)


lib/signaling/v2/room.js
```diff
@@ -313,9 +313,9 @@ class RoomV2 extends RoomSignaling {
 
     this._update(initialState);
 
-    // NOTE(mpatwardhan) after initial state we know if publisher_hints are enabled or not
-    // if they are not enabled. we need to undo simulcast that if it was enabled with initial offer.
-    this._peerConnectionManager.setEffectiveAdaptiveSimulcast(this._publisherHintsSignaling.isSetup);
+    // NOTE(mpatwardhan) after initial state we know if we are connected to group room or not.
+    // if we are not connected to group room turn off adaptive simulcast.
```
mmartikainen, line 317:
```suggestion
    // NOTE(mpatwardhan) after initial state we can find out if we are connected to a P2P or group room by checking if there is a mediaRegion.
    // if we are not connected to group room, turn off adaptive simulcast.
```

mmartikainen (approved):
Just a suggestion for the comment

PR #1779, opened 2022-05-23
VIDEO-8282 - Adding iPad Detection

PikaJoyce:
JIRA Ticket : [VIDEO-8282](https://issues.corp.twilio.com/browse/VIDEO-8282)

Adding in iPad detection function to send to insights.

**Contributing to Twilio**

### Description

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/util/browserdetection.js
+++ b/lib/util/browserdetection.js
@@ -18,13 +18,20 @@ function isIOS() {
 }
 
 /**
+ * Detects whether or not a device is an Apple touch screen device.
+ * @returns {boolean}
+ */
+function isTabletOrMobile() {
+  return navigator && navigator.maxTouchPoints &&
+  navigator.maxTouchPoints > 2;
+}
+
+/**
  * Detects whether or not a device is an iPad
  * @returns {boolean}
  */
 function isIpad() {
-  return navigator && navigator.maxTouchPoints &&
-  navigator.maxTouchPoints > 2 &&
-  window.screen.width >=  768 &&
+  return isTabletOrMobile() && window.screen.width >=  744 &&
   (/Macintosh/i.test(navigator.userAgent) || /iPad/.test(navigator.userAgent) || /iPad/.test(navigator.platform));
 }
 
@@ -33,9 +40,7 @@ function isIpad() {
  * @returns {boolean}
  */
 function isIphone() {
-  return navigator && navigator.maxTouchPoints &&
-  navigator.maxTouchPoints > 2 &&
-  window.screen.width <=  476 &&
+  return isTabletOrMobile() && window.screen.width <=  476 &&
   (/Macintosh/i.test(navigator.userAgent) || /iPhone/.test(navigator.userAgent) || /iPhone/.test(navigator.platform));
 }
 
--- a/lib/util/browserdetection.js
+++ b/lib/util/browserdetection.js
@@ -21,7 +21,7 @@ function isIOS() {
  * Detects whether or not a device is an Apple touch screen device.
  * @returns {boolean}
  */
-function isTabletOrMobile() {
+function hasTouchScreen() {
   return navigator && navigator.maxTouchPoints &&
   navigator.maxTouchPoints > 2;
 }
@@ -31,7 +31,7 @@ function isTabletOrMobile() {
  * @returns {boolean}
  */
 function isIpad() {
-  return isTabletOrMobile() && window.screen.width >=  744 &&
+  return hasTouchScreen() && window.screen.width >=  744 &&
   (/Macintosh/i.test(navigator.userAgent) || /iPad/.test(navigator.userAgent) || /iPad/.test(navigator.platform));
 }
 
@@ -40,7 +40,7 @@ function isIpad() {
  * @returns {boolean}
  */
 function isIphone() {
-  return isTabletOrMobile() && window.screen.width <=  476 &&
+  return hasTouchScreen() && window.screen.width <=  476 &&
   (/Macintosh/i.test(navigator.userAgent) || /iPhone/.test(navigator.userAgent) || /iPhone/.test(navigator.platform));
 }
 
--- a/lib/util/browserdetection.js
+++ b/lib/util/browserdetection.js
@@ -25,6 +25,7 @@ function isIOS() {
 function isIpad() {
   return navigator && navigator.maxTouchPoints &&
   navigator.maxTouchPoints > 2 &&
+  window.screen.width >=  768 &&
   (/Macintosh/i.test(navigator.userAgent) || /iPad/.test(navigator.userAgent) || /iPad/.test(navigator.platform));
 }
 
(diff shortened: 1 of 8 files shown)


lib/util/index.js
```diff
@@ -133,6 +133,17 @@ function flatMap(list, mapFn) {
   }, []);
 }
 
+/**
+ * Detects whether or not a device is an iPad
+ * @returns {boolean}
+ */
+
+function isIpad() {
```
charliesantos, line 141: We should probably put this inside browserdetection.js

lib/util/insightspublisher/index.js
```diff
@@ -60,6 +62,9 @@ class InsightsPublisher extends EventEmitter {
       },
       _WebSocket: {
         value: options.WebSocket
+      },
+      _isIpad: {
```
charliesantos, line 66: Why create another internal property? Can we call util.isIpad() directly on line 205?

lib/util/insightspublisher/index.js
```diff
@@ -197,6 +202,11 @@ function connect(publisher, token, sdkName, sdkVersion, roomSid, participantSid,
       roomSid: roomSid,
     };
 
+    if (publisher._isIpad) {
+      connectRequest.publisher = { ...connectRequest.publisher, ...PUBLISHER_METADATA_IPAD };
+      // connectRequest.publisher = Object.assign(connectRequest.publisher, PUBLISHER_METADATA_IPAD);
```
charliesantos, line 207: Don't forget to remove commented out code.

test/unit/spec/util/index.js
```diff
@@ -336,4 +338,48 @@ describe('util', () => {
       });
     });
   });
+
+  describe('isIpad', () => {
```
charliesantos, line 342: We should move this in browserdetection file.

lib/util/insightspublisher/index.js
```diff
@@ -197,6 +202,11 @@ function connect(publisher, token, sdkName, sdkVersion, roomSid, participantSid,
       roomSid: roomSid,
     };
 
+    if (publisher._isIpad) {
```
charliesantos, line 205: Please add unit tests for this.

charliesantos (changes requested):

lib/util/index.js
```diff
@@ -133,6 +133,17 @@ function flatMap(list, mapFn) {
...
+  return navigator && navigator.maxTouchPoints &&
+  navigator.maxTouchPoints > 2 &&
+  (/Macintosh/i.test(navigator.userAgent) || /iPad/.test(navigator.userAgent) || /iPad/.test(navigator.platform));
```
makarandp0, line 144: how about iPhone - do want to report iPhones as well?

lib/util/browserdetection.js
```diff
@@ -17,6 +17,28 @@ function isIOS() {
...
+}
+
+/**
+ * Detects whether or not a device is an iPhone
+ * @returns {boolean}
+ */
+function isIphone() {
+  return navigator && navigator.maxTouchPoints &&
+  navigator.maxTouchPoints > 2 &&
+  window.screen.width <=  476 &&
```
makarandp0, line 38: just curious - does the width stay at `window.screen.width <=  476` when iPhone is rotated? same question for iPad

PikaJoyce: Yes. Just tested this on my iPhone and it's the same value!

PikaJoyce: I used the largest width from iPhones (iPhone 7 Plus was the last of this size!) and the smallest width from iPads (Starting from the "smallest" iPads which were 768 in width). 

However, I can go one bit smaller and that this helped me realize that iPad minis (744 width) which are no longer in production have a slightly smaller screen size but I'll include it as part of this.

Otherwise, there doesn't seem to be any devices currently available from Apple that have overlaps in widths for these devices.

test/unit/spec/util/index.js
```diff
@@ -11,10 +11,11 @@ const {
   makeUUID,
   promiseFromEvents,
   isChromeScreenShareTrack,
-  createRoomConnectEventPayload
+  createRoomConnectEventPayload,
 } = require('../../../../lib/util');
 
 const { sessionSID } = require('../../../../lib/util/sid');
+const { createFalse } = require('typescript');
```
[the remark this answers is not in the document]
PikaJoyce: Unsure how that even got added. Will remove this! Thanks!

lib/util/browserdetection.js
```diff
@@ -17,14 +17,21 @@ function isIOS() {
   return /iPad|iPhone|iPod/.test(navigator.userAgent);
 }
 
+/**
+ * Detects whether or not a device is an Apple touch screen device.
+ * @returns {boolean}
+ */
+function isTabletOrMobile() {
```
charliesantos, line 24: `hasTouchScreen`?

lib/util/insightspublisher/index.js
```diff
@@ -60,6 +65,12 @@ class InsightsPublisher extends EventEmitter {
       },
       _WebSocket: {
         value: options.WebSocket
+      },
+      _isIpad: {
+        value: options.isIpad()
+      },
+      _isIphone: {
+        value: options.isIphone()
```
charliesantos, line 73: @PikaJoyce I think you missed my comments here in my last 2 reviews :)
These internal properties are not needed right? Can we call isIpad() and isIphone() directly on line 211?

charliesantos (changes requested):

lib/util/insightspublisher/index.js
PikaJoyce: I've added these back in order to unit test insightspublisher, so mostly for dependency injection.

charliesantos (approved):
Good job Joyce. Let's make sure tests passes and that we also get a +1 from @makarandp0 before merging.

lib/util/insightspublisher/index.js
```diff
@@ -60,6 +65,12 @@ class InsightsPublisher extends EventEmitter {
       },
       _WebSocket: {
         value: options.WebSocket
+      },
+      _isIpad: {
```
makarandp0, line 69: @PikaJoyce , For testing, instead of passsing isIpad, isIphone in options, you can mock them using `sinon.stub` 
This lets you stub the whole `../browserdetection` module from the stub. That way you wont have to add these extra member in the class.  You can see an example of this here: https://github.com/twilio/twilio-video.js/blob/76dd81b6fcef180dbf9c030436e8a20a09639ac4/test/unit/spec/signaling/v2/peerconnection.js#L87

makarandp0 (commented):
I have one suggestion for avoiding the _isIphone, _isIpad properties. but LGTM otherwise.

lib/util/insightspublisher/index.js
PikaJoyce: I'll try this out!

makarandp0 (approved):

PR #1781, opened 2022-05-25
Exposing twilio warning

charliesantos:
**Contributing to Twilio**

## Pull Request Details

Exposing TwilioWarning that way developers can also do 

```js
import { connect, TwilioWarning } from 'twilio-video';

const room = await connect('token', {
  notifyWarnings: [TwilioWarning.recordingMediaLost]
  // Other connect options
});
```

### Before review
* [ ] Ready for review

### DIFF
--- a/tsdef/index.d.ts
+++ b/tsdef/index.d.ts
@@ -42,7 +42,6 @@ export { Room } from './Room';
 export { Track } from './Track';
 export { TrackPublication } from './TrackPublication';
 export { TwilioError } from './TwilioError';
-export { TwilioWarning } from './TwilioWarning';
 export {
   AudioCodec,
   AudioCodecSettings,
@@ -81,6 +80,7 @@ export {
   StatsReport,
   TrackStats,
   TrackSwitchOffMode,
+  TwilioWarning,
   VideoBandwidthProfileOptions,
   VideoCodec,
   VideoCodecSettings,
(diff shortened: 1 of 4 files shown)


makarandp0:
@charliesantos, It feels bit odd that we have `TwiloError` which is a class, but `TwilioWarning` is an enum. 
Why not recommend that they connect with a string value instead of exposing this enum?
```
const room = await connect('token', {
  notifyWarnings: ["recording-media-lost"]
  // Other connect options
});
```

charliesantos:
> @charliesantos, It feels bit odd that we have `TwiloError` which is a class, but `TwilioWarning` is an enum. Why not recommend that they connect with a string value instead of exposing this enum?
> 
> ```
> const room = await connect('token', {
>   notifyWarnings: ["recording-media-lost"]
>   // Other connect options
> });
> ```

I was thinking to do it this way to make it more strict on TS environment. But I understand your point. Let me adjust it for TS.

charliesantos:
@makarandp0 can you please take another look?

tsdef/index.d.ts
```diff
@@ -81,6 +80,7 @@ export {
   StatsReport,
   TrackStats,
   TrackSwitchOffMode,
+  TwilioWarning,
```
makarandp0, line 83: is this needed anymore?

charliesantos: Yes for TS. That way we can do this

```ts
videoPublication?.on('warning', (name: TwilioWarning) => {

});
```

makarandp0:
LGTM!

PR #1786, opened 2022-05-29
fix dimensionsChanged/videoWidth bug

gburt:
Avoid 'Cannot read property of undefined/null' by checking if elem is still set within the func called by the resize callback (where previously it was only guarded before setting up the callback).

I didn't dive deeper to understand _why_ `this._dummyEl` is null when this onresize callback is called - I guess because `_end()` was called and null'd it out.  Seems unlikely to cause any real issues though, since the net effect is just to guard _setting values on the now-null _dummyEl_ anyway.

We've had over a dozen uesrs run into this bug in production.  All on iOS/Safari.
image

### DIFF
--- a/lib/media/track/videotrack.js
+++ b/lib/media/track/videotrack.js
@@ -457,6 +457,10 @@ class VideoTrack extends MediaTrack {
 VideoTrack.DIMENSIONS_CHANGED = 'dimensionsChanged';
 
 function dimensionsChanged(track, elem) {
+  if (!elem) {
+    return false;
+  }
+
   return track.dimensions.width !== elem.videoWidth
     || track.dimensions.height !== elem.videoHeight;
 }


makarandp0:
@charliesantos, Can you please take a look - This looks safe fix - although I do not fully understand why _dummyEl is null here.

charliesantos:
The code change looks safe but we should understand why this is happening before merging the PR. I don't understand why the element would be null in this function because the handlers are attached to the element itself. Maybe this error is coming from a different line? Do you have steps to reproduce @gburt ?

gburt:
> The code change looks safe but we should understand why this is happening before merging the PR. I don't understand why the element would be null in this function because the handlers are attached to the element itself. Maybe this error is coming from a different line? Do you have steps to reproduce @gburt ?

@charliesantos no, sorry, this is just stuff captured by our production error logging.  But as I noted in the PR description, the `_dummyEl` _does_ get null'd out in the `_end()` call, without unregistering the onresize handler.

gburt:
@charliesantos any more thoughts on this? We continue to have folks hitting this in production.  Does my thinking that it's been nulled out without unregistering the onresize handler make sense to you?

gburt:
@manjeshbhargav did you mean to close this?  Was this issue resolved in some other PR/commit?

gburt:
@manjeshbhargav ping - we're still seeing this bug in production

edleatherbury:
@gburt @manjeshbhargav We still see this in production as well, version 2.26.2.

`Cannot read properties of null (reading 'videoWidth') at _dummyEl._dummyEl.onresize [as __zone_symbol__ON_PROPERTYresize]`

`Cannot read properties of null (reading 'videoWidth') at _dummyEl._dummyEl.onloadedmetadata [as __zone_symbol__ON_PROPERTYloadedmetadata]`

PR #1795, opened 2022-06-02
large room audio dummy element handling 

makarandp0:
In large rooms, as we switch m-lines RemoteAudioTracks are attached to new MediaStreamTrack. In this process we were creating new audio elements to listen to "started" event. Creation of new audio elements and attaching srcObject properties causes creation of new WebMediaPlayers. With frequent track switching (with large room's n-loudest algorithm) causes very frequent creation and destruction of media players. 

This change avoid creating/destroying web media players by letting dummy elements stay longer. This essentially undo's the optimization introduced in https://github.com/twilio/twilio-video.js/pull/1534. Its okay because chrome 92 [changes](https://bugs.chromium.org/p/chromium/issues/detail?id=1232649) that prompted  https://github.com/twilio/twilio-video.js/pull/1534 were reverted. 
**Contributing to Twilio**

### Description

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -125,7 +125,9 @@ class MediaTrack extends Track {
 
     const self = this;
     log.debug('Initializing');
-    this._dummyEl = this._createElement();
+    if (!this._dummyEl) {
+      this._dummyEl = this._createElement();
+    }
 
     this.mediaStreamTrack.addEventListener('ended', function onended() {
       self._end();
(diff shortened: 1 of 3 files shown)


charliesantos:
@makarandp0 if we're destroying the media player, it shouldn't cause a leak correct? Is this PR just an optimization so we don't have to create/destroy media players, or is it really fixing a leak?

makarandp0:
> @makarandp0 if we're destroying the media player, it shouldn't cause a leak correct? Is this PR just an optimization so we don't have to create/destroy media players, or is it really fixing a leak?

Thanks correct. There weren't real leak of media players, but just too many getting created and destroyed. This change avoids those.

lib/media/track/mediatrack.js
```diff
@@ -195,9 +197,9 @@ class MediaTrack extends Track {
       mediaStream.addTrack(mediaStreamTrack);
     }
 
-    // NOTE(mpatwardhan): resetting `srcObject` here, causes flicker (JSDK-2641), but it lets us
-    // sidestep the chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1052353
-    el.srcObject = mediaStream;
+    if (!el.srcObject) {
+      el.srcObject = mediaStream;
+    }
```
charliesantos, line 202: Is there a situation where we have to update the srcObject after it's already set?

makarandp0: yes, for dummyElement now _attach gets called everytime track switches. But in that case it woudl already have a srcObject from previous track.

charliesantos (approved):

PR #1809, opened 2022-06-18
integration with krisp rsp rsp

makarandp0:
**Contributing to Twilio**

### Description

This change integrates the noise cancellation with RSP messages. Now room service will let client know which noise cancellation vendors are supported in connected RSP message. 
-  Client will disable noise cancellation if its not supported for the room

Related: [BluePrint](https://docs.google.com/document/d/1tN1zqI9TezffMwjnfP1xCB68jh9kD9aeZ9yMt8zG7-s/edit?usp=sharing), [RSP](https://code.hq.twilio.com/client/room-signaling-protocol/pull/86)
## Burndown

TODO: add integration tests.

### Before review
* [ ] Ready for review


charliesantos:
@makarandp0 do we need to update IDL?

lib/media/track/noisecancellationimpl.ts
```diff
@@ -65,6 +71,14 @@ export class NoiseCancellationImpl implements NoiseCancellation {
     }
     return processedTrack;
   }
+
+  /**
+   * disables noise cancellation permanently.
+   */
+  disablePermanently(): Promise<void> {
```
charliesantos, line 78: Instead of another public API, what are your thoughts on update `.disable()` and add a boolean value?

```ts
disable(permanently: boolean = false): Promise<void>;
```

lib/signaling/v2/room.js
```diff
@@ -52,6 +52,13 @@ class RoomV2 extends RoomSignaling {
     }, options);
 
     localParticipant.setBandwidthProfile(options.bandwidthProfile);
+
+    const { options: { signaling_region: signalingRegion, audio_processors: audioProcessors = [] } } = initialState;
+    localParticipant.setSignalingRegion(signalingRegion);
+
+    audioProcessors.push('rnnoise');
```
charliesantos, line 59: Are we going to ship this feature with rnnoise by default?

charliesantos (changes requested):
Left a few questions @makarandp0 . Also, I noticed there are no tests for the newly added logic. Am I missing them?

makarandp0:
> @makarandp0 do we need to update IDL?

no, because this does not change any public interface. this  API - is internal to the SDK.

charliesantos:
> > @makarandp0 do we need to update IDL?
> 
> no, because this does not change any public interface. this API - is internal to the SDK.

Do we need to at least update this? https://code.hq.twilio.com/client/video-sdk-api/blob/master/idl/noisecancellation.md

lib/signaling/v2/room.js
makarandp0: No, this is internal use only.

lib/localparticipant.js
```diff
@@ -128,7 +128,9 @@ class LocalParticipant extends Participant {
    * @returns {void}
    */
   _addLocalTrack(track, priority) {
-    this._signaling.addTrack(track._trackSender, track.name, priority);
+    // check if track has noise cancellation enabled.
+    const vendor = track?.noiseCancellation?.vendor;
```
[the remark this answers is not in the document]
makarandp0: Optional chaining is needed since `track.noiseCancellation` can be null. But I think your question is valid - we don't  need it for track I will change to 

`track?.noiseCancellation?.vendor` => `track.noiseCancellation?.vendor`

makarandp0:
> @makarandp0 do we need to update IDL?

@charliesantos, no since there is no public API changes in this.

makarandp0:
> > > @makarandp0 do we need to update IDL?
> > 
> > 
> > no, because this does not change any public interface. this API - is internal to the SDK.
> 
> Do we need to at least update this? https://code.hq.twilio.com/client/video-sdk-api/blob/master/idl/noisecancellation.md

This is the public interface, We do not want to expose this method publicly - so its not part of the interface, but exposed only by the `noisecancellationImpl`

makarandp0:
@manjeshbhargav, @charliesantos Now that RSP changes are deployed to stage, I have added integration tests to the PR: https://github.com/twilio/twilio-video.js/pull/1809/commits/2b607aca0a987819c00daa5275fc3cdc1e3d28d9 

Please take another look.

charliesantos (approved):
It'd be great if we can test whether noise has been removed or not :)

PR #1852, opened 2022-08-10
adds getVersion and IsSupported checks for audio plugin

makarandp0:
Audio Plugins are now required to support getVersion and isSupported functions. Those will be used by SDK when loading the audio plugins. When version or isSupported mismatch is found SDK will not use noise cancellation plugin.

**Contributing to Twilio**

### Description

### Before review
* [ ] Ready for review


charliesantos (commented):
@makarandp0 how does this work if the sdk supports multiple versions?

makarandp0:
> @makarandp0 how does this work if the sdk supports multiple versions?

@charliesantos, SDK will require that major version must match and minor version must be greater than supported minor version. As long as those parameters work, sdk can support multiple minor versions.

charliesantos (commented):
I have a few concerns on this
- Since the new assets file are for testing only, I don't think they belong in the root directory. Maybe put them inside tests folder?
- The way I understand it, if some customers update the sdk that has a new minor version support, it breaks all customers that are still on older version of the plugin. Should it be backwards compatible? That is, only throw the error if the customer updated the plugin but did not update the video sdk
- Do we really need to do this? If we're using semver for both the sdk and the plugin, the rules are already implied correct?

@manjeshbhargav can you chime in?

makarandp0:
got +1 from @charliesantos offline: https://twilio.slack.com/archives/C01LMDCUF37/p1660579253767719?thread_ts=1660253627.569549&cid=C01LMDCUF37

Issue #1880, opened 2022-09-17
Noise cancellation requires allowing `blob:` in CSP


Dosant:
After upgrading to 2.24.1, noise cancellation stopped working for us. 

When trying to load the noise cancellation code, there is a CSP-related error and the noise cancellation code fails to load: 

```
dynamicimport.js:40 Refused to load the script 'blob:http://localhost:4200/f39c03f4-9122-4e7f-b7ed-fb08f96dc102' because it violates the following Content Security Policy directive: "script-src 'self'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

Looks like this is caused by the workaround introduced in https://github.com/twilio/twilio-video.js/pull/1872, and to make the feature work with this workaround, we'll have to loosen up the CSP to allow `blob:` which could potentially increase the attack surface. 

Note that 2.24.0 worked fine in our environment without [the import workaround](https://github.com/twilio/twilio-video.js/pull/1872).

I suggest the team to consider if it is worth keeping the workaround considering this CSP downside.

PikaJoyce:
Hi @Dosant,

Thank you for opening this issue with us. I'll bring this up to the team and will respond on this thread as soon as we have any updates! 

Best,
Joyce

Dosant:
> Thank you for opening this issue with us. I'll bring this up to the team and will respond on this thread as soon as we have any updates!

Thank you!

> However, I do have a question. Are you using the relative path for the sdk assets? Or are you trying to load the krisp SDK from another domain?

relative path. the CSP error happens it even tries to download krisp. It happens when it tries to execute script from blob

PikaJoyce:
Hey @Dosant,

Gotcha, thanks for your response. I've created a an internal ticket (VIDEO-11537) to track this issue! I'll continue to post updates to this thread as we progress through our investigations. 

Thank you for your patience!

Best,
Joyce

PikaJoyce:
Hi @Dosant, 

The team has begun investigations for this. I was wondering if it was possible for you to provide an example of this happening as I haven't been able to reproduce this issue across several applications. Thanks!

Best,
Joyce

Dosant:
@PikaJoyce, to reproduce this, you'll need an app that has a [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) configured. For example, the one where all content can come only from the site's own origin: 

```
Content-Security-Policy: default-src 'self'
```

[...]

PR #1885, opened 2022-09-26
VIDEO-11537 Reverting to using import(), but within a Function.

manjeshbhargav:
**Contributing to Twilio**

### Description

Since we [don't want to depend on Blob](https://github.com/twilio/twilio-video.js/issues/1880) anymore, the dynamic import module will now revert to calling `import()` like before, but within a Function whose body is specified as a string, so TypeScript and Webpack (and other frameworks) should not have a problem with it. However, this change requires that any application using the CSP directive `default-src self` must also add the `unsafe-eval` directive, which is already required by the Krisp Noise Cancellation plugin.

### Before review
* [x] Ready for review


README.md
```diff
@@ -154,6 +154,14 @@ you should also include the following `script-src` directive:
 script-src https://sdk.twilio.com
 ```
 
+If you are enabling [Krisp Noise Cancellation](https://www.twilio.com/docs/video/noise-cancellation) for
+your local audio, and you are using the following `default-src self` directive, you should also add the
+`unsafe-eval` directive:
+
+```
+default-src self unsafe-eval
+```
```
PikaJoyce, line 163:
```
default-src 'self' 'unsafe-eval';
```

lib/util/dynamicimport.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
```
PikaJoyce, line 2: I'm not sure if there should be a note here for us to come back around to fix this in the future? Might be useful for future us as a reminder.

PikaJoyce (approved):
Thanks Manjesh, this is a super awesome fix! 

I've also tested this locally and it works with those new directives! Left a few comments otherwise it LGTM 💯

README.md
manjeshbhargav: What you've written is the client-side meta tag syntax. What I've specified is the server side syntax.

lib/util/dynamicimport.js
manjeshbhargav: I've added a note above the line where we create the function.

PR #1923, opened 2022-11-09
VIDEO-11859 Select the actual default device ID instead of "default" (Chrome).

manjeshbhargav:
**Contributing to Twilio**

### Before review
* [x] Ready for review

### DIFF
--- a/lib/media/track/localaudiotrack.js
+++ b/lib/media/track/localaudiotrack.js
@@ -35,15 +35,12 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
     super(mediaStreamTrack, options);
 
     const { _log: log } = this;
-    const settings = mediaStreamTrack.getSettings();
+    const { label: defaultDeviceLabel = '' } = mediaStreamTrack;
+    const { deviceId: defaultDeviceId = '', groupId: defaultGroupId = '' } = mediaStreamTrack.getSettings();
 
     Object.defineProperties(this, {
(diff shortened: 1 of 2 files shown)


lib/media/track/localaudiotrack.js
```diff
@@ -35,14 +35,20 @@ class LocalAudioTrack extends LocalMediaAudioTrack {
...
+    const settings = mediaStreamTrack.getSettings();
 
     Object.defineProperties(this, {
       _currentDefaultDeviceInfo: {
-        value: { deviceId: '', groupId: '', label: '' },
+        value: {
+          deviceId: settings.deviceId,
+          groupId: settings.groupId,
+          label: mediaStreamTrack.label
+        },
```
PikaJoyce, line 46: In this case, is there ever a case where `deviceId` or any of the settings come back as undefined? Should that case be handled? 

Something like 
```JavaScript
value: {
    deviceId: typeof settings.deviceId === 'string' ? settings.deviceId : '',
    ...
    }
```

PikaJoyce (commented):
Awesome! I have one quick question, but it otherwise LGTM. Do we need to write any tests for this feature as well?

manjeshbhargav:
> Awesome! I have one quick question, but it otherwise LGTM. Do we need to write any tests for this feature as well?

I've already written the tests in an earlier PR.

PikaJoyce (approved):

Issue #1968, opened 2023-01-20
Cannot read properties of null (reading 'frameWidth') at getstats


Dosant:
Unfortunately, I don't have a reproduction for this issue, but this error relatively often occurs in our error monitoring. 

The stack trace is: 

```
TypeError: Cannot read properties of null (reading 'frameWidth')
  at <anonymous>(webpack:/./node_modules/twilio-video/es5/webrtc/getstats.js:490:59)
  at Array.forEach(<anonymous>)
  at standardizeChromeOrSafariStats(webpack:/./node_modules/twilio-video/es5/webrtc/getstats.js:449:17)
  at apply(webpack:/./node_modules/twilio-video/es5/webrtc/getstats.js:318:21)
```

And the error is coming from here: 

https://github.com/twilio/twilio-video.js/blob/6b625db655e8f6e23c9232616adb24bb2c5886c0/lib/webrtc/getstats.js#L542

The type error happens because `track` is null. 

I don't know what's precisely leads to this error. It happens sproadically during the call. All the errors I see are coming from desktop users on Chrome 109.0.0.
I also don't know what is the impact on the end users.

I would appreciate help in troubleshooting this error. If there is anything else I should attach, please let me know.
I didn't try to upgrade to a newer version, as I didn't see anything related to this error mentioned in the release notes.

**Software versions:**

 -  Browser(s): Chrome 109.0.0
 - Operating System: Mac OS / Windows
 -  twilio-video.js: 2.24.3
 -  Third-party libraries (e.g., Angular, React, etc.): Angular

fippo:
This looks like @henbos [removal of the deprecated track and stream stats in Chrome 109](https://groups.google.com/g/discuss-webrtc/c/Bxrxr8gw08Y/m/RGn-iME-AgAJ)
If you can reproduce this when starting Chrome with --force-fieldtrials=WebRtcUnshipDeprecatedStats/Enabled/ (or it reproduces in Beta/Canary) then that is it.

henbos:
Thanks @fippo. If you want the "input" frame width (as opposed to the sent frame width, which can be lower if encoding in a downscaled resolution) the place to look is the [RTCVideoSourceStats](https://w3c.github.io/webrtc-stats/#dom-rtcvideosourcestats) which has type "media-source". It is referenced by the RTP stream's [mediaSourceId](https://w3c.github.io/webrtc-stats/#dom-rtcoutboundrtpstreamstats-mediasourceid).

[...]

Issue #1960, opened 2023-01-04
Type mismatch between mediatrack.js and AudioTrack.d.ts and VideoTrack.d.ts


CamilleDrapier:
- [x] I have verified that the issue occurs with the latest twilio-video.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-video.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/twilio/video-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

```js
track.detach(element).forEach
```

This sometimes (when passed a non-falsy parameter) produces the following error: 

```
track.detach(...).forEach is not a function
``` 

**Expected behavior:**

There is no problem, or developers that are using `detach` are invited by TS to check for the type of result before proceeding with it.

**Actual behavior:**

The reason why we wrote `track.detach(...).forEach` is because the Typescript definition for [AudioTrack](https://github.com/twilio/twilio-video.js/blob/2.26.0/tsdef/AudioTrack.d.ts#L10) or [VideoTrack](https://github.com/twilio/twilio-video.js/blob/2.26.0/tsdef/VideoTrack.d.ts) says that `detach` always returns an array.  But in the implementation of [mediaTrack's detach](https://github.com/twilio/twilio-video.js/blob/2.26.0/lib/media/track/mediatrack.js#L253) seems to either return an element or an array (depending on the passed 'element').

Could we consider either:
- Changing the implementation so it follows the typing (this might be a breaking change, so it might be difficult to push as a simple update)
- Changing the TS definition so it accounts for the function returning either an element (or `undefined` in some edge cases maybe) or an array of elements.

**Software versions:**

 - [x] Browser(s): Chrome Version:108.0.0
 - [x] Operating System: Windows Version:10
 - [x] twilio-video.js: 2.26.0
 - [x] Third-party libraries (e.g., Angular, React, etc.): None relevant?

PikaJoyce:
Hi @CamilleDrapier,

[...]

PR #1970, opened 2023-01-25
VIDEO-12411 [Chrome 112+][Room.getStats()] Read from RTCMediaSourceStats instead of RTCMediaStreamTrackStats

manjeshbhargav:
**Contributing to Twilio**

### Description

* Fix for #1968.
* Fix for #1960.

### Before review
* [x] Ready for review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The Twilio Programmable Video SDKs use [Semantic Versioning](http://www.semver.o
 2.26.1 (in progress)
 ====================
 
+- Fixed a bug that manifests on Chrome versions 112+ where `Room.getStats()` rejects the returned Promise due to a
+  TypeError caused by trying to read from the now-removed `RTCMediaStreamTrackStats`. Instead, the SDK now reads from
+  the `RTCMediaSourceStats`. (VIDEO-12411) 
 - Fixed an error in the type definition for `createLocalAudioTrack()`. (VIDEO-12383)
 
 2.26.0 (December 14, 2022)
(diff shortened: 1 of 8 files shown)


PikaJoyce (approved):

schayapathi (approved):

PikaJoyce (approved):

henbos:
FYI this is not "Chrome 112+". The entire track stats object is removed-by-default in M111, but even before then there is an experiment (currently 50% Stable) that removes track in M109. So consider this an M109+ change.

anna-vasilko:
Hi @henbos ,
Just wanted to double check that the latest plan (based on [this comment](https://groups.google.com/a/chromium.org/g/blink-dev/c/RsIktnGhHqw/m/K5bMuDIzAAAJ)) is to revert the experiment, correct?
At Twilio we really hope the experiment can be reverted asap given that it is breaking for some of our customers and their end users in the wild. The adoption of our new sdk release (with this PR change) will take long time for most of the customers as they need to 1) learn about the issue 2) pick-up the new sdk version 3) roll out their applications with the new sdk version. For some customers it takes months. 
Basically the only way to really stop the breaking impact is to stop the experiment and give some notice and time to adopt the changes.

I have to acknowledge it took unusually long time for us to discover this breaking change and act on it. Holidays probably played a role here. But we will do a retrospective internally to understand the reasons better and to catch such issues sooner in the future.

henbos:
Hi @anna-vasilko, as this was causing more trouble than I had anticipated, the experiment is being rolled back to 0% and the rollback should go live on Monday, 9AM Pacific Time. Please note that users need to restart their browser to get the new config.

Please provide input on the new [Intent to Deprecate](https://groups.google.com/a/chromium.org/g/blink-dev/c/NZVXsJQ7tV8). Is my understanding correct that the reason this breakage was not discovered in testing is that the test environments run with clean profiles that don't get the finch experiment configs? While I've not had the intent approved yet, I did propose to use a strategy of having enabled-by-default (in M111 or M112) but using the Finch config to overwrite the experiment to an appropriate percentage (0% for now). My thinking is that this would act as a "fire alarm" to any test environments that don't get finch configs, but that real users (assuming they have finch) would not get the experiment until further rollout. Would that help here or cause more trouble?

For twilio-video.js, please make sure there are no hard dependencies on stats of type "track" or "stream". A general comment on parsing any metric is that it is highly advisable never to assume a metric or stats object exist but to first check that it is not `undefined` before reading.

Lastly, I'm sorry this issue affected you.

PR #2080, opened 2025-03-04
[VBLOCKS-3977] Polyfill getStats in VDI

luisrivas:
### JIRA link(s):
- [VBLOCKS-3977](https://twilio-engineering.atlassian.net/browse/VBLOCKS-3977)

### Description
We require `getStats()` to be able to accept a track as a selector and retrieve only relevant information. Some WebRTC implementations do not support this functionality, so this polyfills it.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/webrtc/getstats.js
+++ b/lib/webrtc/getstats.js
@@ -11,7 +11,9 @@ const isFirefox = guess === 'firefox';
 const isSafari = guess === 'safari';
 
 const chromeMajorVersion = isChrome ? guessVersion.major : null;
-const log = new Log('webrtc', null);
+const log = new Log('webrtc', {
+  toString: () => 'getStats'
+});
 
 const CHROME_LEGACY_MAX_AUDIO_LEVEL = 32767;
 
--- a/lib/webrtc/getstats.js
+++ b/lib/webrtc/getstats.js
@@ -621,9 +621,9 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
         .then(stats => {
           log.info('getStats by default successful');
           const isRemote = isRemoteTrack(peerConnection, track);
-          log.info(`Starting filtering stats for ${isRemote ? 'remote' : 'local'} track with id: ${track.id}`);
+          log.info(`Starting filtering stats for ${isRemote ? 'remote' : 'local'} track`);
           const filteredStats = filterStatsByTrack(stats, track, isRemote);
-          log.info(`Completed filtering stats for ${isRemote ? 'remote' : 'local'} track with id: ${track.id}`);
+          log.info(`Completed filtering stats for ${isRemote ? 'remote' : 'local'} track`);
           return standardizeChromeOrSafariStats(filteredStats, options);
         });
     });
--- a/lib/webrtc/getstats.js
+++ b/lib/webrtc/getstats.js
@@ -591,20 +591,6 @@ function getTrackStats(peerConnection, track, options = {}) {
   return Promise.reject(new Error('RTCPeerConnection#getStats() not supported'));
 }
 
-/**
- * Safely logs a message using the provided logger instance.
- * If no logger is provided, this function performs no operation.
- *
- * @param {object} [logger] - Optional logger instance with logging methods (e.g., info, warn, error).
- * @param {string} level - Logging level method name (e.g., 'info', 'warn', 'error').
- * @param {string} message - Message to log.
- * @returns {void}
- */
-function safeLog(logger, level, message) {
-  if (logger) {
-    logger[level](message);
-  }
-}
 
 /**
  * Get the standardized statistics for a particular MediaStreamTrack in Chrome or Safari.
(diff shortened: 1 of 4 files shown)


charliesantos (commented):
I'm still reviewing but I have a question. Do we have existing unit tests for getstats.js? If yes, can you please update it and add test cases for these changes?

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
       }, null, reject);
     });
   }
-  return peerConnection.getStats(track).then(response => {
-    return standardizeChromeOrSafariStats(response, options);
-  });
```
charliesantos, line 334: Do we also need to update this [code section](https://github.com/twilio/twilio-video.js/blob/master/lib/preflight/getCombinedConnectionStats.ts#L35) since it's using pc.getStats()

lib/webrtc/getstats.js
```diff
@@ -13,6 +13,242 @@ const chromeMajorVersion = isChrome ? guessVersion.major : null;
...
+ * It includes specific logic to filter the statistics report returned by Citrix and should
+ * only be used when getStats(track) fails.
+ *
+ * @param {RTCStatsReport|Array<RTCStats>} arrayOrMap - Full stats report or array of stats
+ * @param {MediaStreamTrack} track - The track to filter by
+ * @param {boolean} [isRemote=false] - Whether this is a remote track
+ * @returns {RTCStatsReport} Filtered stats
+ * @private
+ */
+function filterStatsByTrack(arrayOrMap, track, isRemote = false) {
```
charliesantos, line 41: I like how well commented this method is :)

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
       }, null, reject);
     });
   }
-  return peerConnection.getStats(track).then(response => {
-    return standardizeChromeOrSafariStats(response, options);
-  });
+
+  return peerConnection.getStats(track)
+    .then(response => {
+      return standardizeChromeOrSafariStats(response, options);
```
charliesantos, line 611: Add an info level log 'getStats by track successful'.

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
...
+    })
+    .catch(() => {
+      // NOTE(lrivas): Citrix doesn't support track-specific getStats,
+      // so this workaround tries getting all stats and filtering by track.
```
charliesantos, line 615: Add a warn level log 'getStats by track failed. Getting default stats'

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
...
+      return peerConnection.getStats()
+        .then(stats => {
```
charliesantos, line 617: Before isRemote, add info level log 'get default stats successful'

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
...
+          const isRemote = isRemoteTrack(peerConnection, track);
```
charliesantos, line 618: After remote, add info level log. Log the value of isRemote.

lib/webrtc/getstats.js
```diff
@@ -329,9 +605,21 @@ function chromeOrSafariGetTrackStats(peerConnection, track, options) {
...
+          const filteredStats = filterStatsByTrack(stats, track, isRemote);
```
charliesantos, line 619: Add an info level log before and after calling filterStatsByTrack.

lib/webrtc/getstats.js
luisrivas: @charliesantos We don't have support for custom WebRTC overrides for preflight, that is the reason why I left that section as it is. However, we could consider supporting WebRTC overrides for preflight.

lib/webrtc/getstats.js
luisrivas: Done!

lib/webrtc/getstats.js
luisrivas: Done!

lib/webrtc/getstats.js
luisrivas: Done!

lib/webrtc/getstats.js
luisrivas: Done. I combined the wording to show the value of the isRemote and the pre/post-filtering messages.

lib/webrtc/getstats.js
```diff
@@ -10,9 +11,246 @@ const isFirefox = guess === 'firefox';
 const isSafari = guess === 'safari';
 
 const chromeMajorVersion = isChrome ? guessVersion.major : null;
+const log = new Log('webrtc', 'getStats');
```
charliesantos, line 14: If we're creating a new log instance, we need to use the loggerName and loglevels passed from connect(). I'm expecting logger initialization would like something like [this](https://github.com/twilio/twilio-video.js/blob/master/lib/connect.js#L213), or [this](https://github.com/twilio/twilio-video.js/blob/master/lib/signaling/v2/peerconnection.js#L147). If it's too much work at this time, I'm ok putting this work post beta.

charliesantos (changes requested):

charliesantos (approved):

PR #2082, opened 2025-03-12
[VBLOCKS-4179] [VBLOCKS-4191] fix: MediaTrack lifecycle on Citrix

luisrivas:
### JIRA link(s):
- [VBLOCKS-4179](https://twilio-engineering.atlassian.net/browse/VBLOCKS-4179)
- [VBLOCKS-4191](https://twilio-engineering.atlassian.net/browse/VBLOCKS-4191)

### Description
This PR removes the burden of manually attaching and detaching MediaStream to elements on Citrix by extracting that logic from the consumer and putting it in the SDK. After this change, customers using Citrix should be able to call `track.attach(element)` or `track.detach(element)`, maintaining the same API.

### Changes
- Added support for [trackStarted](https://sdk.twilio.com/js/video/releases/2.29.0/docs/Room.html#event:trackStarted) event on Citrix  environments.
- Updated MediaTrack to validate `MediaStream` constructor with an additional check for `createMediaStream` method.
- Introduced `_prepareCustomMediaElement` and `_removeMediaStream` methods to handle custom MediaStream mapping and disposal for Citrix.

The property names chosen for this `MediaStream` custom constructor are still under active development and can be changed.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/media/track/mediatrack.js
+++ b/lib/media/track/mediatrack.js
@@ -45,7 +45,7 @@ class MediaTrack extends Track {
       MediaStream
     }, options);
 
-    if (typeof options.MediaStream !== 'function') {
+    if (typeof options.MediaStream !== 'function' && typeof options.MediaStream?.createMediaStream !== 'function') {
       throw new Error('MediaTrack received an invalid MediaStream constructor: ' + options.MediaStream);
     }
 
@@ -132,7 +132,7 @@ class MediaTrack extends Track {
         self._end();
         self.mediaStreamTrack.removeEventListener('ended', onended);
       });
-    } else if (this.mediaStreamTrack && this.mediaStreamTrack.onended) {
+    } else if (this.mediaStreamTrack) {
       this.mediaStreamTrack.onended = function onended() {
         self._end();
         self.mediaStreamTrack.onended = null;
@@ -190,10 +190,7 @@ class MediaTrack extends Track {
    * @private
    */
   _attach(el, mediaStreamTrack = this.processedTrack || this.mediaStreamTrack) {
-    let mediaStream = el.srcObject;
-    if (!(mediaStream instanceof this._MediaStream)) {
-      mediaStream = new this._MediaStream();
-    }
+    let mediaStream = this._getOrCreateMediaStream(el);
 
     const getTracks = mediaStreamTrack.kind === 'audio'
       ? 'getAudioTracks'
@@ -204,9 +201,13 @@ class MediaTrack extends Track {
     });
     mediaStream.addTrack(mediaStreamTrack);
 
+    // NOTE(lrivas): This is a workaround for Citrix, which requires the element
+    // to be mapped before attaching the media stream
+    if (this._MediaStream.mapElement) {
+      this._prepareCustomMediaElement(el);
+    }
     // NOTE(mpatwardhan): resetting `srcObject` here, causes flicker (JSDK-2641), but it lets us
     // to sidestep the a chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1052353
-    //
     el.srcObject = mediaStream;
     el.autoplay = true;
     el.playsInline = true;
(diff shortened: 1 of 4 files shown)


lib/media/track/mediatrack.js
```diff
@@ -293,6 +344,27 @@ class MediaTrack extends Track {
...
+  /**
+   * @private
+   * This method is used to remove the MediaStream from the given element
+   * NOTE(lrivas): Citrix has a custom implementation of MediaStream,
+   * so we need to handle it differently.
+   */
+  _removeMediaStream(el) {
+    const mediaStream = el.srcObject;
+    const hasDisposeElementMethod =
+      typeof this._MediaStream.disposeElement === 'function';
```
charliesantos, line 356: If I understand this correctly, `this._MediaStream` is from `options.MediaStream`, which is one of the overrides. Is this `MediaStream` a Citrix object from their SDK? If so, can you please show me where it's documented? I cannot see `MediaStream.disposeElement` method in UCSDK v4 doc :(

luisrivas: You are correct that `this._MediaStream` comes from `options.MediaStream`. However, Citrix does not provide a `MediaStream` object that we can instantiate. Instead, they offer a `createMediaStream` method, which serves as a factory for creating `MediaStream` objects. As a result, I ended up drafting an API that looks like this:

```js
connect('token', {
  ...connectionOptions,
  MediaStream: {
    createMediaStream: function (streamTracks = []) {
      return window.CitrixWebRTC.createMediaStream(streamTracks);
    }
  },
})
```

The constructor can now be used conditionally within the SDK if available. Since the custom `MediaStream` requires extra steps for attaching or detaching media elements, I added the `disposeElement` and `mapElement` functions for management. This is why they may not appear under the same name in the documentation. Users will need to provide those methods in the following format:

```js
connect('token', {
  ...connectionOptions,
  MediaStream: {
    createMediaStream: function (streamTracks = []) {
      return window.CitrixWebRTC.createMediaStream(streamTracks);
    }
  },
  mapElement: function (el) {
    const tagName = el.tagName.toLowerCase();
    if (tagName === 'video') {
      window.CitrixWebRTC.mapVideoElement(el);
    } else if (tagName === 'audio') {
      window.CitrixWebRTC.mapAudioElement(el);
    }
  },
  disposeElement: function (el) {
    const tagName = el.tagName.toLowerCase();
    if (tagName === 'video') {
      window.CitrixWebRTC.disposeVideoElement(el);
    } else if (tagName === 'audio') {
      window.CitrixWebRTC.disposeAudioElement(el);
    }
  },
  audioLoadingTimeout: 250
})
```

Alternatively, we could add media element validation to the SDK and receive only the following: 

``` js
connect('token', {
  ...connectionOptions,
  MediaStream: {
    createMediaStream: function (streamTracks = []) {
      return window.CitrixWebRTC.createMediaStream(streamTracks);
    },
    mapVideoElement: window.CitrixWebRTC.mapVideoElement,
    mapAudioElement: window.CitrixWebRTC.mapAudioElement,
    disposeVideoElement:  window.CitrixWebRTC.disposeVideoElement,
    disposeAudioElement:  window.CitrixWebRTC.disposeAudioElement,
    audioLoadingTimeout: 250
  },
})
```

In both cases, I include the methods inside the `MediaStream` object, but I am open to any suggestion.

charliesantos: Thanks for the details. I think I understand it better now.
With the approach above, it's a bit more specific to Citrix. My suggestion is to use listeners instead. Since these operations happen on `attached` and `detached`, we can emit events when these happens. This is the same pattern used if you take a look at the [AudioTrack](https://sdk.twilio.com/js/video/releases/2.29.0/docs/AudioTrack.html) and [VideoTrack](https://sdk.twilio.com/js/video/releases/2.29.0/docs/VideoTrack.html) classes.

Basically, add `attached` and `detached` events in `MediaTrack` class. That way, AudioTrack and VideoTrack inherits from it and the customer can listen to these events.

On the Custom MediaStream, I think it can be a simple  MediaStream override that way, we don't need to change much of our SDK implementation. You can remove the `createMediaStream` field. Here's a full example of what I'm thinking

```js

const room = await connect('token', {
  ...connectOptions,
  ...webRTCOverrides,
  // Pass custom MediaStream that has the same signature as the native MediaStream
  // This way, you don't need to check hasCreateMethod. You can just keep the new this._MediaStream() call in the SDK.
  MediaStream: function(streamTracks) {
    return window.CitrixWebRTC.createMediaStream(streamTracks);
  }
});

// Track related logic
const remoteVideoTrack = getOrCreateRemoteVideoTrack(); // Application specific logic
const remoteAudioTrack = getOrCreateRemoteAudioTrack(); // Application specific logic

remoteVideoTrack.on('attached', (el) => window.CitrixWebRTC.mapVideoElement(el));
remoteVideoTrack.on('detached', (el) => window.CitrixWebRTC.disposeVideoElement(el));

remoteAudioTrack.on('attached', (el) => window.CitrixWebRTC.mapAudioElement(el));
remoteAudioTrack.on('detached', (el) => window.CitrixWebRTC.disposeAudioElement(el));
```

luisrivas: Hi, @charliesantos! I really like this idea. Although it was not possible to implement all the methods as you proposed, I changed my last implementation to progressively align with your approach.

charliesantos (approved):

PR #2115, opened 2025-06-27
[VBLOCKS-4849] Fix race condition in LocalVideoTrack processor restart on certain devices

luisrivas:
### Description
This PR resolves the issue where `LocalVideoTrack.restart()` with a video processor fails on some devices due to race conditions between the restart process and processor restart operations. The fix ensures proper promise chaining so restart completes after all operations, maintaining processed video transmission to remote participants.

#### Changes
- Override `_restartProcessor()` method in `LocalVideoTrack` class
- Chain processor restart promise to main restart promise in `restart()` method
- Add try-catch error handling for processor restart operations

### Before review
* [x] Ready for review

### DIFF
--- a/lib/media/track/localvideotrack.js
+++ b/lib/media/track/localvideotrack.js
@@ -118,9 +118,16 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
    * const localVideoTrack = Array.from(room.localParticipant.videoTracks.values())[0].track;
    * localVideoTrack.addProcessor(new GrayScaleProcessor(100));
    */
-  addProcessor() {
+  addProcessor(processor, options) {
+    return this._addProcessor(processor, options);
+  }
+
+  /**
+ * @private
+ */
+  _addProcessor(processor, processorOptions) {
     this._log.debug('Adding VideoProcessor to the LocalVideoTrack');
-    const result = super.addProcessor.apply(this, arguments);
+    const result = super._addProcessor(processor, processorOptions);
 
     if (!this.processedTrack) {
       return this._log.warn('Unable to add a VideoProcessor to the LocalVideoTrack');
@@ -154,13 +161,21 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
    *
    * document.getElementById('remove-button').onclick = () => localVideoTrack.removeProcessor(grayScaleProcessor);
    */
-  removeProcessor() {
+  removeProcessor(processor) {
+    this._removeProcessor(processor);
+    return this;
+  }
+
+  /**
+   * @private
+   */
+  async _removeProcessor(processor) {
     this._log.debug('Removing VideoProcessor from the LocalVideoTrack');
-    const result = super.removeProcessor.apply(this, arguments);
+    const result = super._removeProcessor(processor);
 
     this._log.debug('Updating LocalVideoTrack\'s MediaStreamTrack with the original MediaStreamTrack');
-    this._setSenderMediaStreamTrack()
-      .then(() => this._updateElementsMediaStreamTrack());
+    await  this._setSenderMediaStreamTrack();
+    this._updateElementsMediaStreamTrack();
 
     return result;
   }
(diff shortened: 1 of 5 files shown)


charliesantos (changes requested):
This is unnecessarily complicated :sweat_smile: (it's not your fault). I understand it's because in our internal restarProcessor, we are calling public APIs removeProcessor and addProcessor. To make it a bit cleaner, we can create private methods and call those internally instead. Then the public APIs will call the internal methods. This requires a bit of plumbing but it's less confusing than the existing implementation. Here's what I'm thinking:

- In LocalVideoTrack, VideoTrack, and RemoteVideoTrack, create private methods
  - `_addProcessor` - implementation is the same as the existing `addProcessor`
  - `_removeProcessor` - implementation is the same as the existing `removeProcessor`, except in LocalVideoTrack, make sure it returns a promise and add await when calling _setSenderMediaStreamTrack.
  - `_restartProcessor` - calls the private `_removeProcessor` and `_addProcessor` instead. Make sure to 'await' `_removeProcessor`
- Update all internal usage of `addProcessor` and `removeProcessor` to make sure we use the private counterparts. There should be only a few of them.
- Now in LocalVideoTrack, VideoTrack, and RemoteVideoTrack, recreate the public methods `addProcessor` and `removeProcessor`. But internally, they only call the private counterparts.

luisrivas:
> This is unnecessarily complicated 😅 (it's not your fault). I understand it's because in our internal restarProcessor, we are calling public APIs removeProcessor and addProcessor. To make it a bit cleaner, we can create private methods and call those internally instead. Then the public APIs will call the internal methods. This requires a bit of plumbing but it's less confusing than the existing implementation. Here's what I'm thinking:
> 
> * In LocalVideoTrack, VideoTrack, and RemoteVideoTrack, create private methods
>   
>   * `_addProcessor` - implementation is the same as the existing `addProcessor`
>   * `_removeProcessor` - implementation is the same as the existing `removeProcessor`, except in LocalVideoTrack, make sure it returns a promise and add await when calling _setSenderMediaStreamTrack.
>   * `_restartProcessor` - calls the private `_removeProcessor` and `_addProcessor` instead. Make sure to 'await' `_removeProcessor`
> * Update all internal usage of `addProcessor` and `removeProcessor` to make sure we use the private counterparts. There should be only a few of them.
> * Now in LocalVideoTrack, VideoTrack, and RemoteVideoTrack, recreate the public methods `addProcessor` and `removeProcessor`. But internally, they only call the private counterparts.

I will explore this idea and tag you again for review. Thanks for the fresh perspective!

charliesantos:
> Hey, @charliesantos I updated the PR with your recommendations. However, it does not fix the issue. I still uploaded the changes, so you can confirm that is what you meant.

hmmm. I would think it would work since we're making removing and adding synchronous. For the purposes of time, I'm ok if we revert to your original implementation. We need to add comments in the code that describes why it's done that way since it's not immediately obvious the reason behind it.

luisrivas:
@charliesantos I made some adjustments to simplify the fix. Please let me know what you think!

lib/media/track/localvideotrack.js
```diff
@@ -261,6 +268,24 @@ class LocalVideoTrack extends LocalMediaVideoTrack {
...
+   *
+   * @returns {Promise<void>}
+   * @private
+   */
+  _restartProcessor() {
+    const processor = this.processor;
+    const processorOptions = Object.assign({}, this._processorOptions);
+    super.removeProcessor.apply(this, [processor]);
+    super.addProcessor.apply(this, [processor, processorOptions]);
+    return this._setSenderMediaStreamTrack(true);
```
charliesantos: Is this the same as the following (for simplicity)? Also, please add the comments below for clarity.
```suggestion
    // Local video track's removeProcessor calls setSenderMediaStreamTrack async
    // But we need to wait for that operation to finish for a successful restart
    // So we call the parent's restartProcessor, then setSenderMediaStreamTrack right after
    super._restartProcessor.call(this);
    return this._setSenderMediaStreamTrack(true);
```

luisrivas: Thanks, @charliesantos! I just added your comments. I kept the calls the same since they're not equivalent. Since we're overriding the `removeProcessor` and `addProcessor` methods in `LocalVideoTrack.

charliesantos: Ok, looking at it more, I understand. This is why I wanted to clean up the implementation a bit so it's more verbose and less confusing. But I understand the situation so +1 on the code changes. Should we add unit tests next? After that we can merge.

luisrivas: Yes, that's what I liked the most about your alternative. We could revisit this in the future, as I believe we can enhance the logic's readability by adding your suggestions and better managing the function's side effects. I just added unit tests that cover this new functionality. Again, thanks for all the reviews.

charliesantos (approved):
Thanks Luis.

PR #2123, opened 2025-07-15
[VBLOCKS-5005] feat: add Compute Pressure monitor

luisrivas:
### Description
- Created a simple computed pressure monitor based on the Compute Pressure API
- Added unit tests and an example usage of the designed API for review

### Before review
* [x] Ready for review

### DIFF
--- a/lib/insights/computepressuremonitor.js
+++ b/lib/insights/computepressuremonitor.js
@@ -1,0 +1,121 @@
+/**
+ * @fileOverview
+ * Custom Compute Pressure monitor using the Compute Pressure API.
+ *
+ * The Compute Pressure API is a JavaScript API that enables you to observe the pressure of system resources such as the CPU.
+ * For more information, see:
+ * {@link https://developer.mozilla.org/en-US/docs/Web/API/Compute_Pressure_API}
+ *
+ * This class is a simple implementation of a Compute Pressure monitor using the Compute Pressure API.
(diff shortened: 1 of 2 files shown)


luisrivas:
> Thanks Luis. Is the next step is to integrate this to the SDK?

Yes, I just wanted to make it easier to review.

lib/insights/computepressuremonitor.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+'use strict';
+
+/* globals PressureObserver */
+
+class ComputePressureMonitor {
+  _lastCpuPressure;
+  _cpuPressureObserver;
+  _sampleInterval = 10000; // 10 seconds
```
charliesantos, line 38: 10s might be too long to catch pressure changes. How about we change this to 2 seconds? Please note this is different than how often we send metrics which is every 10s. So to summarize, we observe every 2 seconds so we can catch pressure changes quickly, but we send CPU pressure information to insights every 10s.

luisrivas: Isn’t CPU pressure change its own event? I agree with changing the sample interval to 2 seconds, but since only changes will be sent, the delivery interval is not constant.

charliesantos: Not sure if I fully understand. I'll approve this for now and I like to see how you use this class in the SDK.

charliesantos (approved):

PR #2134, opened 2025-08-04
feat: add ApplicationEventPublisher for application lifecycle events

luisrivas:
### Description
- Publish the application lifecycle events to the room insights event reporting logic using an ApplicationEventPublisher.
- Reusing the DocumentVisibilityMonitor for consistency, and considering this monitor is already covered by our E2E tests. 

### Before review
* [x] Ready for review

### DIFF
--- a/lib/insights/applicationeventpublisher.js
+++ b/lib/insights/applicationeventpublisher.js
@@ -1,0 +1,103 @@
+'use strict';
+
+const documentVisibilityMonitor = require('../util/documentvisibilitymonitor');
+
+/**
+ * ApplicationEventPublisher handles publishing application lifecycle events to the Twilio Video SDK insights.
+ *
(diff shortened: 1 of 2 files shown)


lib/room.js
```diff
@@ -94,6 +95,11 @@ class Room extends EventEmitter {
           ? new ResourceEventPublisher(options.eventObserver, log)
           : null
       },
+      _applicationEventPublisher: {
```
kpchoy, line 98: nit: for ordering would this be better before `_clientTrackSwitchOffControl`, since this list is somewhat alphabetized?

luisrivas: Yeah, I do not see why not. We could do it later, if you'd like. For now, I prefer to put it next to the `ResourceEventPublisher` so you can have more context while reviewing it.

PR #2142, opened 2025-09-02
[VBLOCKS-5148] Publish track warning insights events

luisrivas:
### Description
This PR implements the `TrackWarningPublisher`. It consumes the `stats` information from the `periodicallyPublishStats` for now, but it should be able to `getStats` at a different pace once we define more details about it.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/insights/trackwarningpublisher.js
+++ b/lib/insights/trackwarningpublisher.js
@@ -1,0 +1,155 @@
+'use strict';
+
+/**
+ * TrackWarningPublisher handles publishing track warning events to the Twilio Video SDK insights.
+ * It monitors video track frame rates from WebRTC stats to detect stalled tracks.
+ *
+ * @example
+ * const trackWarningPublisher = new TrackWarningPublisher(eventObserver, log);
+ * trackWarningPublisher.processStats(stats);
+ *
+ * // Track warning events for stalled tracks are published in the following format:
+ * {
+ *   group: 'quality',
+ *   name: 'track-warning-raised',
+ *   level: 'warning',
+ *   payload: {
+ *     trackSID: string, // Track SID
+ *     frameRate: number, // Current frame rate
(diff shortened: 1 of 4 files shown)


lib/insights/trackwarningpublisher.js
```diff
@@ -0,0 +1,155 @@
...
+ * {
+ *   group: 'quality',
+ *   name: 'track-warning-cleared',
+ *   level: 'info',
+ *   payload: {
+ *     trackSID: string, // Track SID
+ *     frameRate: number, // Current frame rate
+ *     threshold: number // Frame rate threshold
+ *   }
+ * }
```
charliesantos: Per schema..

```suggestion
 *   group: 'track-warning-raised',
 *   name: 'track-stalled,
 *   level: 'warning',
 *   payload: {
 *     trackSID: string, // Track SID
 *     frameRate: number, // Current frame rate
 *     threshold: number // Frame rate threshold
 *   }
 * }
 *
 * // Track warning cleared events for resumed tracks are published in the following format:
 *
 * {
 *   group: 'track-warning-cleared',
 *   name: 'track-stalled',
 *   level: 'info',
 *   payload: {
 *     trackSID: string, // Track SID
 *     frameRate: number, // Current frame rate
 *     threshold: number // Frame rate threshold
 *   }
 * }
```

charliesantos: Let's also add 'trackType' since it's in the schema although it's value will be only 'video'.

charliesantos: Please update the implementation to reflect my comments above.

luisrivas: Good catch! Thank you, @charliesantos

charliesantos (approved):

PR #2148, opened 2025-09-15
fix: update network monitoring and freezeCount on stats handler

luisrivas:
### Description
- Fix the network monitoring logic to replace the `NetworkMonitor` dependency with a more straightforward alternative since `NetworkMonitor` relays on the connection's `type` property to detect changes, which is not supported in most browsers.
- Add the `freezeCount` property to the standardized `getStats` response.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/insights/networkeventpublisher.js
+++ b/lib/insights/networkeventpublisher.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const NetworkMonitor = require('../util/networkmonitor');
-
 /**
  * NetworkEventPublisher handles publishing network information events to the Twilio Video SDK insights.
  *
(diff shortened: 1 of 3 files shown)


lib/insights/networkeventpublisher.js
```diff
@@ -59,8 +57,19 @@ class NetworkEventPublisher {
     };
 
     try {
-      this._networkMonitor = new NetworkMonitor(networkChangeHandler);
-      this._networkMonitor.start();
```
charliesantos, line 63: Do you need to call this somewhere?

luisrivas: No, we don't

PR #2150, opened 2025-09-17
[VBLOCKS-5203] change: adjust the quality and track warning publishers stats handling

luisrivas:
### Description
* adjusts `QualityEventPublisher` and `TrackWarningPublisher` to process stats every second, update unit tests/JSDoc, and remove an unused import.
* increase CPU pressure sample interval and add validation for records handling

### Before review
* [x] Ready for review

### DIFF
--- a/lib/signaling/v2/room.js
+++ b/lib/signaling/v2/room.js
@@ -176,8 +176,7 @@ class RoomV2 extends RoomSignaling {
     handleLocalParticipantEvents(this, localParticipant);
     handlePeerConnectionEvents(this, peerConnectionManager);
     handleTransportEvents(this, transport);
-    periodicallyPublishStats(this, transport, options.statsPublishIntervalMs);
-    periodicallyCollectStats(this);
+    periodicallyCollectAndPublishStats(this, transport, options.statsPublishIntervalMs);
 
     this._update(initialState);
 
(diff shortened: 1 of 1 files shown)


lib/insights/computepressuremonitor.js
```diff
@@ -35,7 +35,7 @@
 class ComputePressureMonitor {
   _lastCpuPressure;
   _cpuPressureObserver;
-  _sampleInterval = 2000; // 2 seconds
+  _sampleInterval = 10000; // 10 seconds
```
charliesantos: Does this mean we're going to poll every 10s? If yes, this also means we cannot catch pressure changes within that 10 seconds?

lib/insights/networkeventpublisher.js
```diff
@@ -44,6 +44,9 @@ class NetworkEventPublisher {
    */
   _setupNetworkMonitoring() {
     const networkChangeHandler = () => {
+      if (typeof navigator === 'undefined' || !navigator.connection) {
+        return;
```
charliesantos, line 49: Can we log 'NetworkInformation not supported'.

lib/signaling/v2/room.js
```diff
@@ -172,6 +177,7 @@ class RoomV2 extends RoomSignaling {
     handlePeerConnectionEvents(this, peerConnectionManager);
     handleTransportEvents(this, transport);
     periodicallyPublishStats(this, transport, options.statsPublishIntervalMs);
+    periodicallyCollectStats(this);
```
charliesantos: I recall we already have something that publishes stats at a certain interval. Is this an addition to that? Does this mean we now have two polling mechanisms that reads stats?

charliesantos (changes requested):

lib/insights/computepressuremonitor.js
luisrivas: The API specification is not clear about it, but it mentions the following:

_In the case the value it set to 0, the system will only call the PressureUpdateCallback in the case there is a change to the PressureState._
[the sampleinterval member](https://w3c.github.io/compute-pressure/#the-sampleinterval-member)

What do you think?

lib/signaling/v2/room.js
luisrivas: Yes, that's the case at the moment

lib/insights/networkeventpublisher.js
luisrivas: Done!

lib/insights/computepressuremonitor.js
charliesantos: The examples uses 1 second. 10s seems a bit too long and we might missed some pressure changes. I think we should keep it at 2s.

lib/signaling/v2/room.js
charliesantos: I think we should combine the two. It seems unnecessary to call getStats twice (for each polling). We can do it on a separate ticket if you want.

luisrivas: Correct, that's for sure the next step

charliesantos: When do we plan to combine the two?

luisrivas: Let's do that optimization before sending the RC for QE, it shouldn't take much time. wdyt?

charliesantos: That sounds good!

lib/insights/computepressuremonitor.js
luisrivas: Done!

lib/signaling/v2/room.js
luisrivas: @charliesantos Please take a look at the last commit, I refactored the collect stats logic to have just one polling mechanism

PR #2169, opened 2025-10-24
[VBLOCKS-5544] telemetry: add DTLS transport state change event

luisrivas:
### Description

This PR adds the [RTCDtlsTransportState](https://w3c.github.io/webrtc-pc/#dom-rtcdtlstransportstate) change events.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/insights/events/peerconnection.js
+++ b/lib/insights/events/peerconnection.js
@@ -76,12 +76,7 @@ class PeerConnectionEvents {
    * @returns {void}
    */
   dtlsTransportState(peerConnectionId, state) {
-    const levelMap = {
-      failed: 'error',
-      closed: 'warning'
-    };
-    const level = levelMap[state] || 'debug';
-
+    const level = state === 'failed' ? 'error' : 'debug';
     this._telemetry[level]({
       group: 'dtls-transport-state',
       name: state,
--- a/test/unit/spec/insights/telemetry.js
+++ b/test/unit/spec/insights/telemetry.js
@@ -368,7 +368,7 @@ describe('Telemetry', () => {
       });
 
       describe('dtlsTransportState', () => {
-        ['new', 'connecting', 'connected'].forEach(state => {
+        ['new', 'connecting', 'connected', 'closed'].forEach(state => {
           it(`should publish ${state} event with debug level`, () => {
             telemetry.pc.dtlsTransportState('PC123', state);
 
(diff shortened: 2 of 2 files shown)


lib/signaling/v2/peerconnection.js
```diff
@@ -904,6 +908,20 @@ class PeerConnectionV2 extends StateMachine {
...
   }
 
+  /**
+   * Handle a DTLS transport state change event.
+   * @private
+   * @returns {void}
+   */
+  _handleDtlsTransportStateChange() {
+    const senders = this._peerConnection.getSenders();
+    const transport = senders[0]?.transport;
```
charliesantos: Will we get only one senders ever?

lib/signaling/v2/peerconnection.js
```diff
@@ -1837,6 +1855,29 @@ function updateRecycledTransceivers(pcv2) {
...
 }
 
+/**
+ * Update DTLS transport monitoring.
+ * @param {PeerConnectionV2} pcv2
+ * @returns {void}
+ */
+function updateDtlsTransport(pcv2) {
+  const senders = pcv2._peerConnection.getSenders();
+  const transport = senders[0]?.transport;
```
charliesantos: Will we get only one senders ever?

lib/insights/events/peerconnection.js
```diff
@@ -68,6 +68,26 @@ class PeerConnectionEvents {
...
+  /**
+   * Emit DTLS transport state change event
+   * @param {string} peerConnectionId - Peer connection identifier
+   * @param {('new'|'connecting'|'connected'|'closed'|'failed')} state - DTLS transport state
+   * @returns {void}
+   */
+  dtlsTransportState(peerConnectionId, state) {
+    const levelMap = {
+      failed: 'error',
+      closed: 'warning'
```
charliesantos: Closed shouldn't be a warning. We close this manually.

luisrivas: Do you prefer to use `debug` for the `closed` state instead? I took the `warning` level from the rtc-spec

lib/signaling/v2/peerconnection.js
luisrivas: No, we will typically have multiple senders, they just happen to share the same dtls transport because we are using `max-bundle` by default. Let me update this implementation just in case the users overwrite the `bundlePolicy`.

lib/signaling/v2/peerconnection.js
luisrivas: answered above

lib/insights/events/peerconnection.js
charliesantos: Yes I think it should be the same level as connecting, connected, etc. No need to change it now if it's already in the spec. We can change it later if you prefer since it requires another deployment right.

luisrivas: No worries, that part of the spec is not deployed yet. And also now that we are considering multiple dtls transports we may need to re-evaluate that payload as well.

lib/signaling/v2/peerconnection.js
luisrivas: @charliesantos I just updated the implementation to consider multiple sender. I also tested it with all the `bundlePolicy` configurations, and I confirmed the media server always bundle everything in one `transport` which is the recommended approach for our use case.

charliesantos (approved):

PR #2180, opened 2025-11-25
change: apply codec preferences via native browser API

luisrivas:
### Jira link(s)
- https://twilio-engineering.atlassian.net/browse/VBLOCKS-4994
- https://twilio-engineering.atlassian.net/browse/VBLOCKS-5049

### Description
Replace manual SDP manipulation for codec preferences with native `RTCRtpTransceiver.setCodecPreferences()` to align with modern practices for codec negotiation.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/webrtc/rtcpeerconnection/firefox.js
+++ b/lib/webrtc/rtcpeerconnection/firefox.js
@@ -27,6 +27,9 @@ const { delegateMethods, interceptEvent, legacyPromise, proxyProperties } = requ
 //     https://bugzilla.mozilla.org/show_bug.cgi?id=1240897
 //
 class FirefoxRTCPeerConnection extends EventTarget {
+  // NOTE(lrivas): Unlike the Chrome implementation, Firefox's RTCPeerConnection does not
+  // need a second argument for options and thus we only pass the configuration parameter
+  // keeping the signature consistent with the native RTCPeerConnection.
   constructor(configuration) {
     super();
 


kpchoy (approved):

lib/webrtc/rtcpeerconnection/safari.js
```diff
@@ -13,7 +13,7 @@ const updateTrackIdsToSSRCs = isUnifiedPlan
   : updatePlanBTrackIdsToSSRCs;
 
 class SafariRTCPeerConnection extends EventTarget {
-  constructor(configuration) {
+  constructor(configuration, options = {}) {
```
mhuynh5757, line 16: I see that the Chrome RTCPeerConnection has a default value for `configuration` where this Safari one does not.

lib/webrtc/rtcpeerconnection/chrome.js
```diff
@@ -23,9 +23,11 @@ const isUnifiedPlan = getSdpFormat() === 'unified';
 //   3. Set iceTransportPolicy.
 //
 class ChromeRTCPeerConnection extends EventTarget {
-  constructor(configuration = {}, constraints) {
+  constructor(configuration = {}, options = {}) {
```
charliesantos, line 26: This makes me sad since we're changing the default signature. But I understand it's necessary for this change.

lib/webrtc/rtcpeerconnection/safari.js
charliesantos, line 16: Do we need to change for firefox too?

charliesantos (commented):
Left a couple of comments. In my previous investigation, I saw addTransceiver calls inside lib/webrtc. See here https://github.com/search?q=repo%3Atwilio%2Ftwilio-video.js%20addTransceiver&type=code
How are we making sure setCodecPreferences is getting called whenever addTransceiver is called in these code paths?

lib/webrtc/rtcpeerconnection/safari.js
luisrivas: No, we don't need to apply this change for the Firefox wrapper. We could add it to keep things consistent, though.

luisrivas: Good catch, thanks @mhuynh5757! I'll create a ticket so we can look into why the SDK modifies the configuration object in Chrome and see if we can simplify or remove that behavior.

charliesantos: I'm ok with just a comment in the file.

charliesantos (approved):

lib/webrtc/rtcpeerconnection/safari.js
luisrivas: Perfect! I just added a comment clarifying it.

PR #2206, opened 2026-03-20
fix: remove unnecessary RegExp construction in enableDtxForOpus

luisrivas:
### Description

Simplify `enableDtxForOpus` by using `String.prototype.replace()` with a string pattern instead of constructing a `RegExp` at runtime.

### Before review
* [x] Ready for review


charliesantos (approved):
Thanks Luis.

PR #2224, opened 2026-04-30
feat: support @twilio/krisp-audio-plugin 2.x

luisrivas:
### Description

Accepts both `@twilio/krisp-audio-plugin` 1.x and 2.x as the Krisp noise cancellation SDK. Version is detected at load time; 1.x plugins are routed through a legacy adapter, 2.x plugins are used directly.

### Before review
* [x] Ready for review

### DIFF
--- a/lib/media/track/noisecancellationimpl.ts
+++ b/lib/media/track/noisecancellationimpl.ts
@@ -16,10 +16,18 @@ const Log = require('../../util/log');
  * const { connect, createLocalAudioTrack } = require('twilio-video');
  *
  * // Create a LocalAudioTrack with Krisp noise cancellation enabled.
+ * // Supported plugin versions: @twilio/krisp-audio-plugin 1.x or 2.x.
  * const localAudioTrack = await createLocalAudioTrack({
  *   noiseCancellationOptions: {
- *     sdkAssetsPath: 'path/to/hosted/twilio/krisp/audio/plugin/1.0.0/dist',
+ *     sdkAssetsPath: 'path/to/hosted/twilio/krisp/audio/plugin/{version}/dist',
  *     vendor: 'krisp'
+ *   },
+ *   // Required for @twilio/krisp-audio-plugin 2.x only, so Krisp receives
+ *   // the raw microphone signal. Omit these for 1.x.
+ *   audio: {
+ *     echoCancellation: false,
+ *     noiseSuppression: false,
+ *     autoGainControl: false
  *   }
  * });
  *
@@ -130,7 +138,7 @@ export class NoiseCancellationImpl implements NoiseCancellation {
     const track = await reacquire();
     this._sourceTrack = track;
 
-    const processedTrack = await this._processor.connect(track);
+    const processedTrack = this._processor.connect(track);
     if (processorWasEnabled) {
       this._processor.enable();
     } else {
--- a/lib/media/track/noisecancellationimpl.ts
+++ b/lib/media/track/noisecancellationimpl.ts
@@ -24,11 +24,9 @@ const Log = require('../../util/log');
  *   },
  *   // Required for @twilio/krisp-audio-plugin 2.x only, so Krisp receives
  *   // the raw microphone signal. Omit these for 1.x.
- *   audio: {
- *     echoCancellation: false,
- *     noiseSuppression: false,
- *     autoGainControl: false
- *   }
+ *   echoCancellation: false,
+ *   noiseSuppression: false,
+ *   autoGainControl: false
  * });
  *
  * if (!localAudioTrack.noiseCancellation) {
--- a/lib/noisecancellationadapter.ts
+++ b/lib/noisecancellationadapter.ts
@@ -1,65 +1,54 @@
 'use strict';
 
+import { LegacyPlugin, LegacyPluginAdapter } from './noisecancellation/legacy-plugin-adapter';
 import type { AudioProcessor } from '../tsdef/AudioProcessor';
 import type { NoiseCancellationOptions } from '../tsdef/types';
+import type { NoiseCancellationPlugin } from './noisecancellation/plugin';
 
 const dynamicImport = require('./util/dynamicimport');
 const Log = require('./util/log');
 
 const PLUGIN_CONFIG = {
   krisp: {
-    supportedVersion: '1.0.0',
+    supportedVersions: ['1.0.0', '2.0.0'],
     pluginFile: 'krispsdk.mjs'
   },
   rnnoise: {
-    supportedVersion: '0.6.0',
+    supportedVersions: ['0.6.0'],
     pluginFile: 'rnnoise_sdk.mjs'
   }
 };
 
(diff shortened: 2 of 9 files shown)


lib/media/track/noisecancellationimpl.ts
```diff
@@ -16,10 +16,18 @@ const Log = require('../../util/log');
...
+ * // Supported plugin versions: @twilio/krisp-audio-plugin 1.x or 2.x.
  * const localAudioTrack = await createLocalAudioTrack({
  *   noiseCancellationOptions: {
- *     sdkAssetsPath: 'path/to/hosted/twilio/krisp/audio/plugin/1.0.0/dist',
+ *     sdkAssetsPath: 'path/to/hosted/twilio/krisp/audio/plugin/{version}/dist',
  *     vendor: 'krisp'
+ *   },
+ *   // Required for @twilio/krisp-audio-plugin 2.x only, so Krisp receives
+ *   // the raw microphone signal. Omit these for 1.x.
+ *   audio: {
```
charliesantos: The `audio: { echoCancellation: false, ... }` block is under the wrong key for `createLocalAudioTrack`. Its option type is `CreateLocalAudioTrackOptions extends CreateLocalTrackOptions extends MediaTrackConstraints`, so those constraints go flat at the top level — there is no `audio` field here (that field belongs on `connect()` / `createLocalTracks()`). Customers copy/pasting this example will silently ship with browser defaults applied on top of Krisp's processing. Please match the shape used in the plugin README (flat `echoCancellation` / `noiseSuppression` / `autoGainControl` on the options object).

lib/media/track/noisecancellationimpl.ts
```diff
@@ -130,7 +138,7 @@ export class NoiseCancellationImpl implements NoiseCancellation {
     const track = await reacquire();
     this._sourceTrack = track;
 
-    const processedTrack = await this._processor.connect(track);
+    const processedTrack = this._processor.connect(track);
```
charliesantos, line 139: The RFC explicitly promotes `AudioProcessor.connect` to `Promise<MediaStreamTrack>`, but this PR leaves the signature sync and removes the pre-existing `await` here. Either the RFC is outdated (the v2 plugin does its async work in `init()` / `createNoiseFilter`, so sync connect is fine) or the signature needs to be promoted and this await restored. Whichever way, please get the RFC and code into agreement — customers writing custom `AudioProcessor`s will read the type definition, not the RFC.

lib/noisecancellationadapter.ts
```diff
@@ -1,65 +1,54 @@
...
   }
+};
 
-  const tempContext = new AudioContext();
-  const isSupported = plugin.isSupported(tempContext);
-  tempContext.close();
-  if (!isSupported) {
-    throw new Error('Noise Cancellation plugin is not supported on your browser');
-  }
+const adaptPlugin = (maybeLegacyPlugin: NoiseCancellationPlugin | LegacyPlugin, major: number): NoiseCancellationPlugin => {
```
charliesantos, line 50: `major >= 2` means a future 3.x plugin silently bypasses the legacy adapter with no structural validation, and the v2 branch does no runtime check that the cast shape is actually correct — most notably `isSupported` (the most-changed method) isn't guarded. A 2.x bundle missing the zero-arg `isSupported` would fail later with a confusing `TypeError`. Suggest `major === 2` here, plus asserting `typeof plugin.isSupported === 'function' && plugin.isSupported.length === 0` after the cast.

charliesantos (changes requested):
Reviewed together with twilio-internal/krisp-audio-plugin#8 and against the Krisp V1→V2 migration RFC. The LegacyPluginAdapter design is clean and the dual-major routing matches the RFC's intent. Two things I'd like resolved before merge: the jsdoc example in `noisecancellationimpl.ts` uses the wrong option shape for the audio constraints (flagged inline), and the RFC explicitly calls for `AudioProcessor.connect` to become `Promise<MediaStreamTrack>` but this PR keeps it sync and removes the pre-existing `await` — please pick a source of truth (RFC or code) and make them match (flagged inline).

Observability regression: the plugin version used to appear in SDK logs via `log.debug('Plugin Version = ...')` inside `ensureVersionSupported`, and the replacement `ensureVersionCompatible` no longer logs it. Now that two majors coexist, the runtime plugin version matters more, not less — please re-add a `log.debug('Plugin Version =', pluginVersion)` in `createNoiseCancellationAudioProcessor` after `getVersion()`.

`supportedVersions: ['1.0.0', '2.0.0']` with custom compare logic doesn't match the RFC's stated `^1.0.0 || ^2.0.0` semver-range framing. Behavior is close (patch releases of either major are accepted) but the shape is the rejected alternative the RFC calls out. Worth aligning the RFC wording or adopting a real semver range — not blocking.

No unit tests landed for `parseVersion`, `ensureVersionCompatible`, `adaptPlugin`, or `LegacyPluginAdapter`. The burndown item is unchecked. Version routing is the most brittle part of this change; a table-driven test of `parseVersion` (NaN, wrong part count, prerelease) plus branch tests for `adaptPlugin(_, 1)` vs `adaptPlugin(_, 2)` would be cheap and high-value. CHANGELOG entry is also unchecked — a one-liner for 2.x support is warranted even though `AudioProcessor` isn't re-exported from `tsdef/index.d.ts`.

Positives: replacing the raw `new AudioContext()` + `close()` dance with `AudioContextFactory.getOrCreate` / `release` inside `LegacyPluginAdapter.isSupported` is strictly better — the old code leaked if `close` threw. `parseVersion` now rejects NaN, which the pre-PR code didn't.

lib/media/track/noisecancellationimpl.ts
luisrivas: Correct, v2 does async work in `init()/createNoiseFilter`. I'll update the RFC.

lib/noisecancellationadapter.ts
luisrivas: `major === 2` applied. Regarding the assertion for `typeof plugin.isSupported === 'function' && plugin.isSupported.length === 0`, shape drift feels like a plugin-test concern.

lib/media/track/noisecancellationimpl.ts
luisrivas: Good catch! Fixed.

charliesantos (approved):
Thanks @luisrivas, approved. A few quick items before merging:
- Do you plan to add unit tests in this PR, or follow up separately?
- I’m assuming the changelog will come in a different PR correct?
- Please ping me once the RFC is updated (no need to block this PR on that).

luisrivas:
Thanks, @charliesantos.

Regarding your questions:
> * Do you plan to add unit tests in this PR, or follow up separately?

I can add the unit tests in this PR to consolidate it now that I got your feedback regarding the implementation.

> * I’m assuming the changelog will come in a different PR correct?

Correct!

> * Please ping me once the RFC is updated (no need to block this PR on that).

Done!

charliesantos (approved):

PR #2228, opened 2026-05-13
VBLOCKS-6627: add selectedEdge and configuredRegion to PreflightTestReport

donaltoomey:
Adds two new optional fields to PreflightTestReport to expose which Twilio edge handled the preflight connection and which region was configured by the caller. selectedEdge is derived from the TURN server hostname in the iced response; configuredRegion reflects the region option passed to runPreflight (defaulting to 'gll').

**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.

### Description

### Before review
* [ ] Ready for review


donaltoomey:
@charliesantos, @luisrivas, I'm planning to chip away at the Video Diagnostics App improvements in the background. This is the only PR for the JS SDK. Have a look when you get a chance.

lib/preflight/preflighttest.ts
```diff
@@ -539,6 +543,9 @@ function initCollectedStats() : PreflightStats {
  * @property {SelectedIceCandidatePairStats} selectedIceCandidatePairStats - Stats for the ice candidates that were used for the connection.
  * @property {Array<ProgressEvent>} [progressEvents] - {@link ProgressEvent} events detected during the test.
  * Use this information to determine which steps were completed and which ones were not.
+ * @property {string} [selectedEdge] - The Twilio TURN edge that handled the connection (e.g. 'ashburn', 'dublin').
+ * Derived from the TURN server hostname in the iced response. Undefined if the test did not complete successfully.
+ * @property {string} [configuredRegion] - The region option passed to {@link runPreflight} (defaults to 'gll').
```
charliesantos, line 548: `selectedEdge` and `configuredRegion` make sense. One thing though: `selectedEdge` has the opposite meaning in Voice. It represents the edge selected by the customer, while `edge` represents the edge that actually handled the connection. Can we rename these to align with the Voice SDK terminology?

https://twilio.github.io/twilio-voice.js/interfaces/PreflightTest.Report.html

charliesantos (changes requested):

lib/preflight/preflighttest.ts
donaltoomey: ok, I will change `selectedEdge` to to be `edge`. I need to keep `configuredRegion` as is because for video we don't have the concept of edges at all - the Video docs talk about regions, us1, ie1, etc.

PR #2230, opened 2026-06-11
ci: add GitHub Actions workflow for lint, build, and unit tests

luisrivas:
### Description

Adds a GitHub Actions workflow running lint, build, unit tests, and server-side-render tests.

### Before review
* [x] Ready for review

### DIFF
--- a/.github/actions/artifactory-oidc/action.yml
+++ b/.github/actions/artifactory-oidc/action.yml
@@ -1,0 +1,62 @@
+name: 'Artifactory OIDC Auth'
+description: 'Exchange GitHub OIDC token for Artifactory access token and configure npm'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Exchange OIDC token for Artifactory access token
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        # jq is used throughout; fail early with a clear message if it's absent
+        # (present on ubuntu-latest today, but not guaranteed on every runner).
+        if ! command -v jq >/dev/null 2>&1; then
+          echo "::error::jq is required by the artifactory-oidc action but was not found on the runner."
+          exit 1
+        fi
+
+        # Normalize the base URL: strip any trailing slashes so REGISTRY and the
+        # //host/... auth line never end up with double slashes.
+        BASE_URL=$(echo "${ARTIFACTORY_URL}" | sed -E 's#/+$##')
+        if [ -z "$BASE_URL" ]; then
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
+          -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${BASE_URL}" \
+          | jq -r '.value')
+
+        # 2. Exchange it for an Artifactory access token. The body MUST be JSON
+        #    and MUST include provider_name, or JFrog issues no token.
+        RESP=$(curl -sS "${BASE_URL}/access/api/v1/oidc/token" \
+          -H 'Content-Type: application/json' \
+          -d "{\"grant_type\":\"urn:ietf:params:oauth:grant-type:token-exchange\",
+               \"subject_token_type\":\"urn:ietf:params:oauth:token-type:id_token\",
+               \"subject_token\":\"${GITHUB_JWT}\",
+               \"provider_name\":\"github-actions\"}")
+        ART_TOKEN=$(echo "$RESP" | jq -r '.access_token // empty')
+
+        if [ -z "$ART_TOKEN" ]; then
+          echo "::error::OIDC token exchange failed. Redacted response below:"
+          # Redact every key whose name contains "token" (access_token,
+          # refresh_token, id_token, ...). Withhold entirely if not valid JSON —
+          # never print a raw, unredacted body.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+          - name: build
+            run: npm run build:quick
+          - name: unit-tests
+            run: npm run test:unit
+          - name: ssr-tests
+            run: |
+              npm run build:es5
+              npm run test:serversiderender
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
charliesantos, line 42: SHA pinned correctly (good), but the `# v6.0.3` comment is wrong — actions/checkout has no v6.x (current stable is v4). Correct the comment so the SHA-to-version audit trail is accurate.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Artifactory OIDC Auth
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: ./.github/actions/artifactory-oidc
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```
charliesantos, line 46: Same as the checkout pin — `# v6.4.0` is wrong; actions/setup-node's current major is v4. Fix the version comment to match the pinned SHA.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,62 @@
...
+        BASE_URL=$(echo "${ARTIFACTORY_URL}" | sed -E 's#/+$##')
+        if [ -z "$BASE_URL" ]; then
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
```
charliesantos, line 30: Consider masking the intermediate GitHub JWT too (`echo "::add-mask::$GITHUB_JWT"` right after assignment) — the Artifactory token is masked but the JWT would surface in logs if step-debug is on.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
```
charliesantos: Consider scoping `id-token: write` to the `check` job rather than workflow-level, so future jobs don't inherit it by default.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Artifactory OIDC Auth
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: ./.github/actions/artifactory-oidc
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
```
charliesantos: The `npm ci` and matrix run steps have no `name:`, so they show as generic labels in the Actions UI while every other step is named. Worth adding names for consistency.

charliesantos (commented):
Right approach — this adopts the official OSS build pattern (GHA OIDC → curated Artifactory), keyless, and the OIDC action's error handling is well done. Fork PRs are intentionally skipped (we don't run untrusted fork code on our runners), which is the correct posture. A few small items inline.

.github/actions/artifactory-oidc/action.yml
luisrivas: Good catch, addressed.

.github/workflows/ci.yml
luisrivas: Double-checked, this SHA is `v6.4.0`. Happy to re-verify together if you're seeing something different.

.github/workflows/ci.yml
luisrivas: Double-checked, this SHA is `v6.0.3`. Happy to re-verify together if you're seeing something different.

.github/workflows/ci.yml
luisrivas: Fixed, `id-token: write` moved to job-level.

.github/workflows/ci.yml
luisrivas: Fixed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+          - name: unit-tests
+            run: npm run test:unit
+          - name: ssr-tests
+            run: |
+              npm run build:es5
+              npm run test:serversiderender
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Artifactory OIDC Auth
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
kpchoy, line 44: Optional: this step-level fork guard is redundant — the `check` job already carries the identical `if: ${{ !github.event.pull_request.head.repo.fork }}` at job level, so this step only ever runs on non-fork PRs anyway. Safe to drop for clarity (harmless as-is).

kpchoy (approved):

charliesantos (approved):
@luisrivas you're right on both version comments. I checked the tags against the pinned SHAs and they both match what you pinned: checkout@v6.0.3 dereferences to df4cb1c0, and setup-node@v6.4.0 resolves to 48b55a01.

My review was based on the stale assumption that v6 didn't exist yet. Thanks for pushing back and verifying.

The other three comments (JWT mask, job-level id-token, and step names) still look good.

PR #2231, opened 2026-06-12
[VBLOCKS-6686] fix: pass module path as an argument to the import function

luisrivas:
### Description

Move the module path out of the function source and into a `url` argument.

### Before review
* [x] Ready for review


charliesantos (approved):
Thanks Luis. I tested and it works.

PR #2232, opened 2026-07-17
Add end-to-end tests running in GitHub Actions

luisrivas:
### Description

Adds `.github/workflows/e2e.yml`, running the integration suite (`test:integration`) in GitHub Actions across stage/prod, both Room topologies, Chrome and Firefox, and each browser's stable/beta/unstable channels.

`test/lib/token.js` and `test/lib/rest.js` are refactored to route through a new `test/lib/vendor.js` client so the same test code works unchanged in the new workflow.

### Before review
* [x] Ready for review


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,93 @@
...
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        uses: ./.github/actions/artifactory-oidc
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
```
charliesantos: Add `--ignore-scripts` here (plus the browser-download skip env if any dep pulls binaries), matching the inc-18715 hardening we set on voice.js - it closes the install-script execution vector the incident used. If the e2e build genuinely needs lifecycle scripts, drop a comment stating that the curated-Artifactory-only + container posture is the accepted control instead. Either way, make the choice explicit.

karma/makeconf.js
```diff
@@ -15,6 +15,43 @@ function getTestFiles(config, defaultFile) {
...
+  const requestUrl = process.env.ACTIONS_ID_TOKEN_REQUEST_URL;
+  if (!requestToken || !requestUrl) {
+    return process.env.VENDOR_TOKEN;
+  }
+  const audience = encodeURIComponent(process.env.VENDOR_AUDIENCE);
+  const response = await fetch(`${requestUrl}&audience=${audience}`, {
+    headers: { Authorization: `bearer ${requestToken}` }
+  });
+  const body = await response.json();
+  return body.value;
```
charliesantos, line 39: Check `response.ok` before reading `body.value`. On a 4xx/5xx from the OIDC endpoint this returns `undefined`, the middleware responds 200 with `{token: undefined}`, and the browser side throws "returned no token" - the real HTTP status is lost. Fails closed, so it's a diagnosability bug, not a security one, but add `if (!response.ok) throw new Error('OIDC token request failed: ' + response.status)`.

test/lib/vendor.js
```diff
@@ -0,0 +1,83 @@
...
+    throw new Error('callVendor: /mint-vendor-token returned no token');
+  }
+  return token;
+}
+
+/**
+ * @returns {Promise<string>} a vendor OIDC token no older than TOKEN_MAX_AGE_MS
+ */
+async function getVendorToken() {
+  if (!cachedToken || Date.now() - cachedTokenMintedAt > TOKEN_MAX_AGE_MS) {
```
charliesantos: Guard concurrent callers with a shared in-flight promise. Two simultaneous `callVendor` calls both read `cachedToken === null` and both mint, doubling OIDC requests on the first test of every spec. Store the promise: `if (!cachedTokenPromise || stale) { cachedTokenPromise = mintVendorToken().then(t => { cachedToken = t; cachedTokenMintedAt = Date.now(); return t; }); } return cachedTokenPromise;`

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo "token=$TOKEN" >> "$GITHUB_OUTPUT"
+
+      - name: Run e2e integration tests
+        env:
+          VENDOR_URL: ${{ matrix.environment == 'stage' && vars.VENDOR_URL_STAGE || vars.VENDOR_URL_PROD }}
+          VENDOR_AUDIENCE: ${{ matrix.environment == 'stage' && vars.VENDOR_URL_STAGE || vars.VENDOR_URL_PROD }}
+          VENDOR_TOKEN: ${{ steps.mint-vendor-token.outputs.token }}
+          ENVIRONMENT: ${{ matrix.environment }}
+          TOPOLOGY: ${{ matrix.topology }}
+          TEST_STABILITY: stable
```
charliesantos, line 93: `TEST_STABILITY` is hardcoded to `stable` while the matrix runs `bver: [stable, beta, unstable]`, so the beta and unstable cells run the same test set at 2x the CI cost with no added coverage. Set `TEST_STABILITY: ${{ matrix.bver }}`, or document why stable-only across all channels is intended.

test/lib/vendor.js
```diff
@@ -0,0 +1,83 @@
...
+async function callVendor(action, params) {
+  const vendorUrl = process.env.VENDOR_URL;
+
+  if (!vendorUrl) {
+    throw new Error('callVendor: VENDOR_URL is not set');
+  }
+
+  const vendorToken = await getVendorToken();
+
+  const body = JSON.stringify(Object.assign({ action, environment: defaults.environment }, params));
```
charliesantos: The `environment: defaults.environment` field is dead - the Function ignores the request body's environment and selects the tier from its own deployment config (by which URL you POST to). Worth dropping the field or noting it's ignored, so it doesn't imply the caller can pick the tier.

test/integration/spec/connect.js
```diff
@@ -71,14 +73,16 @@ describe('connect', function() {
       { grant: null },
       20106
     ]
-  ].forEach(([description, extraOptions, expectedCode]) => {
-    describe(`called ${description}`, () => {
+  ].forEach(([description, extraOptions, expectedCode, skip]) => {
+    // NOTE(lrivas): getToken() can't mint an invalid token anymore. Skipped
+    // until we decide how to reproduce that scenario again.
+    (skip ? describe.skip : describe)(`called ${description}`, () => {
```
charliesantos, line 79: These skip the 20107 (invalid API Key secret) and 20103 (invalid API Key SID) auth-failure cases with no ticket. Worth filing a follow-up - a `mint-invalid-token` vendor action or a vendor-side unit test could restore the coverage - rather than leaving it in open-ended limbo.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,93 @@
...
+      matrix:
+        environment: [stage, prod]
+        topology: [group, peer-to-peer]
+        browser: [chrome, firefox]
+        bver: [stable, beta, unstable]
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Artifactory OIDC Auth
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
charliesantos: This step-level fork guard is redundant - the job-level `if` already blocks the whole job for forks. Worth removing it and adding a comment near the job-level `if` noting it covers all credential-accessing steps, so it doesn't read inconsistently against the mint step which has no step guard.

karma/makeconf.js
```diff
@@ -15,6 +15,43 @@ function getTestFiles(config, defaultFile) {
...
+}
+
+function createMintVendorTokenMiddleware() {
+  return function(request, response, next) {
+    if (request.url !== '/mint-vendor-token') {
```
charliesantos, line 44: The `/mint-vendor-token` endpoint is unauthenticated. Fine while karma is localhost-bound in CI, but worth a comment pinning that localhost assumption - if karma is ever bound to a non-localhost host (remote device testing), any local process could pull a GitHub OIDC token.

test/lib/vendor.js
```diff
@@ -0,0 +1,83 @@
...
+let cachedToken = null;
+let cachedTokenMintedAt = 0;
+
+/**
+ * Fetch a fresh vendor OIDC token from the local karma middleware, which mints
+ * it server-side (where the runner's OIDC request credentials are available).
+ * @returns {Promise<string>}
+ */
+async function mintVendorToken() {
+  const response = await fetch('/mint-vendor-token');
```
charliesantos, line 23: `fetch('/mint-vendor-token')` with a relative URL is browser/karma-only - it throws under Node's built-in fetch (requires absolute URLs). The unit tests stub `global.fetch` so they're fine. Worth a one-line note that vendor.js is browser-only, or make the mint URL configurable.

test/unit/spec/vendor.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  beforeEach(() => {
+    fetchStub = sinon.stub(global, 'fetch');
+    fetchStub.withArgs('/mint-vendor-token').resolves(fakeResponse(200, JSON.stringify({ token: 'fake-oidc-token' })));
+
+    delete require.cache[require.resolve('../../lib/vendor')];
+    delete require.cache[require.resolve('../../env')];
+    process.env.VENDOR_URL = 'https://vendor.example.test/vend';
+    process.env.ACCOUNT_SID = 'ACxxx';
```
charliesantos, line 26: These `ACCOUNT_SID` / `API_KEY_SID` env assignments (and their afterEach deletes) are vestiges of the old raw-key approach - vendor.js never reads them. Worth removing so they don't imply vendor.js still depends on raw API credentials.

test/lib/vendor.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+const defaults = require('./defaults');
+
+// GitHub Actions OIDC tokens are valid for only ~5 minutes; re-mint if the cached
+// one is older than this so a multi-minute test run never signs requests with a
+// stale token.
+const TOKEN_MAX_AGE_MS = 4 * 60 * 1000;
```
charliesantos, line 9: Consider noting the intentional buffer in the comment: re-mint at 4 min gives a 1-min margin before GitHub's ~5-min OIDC expiry.

test/integration/spec/connect.js
```diff
@@ -536,7 +540,8 @@ describe('connect', function() {
       if (rooms) {
         rooms.forEach(room => room.disconnect());
       }
-      sid = sid || (rooms && rooms[0] && rooms[0].sid);
+      // Rooms auto-created by connect() have emptyRoomTimeout 0, so they're already
```
charliesantos, line 543: The cleanup comment is accurate but the removed `rooms[0].sid` fallback is actually unnecessary here because the room is always REST-created (sid always set) - consider noting that too.

charliesantos (changes requested):
Solid PR - moving credential-vending to an OIDC-gated Function removes raw API keys from the test env, and the async getToken refactor is clean. No hard blockers. One thing to settle before I approve: whether the missing `--ignore-scripts` on `npm ci` is intentional (see the install-step comment).

One cross-cutting note that doesn't anchor to a line: the first run will 401 unless each tier's Function deployment is configured to match this workflow - `GITHUB_OIDC_AUDIENCE` equal to that tier's VENDOR_URL, `GITHUB_REPOSITORY_ID` for twilio-video.js, `GITHUB_WORKFLOW_REF` for this file, `GITHUB_ALLOWED_REFS` covering the refs it runs on - plus the repo `vars` (VENDOR_URL_STAGE/PROD, ARTIFACTORY_URL) and the repo added to Artifactory OIDC trusted_builds. Failure is fail-closed (safe), but worth a merge checklist so a first-run failure reads as config, not code.

(Three items from my local pass were resolved on review and aren't raised here: the negative-TTL expired-token case is genuinely expired at the Function, the module-load ENVIRONMENT concern doesn't affect credential selection since the Function ignores the request body's environment and is tier-bound by URL, and the vendor action contract is complete.)

.github/workflows/e2e.yml
luisrivas: Done. Thanks!

karma/makeconf.js
luisrivas: Fixed

test/lib/vendor.js
luisrivas: Fixed

test/lib/vendor.js
luisrivas: Fixed

karma/makeconf.js
luisrivas: Fixed

test/lib/vendor.js
luisrivas: Done

test/unit/spec/vendor.js
luisrivas: Done

test/lib/vendor.js
luisrivas: Done

.github/workflows/e2e.yml
luisrivas: `TEST_STABILITY` and `bver` are independent. `TEST_STABILITY` is used to filter unstable tests and `bver` sets the browser channel.

test/integration/spec/connect.js
luisrivas: `sid` is only set inside `if (withName)`. Since `withName` is `false` in half of this block's cases, the room isn't always REST-created. That's why I had to add the comment.

charliesantos (approved):

PR #2234, opened 2026-08-12
ci: add prepare-release and release workflows

luisrivas:
### Description

Adds the two-workflow release path: `prepare-release.yml` produces the tagged, built commit; `release.yml` is dispatched manually against that tag and publishes it to npm.

The split keeps write access and publishing credentials in jobs that install no dependencies.

Includes a fix to scripts/docs.js, which exited 0 when jsdoc failed and left an empty dist/docs behind.

### Before review
* [x] Ready for review

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -90,21 +90,18 @@ jobs:
 
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
-      # sort -V is not strict semver: it orders 2.36.0-rc.1 after 2.36.0. It still
-      # catches the transposed or stale number this guards against.
+      # Catches the transposed or stale number — this repo once released 2.8.2-rc1
+      # in place of 2.28.2-rc1. Runs before the install, so the script it calls
+      # takes no dependencies.
       - name: Verify the release version moves forward
         env:
           RELEASE_VERSION: ${{ inputs.release_version }}
         run: |
           set -euo pipefail
 
-          CURRENT_VERSION="$(node -p "require('./package.json').version")"
-          HIGHEST_VERSION="$(printf '%s\n%s\n' "$CURRENT_VERSION" "$RELEASE_VERSION" | sort -V | tail -n1)"
-
-          if [ "$RELEASE_VERSION" = "$CURRENT_VERSION" ] || [ "$HIGHEST_VERSION" != "$RELEASE_VERSION" ]; then
-            echo "::error::${RELEASE_VERSION} is not greater than the current version ${CURRENT_VERSION}."
-            exit 1
-          fi
+          node ./scripts/check-version-bump.js \
+            "$(node -p "require('./package.json').version")" \
+            "$RELEASE_VERSION"
 
       - name: Artifactory OIDC Auth
         uses: ./.github/actions/artifactory-oidc
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -61,8 +61,18 @@ jobs:
         env:
           RELEASE_VERSION: ${{ inputs.release_version }}
           DEVELOPMENT_VERSION: ${{ inputs.development_version }}
+          DRY_RUN: ${{ github.event.inputs.dry_run }}
         run: |
           set -euo pipefail
+
+          # The dry-run guards match on 'true', so any other string releases for real.
+          case "${DRY_RUN:-false}" in
+            true|false) ;;
+            *)
+              echo "::error::dry_run must be exactly true or false."
+              exit 1
+              ;;
+          esac
 
           SEMVER='^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.]+)?$'
           for INPUT_VERSION in "$RELEASE_VERSION" "${DEVELOPMENT_VERSION:-}"
@@ -234,10 +244,46 @@ jobs:
             exit 1
           fi
 
+      # The download overwrites but never deletes, so stale files from a build
+      # still tracked from a previous release would be staged alongside this one.
+      - name: Clear any previously committed build output
+        run: |
+          set -euo pipefail
+
+          git rm -r --quiet --ignore-unmatch --cached dist es5
+          rm -rf dist es5
+
       - name: Download release tree
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: release-tree
+
+      # The download above overwrites package.json, so check it before tagging.
+      - name: Verify the downloaded tree is the expected package
+        run: |
+          set -euo pipefail
+
+          PACKAGE_NAME="$(node -p "require('./package.json').name")"
+          if [ "$PACKAGE_NAME" != twilio-video ]; then
+            echo "::error::downloaded package.json is for ${PACKAGE_NAME}, not twilio-video."
+            exit 1
+          fi
+
+          PACKAGE_VERSION="$(node -p "require('./package.json').version")"
+          if [ "$PACKAGE_VERSION" != "$RELEASE_VERSION" ]; then
+            echo "::error::downloaded package.json is ${PACKAGE_VERSION}, not ${RELEASE_VERSION}."
+            exit 1
+          fi
+
+          node -e '
+            const scripts = require("./package.json").scripts || {};
+            for (const name of ["install", "preinstall", "postinstall", "prepare", "prepublishOnly"]) {
+              if (scripts[name]) {
+                console.error("::error::downloaded package.json has an unexpected lifecycle script: " + name);
+                process.exit(1);
+              }
+            }
+          '
 
       # Matches the author on the existing release commits.
       - name: Configure git identity
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,223 @@
+name: Prepare Release
+
+# Builds a release: bumps the version, builds the SDK, commits the built tree,
+# and pushes a version tag.
+#
+# Dispatch this FROM the branch you are releasing (normally `develop`) — pick it
+# in the "Run workflow" dropdown. The commits and the tag are pushed back to
+# whatever ref the run was dispatched from, so the branch selected there is the
+# branch that gets the release. Any branch can be released, including a
+# maintenance branch.
+#
+# This workflow does not publish. Pushing the tag is what triggers release.yml,
+# and only a final X.Y.Z tag matches its filter — so a prerelease version stops
+# here.
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Version to release (e.g. 2.36.0 or 2.36.0-rc.1)"
+        required: true
+      development_version:
+        description: "Optional: next dev version to commit after the release (e.g. 2.36.1-dev). Leave blank to skip."
+        required: false
+      dry_run:
+        description: "Build and pack only — commit nothing, tag nothing, publish nothing."
+        required: false
+        type: boolean
+        default: false
+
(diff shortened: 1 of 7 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,223 @@
...
+# here.
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Version to release (e.g. 2.36.0 or 2.36.0-rc.1)"
+        required: true
+      development_version:
+        description: "Optional: next dev version to commit after the release (e.g. 2.36.1-dev). Leave blank to skip."
```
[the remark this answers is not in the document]
luisrivas: Added, as the first step. No post-bump check needed: with aliases rejected, npm version can only produce the literal string.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,223 @@
...
+      # the built tree again, returning the branch to a clean working state.
+      # Pushed after the tag, so the tag keeps pointing at the built commit.
+      - name: Commit the development version
+        if: ${{ inputs.development_version != '' }}
+        env:
+          DEVELOPMENT_VERSION: ${{ inputs.development_version }}
+        run: |
+          set -euo pipefail
+
+          npm version --no-git-tag-version "$DEVELOPMENT_VERSION"
```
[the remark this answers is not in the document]
luisrivas: Good catch. Added `--ignore-scripts` to both npm version calls.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,223 @@
...
+      # instead of relying on `HEAD` inferring one.
+      #
+      # Branch first, then tag, so the tag is never briefly unreachable. A final
+      # X.Y.Z tag landing here is what triggers release.yml.
+      - name: Push the release commit and tag
+        run: |
+          set -euo pipefail
+
+          git push origin "HEAD:${GITHUB_REF}"
+          git push origin "refs/tags/${RELEASE_VERSION}"
```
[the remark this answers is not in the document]
luisrivas: Fixed. `release.yml` now runs on workflow_dispatch from the release tag, with a guard step replacing the tag-pattern filter. `GITHUB_REF` is still `refs/tags/X.Y.Z`, so tag validation and provenance are unaffected.

.github/workflows/release.yml
```diff
@@ -0,0 +1,97 @@
...
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          # npm >= 11.5.1, bundled with Node 24, is required for OIDC trusted
+          # publishing.
+          node-version: 24
+
+      # Used over a plain `npm publish` for two guarantees: the tag must match
+      # package.json's version, and the publish runs with --ignore-scripts.
+      # tag-prefix is empty because this repo's tags are bare versions (1.2.3).
+      - name: Validate tag and publish to npm
```
kpchoy, line 109: **The publish job never verifies the tag carries a built tree.**

`prepare-release.yml:167-182` has a thorough "Verify the release tree was built" step, but it lives in the job that cannot cause harm, and there is no equivalent in the job that can. The reachability check above constrains *where the tag came from*, not *what is in it*.

Why that matters here specifically: `package.json` has `main: ./es5/index.js` and no `files` field, so tarball contents are decided entirely by `.npmignore`, which excludes `dist/docs` but not `es5/` or `dist/`. A tag missing `es5/` therefore publishes successfully and every `require('twilio-video')` in every consumer resolves to nothing. `validate-tag: 'true'` only asserts the tag equals `package.json`'s version, which says nothing about whether the build ran.

`--atomic` narrowed the window a lot, since a failed prepare-release run no longer leaves a stray tag behind. What is still reachable is a hand-cut tag on `master` or `develop`, or any pre-existing `X.Y.Z` tag. And with `provenance: 'true'` the broken artifact ships signed and attested, on an operation that cannot be undone.

Suggest mirroring the check that already exists, right after the reachability step:

```yaml
      - name: Verify the tag carries a built tree
        run: |
          set -euo pipefail

          for RELEASE_FILE in dist/twilio-video.js dist/twilio-video.min.js
          do
            if [ ! -s "$RELEASE_FILE" ]; then
              echo "::error::${RELEASE_FILE} is missing or empty at ${GITHUB_REF_NAME}; this tag was not built by prepare-release.yml."
              exit 1
            fi
          done

          if [ ! -s "$(node -p "require('./package.json').main")" ]; then
            echo "::error::package.json main does not resolve to a file in this tree."
            exit 1
          fi
```

The `main` check covers `es5/` without needing a second loop, and it stays correct if `main` ever moves.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,314 @@
...
+
+          if git ls-remote --exit-code --tags origin "refs/tags/${RELEASE_VERSION}" >/dev/null 2>&1; then
+            echo "::error::tag ${RELEASE_VERSION} already exists on origin."
+            exit 1
+          fi
+
+      - name: Download release tree
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: release-tree
```
kpchoy, line 259: **The `contents: write` job commits an artifact it never validates.**

The split puts the credentials in the right place, but the privileged job's entire payload comes from the unprivileged one, and nothing checks it on the way across.

`--ignore-scripts` blocks lifecycle scripts only. `build:quick` still runs eslint, `scripts/docs.js` into jsdoc, `tsc`, the rollup plugin chain, and mocha/nyc/ts-node, all from the full dev dependency tree. Any of that can rewrite the tree before the upload at `:195`. Because `dist/` and `es5/` are gitignored they do not exist in this job's checkout at all, so 100% of the committed build output originates in the job that ran third-party code, and `download-artifact` also overwrites the `package.json` that `:261` stages.

The artifact grew this round: `README.md` and `CHANGELOG.md` are now uploaded and staged too, so the surface is a little wider than it was.

Two cheap mitigations that keep the current structure:

1. Re-derive the metadata locally instead of importing it. `setup-node` is already conditionally in this job, so `npm version --no-git-tag-version --ignore-scripts "$RELEASE_VERSION"` here reproduces `package.json` and `package-lock.json` from the checkout, and they can come out of the artifact.
2. Assert the shape of what arrives, after the download and before the commit:

```yaml
      - name: Assert the downloaded tree is sane
        run: |
          set -euo pipefail

          test "$(node -p "require('./package.json').name")" = twilio-video
          test "$(node -p "require('./package.json').version")" = "$RELEASE_VERSION"
          node -e '
            const scripts = require("./package.json").scripts || {};
            for (const name of ["install", "preinstall", "postinstall", "prepare", "prepublishOnly"]) {
              if (scripts[name]) {
                console.error("::error::unexpected lifecycle script: " + name);
                process.exit(1);
              }
            }
          '
```

The lifecycle-script assertion is the one that matters most: an injected `postinstall` in the shipped `package.json` runs on every consumer's machine, and `--ignore-scripts` at publish time does not prevent that.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,314 @@
...
+      #
+      # --no-verify because pre-commit.sh rejects commits on release branches.
+      # Hooks aren't installed on a fresh checkout, so this only matters if these
+      # steps are ever run by hand.
+      - name: Commit the release
+        run: |
+          set -euo pipefail
+
+          git add package.json package-lock.json README.md CHANGELOG.md
+          git add -f dist es5
```
kpchoy, line 308: **Stale tracked `dist`/`es5` from a previous release get re-committed and published.**

`development_version` is documented as optional ("Leave blank to skip"). When it is skipped, the branch is left with `dist/` and `es5/` tracked, since the step that runs `git rm -r --ignore-unmatch dist es5` never fires. On the *next* release from that branch:

1. `actions/checkout` at `:224` restores every tracked file, including the previous release's `es5/*.js` and `dist/docs/*.html`.
2. `download-artifact` unzips the new build *over* them. Unzip overwrites what is in the artifact; it does not delete what is not.
3. `git add -f dist es5` stages the union. A module deleted between the two versions, or a renamed class's doc page, is still on disk from step 1 and still gets staged.
4. That ships. `.npmignore` excludes only `dist/docs`, so dead `es5/*.js` from the previous version goes into the tarball next to the current build.

This is a guard that existed in the process being replaced and was lost in the port. `.release.json`'s `release` plan ran `git rm -rf --ignore-unmatch dist es5` *before* `npm run build:quick`, in one workspace, so the tree could only contain what the build had just produced. Splitting build and commit into two jobs broke that invariant, and nothing re-establishes it.

Restoring it is a step between the checkout at `:224` and the download at `:237`:

```yaml
      - name: Clear any previously committed build output
        run: |
          set -euo pipefail

          git rm -r --quiet --ignore-unmatch --cached dist es5
          rm -rf dist es5
```

Then the artifact extracts into empty directories and `git add -f dist es5` stages exactly the build, deletions included.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,314 @@
...
+            if [ -z "$(ls -A "$RELEASE_DIR" 2>/dev/null)" ]; then
+              echo "::error::${RELEASE_DIR}/ is missing or empty after the build."
+              exit 1
+            fi
+          done
+
+      # A publish cannot be undone: this proves the tree builds and shows exactly
+      # what would ship, without writing anything anywhere.
+      - name: Dry run — pack and preview the publish
+        if: ${{ github.event.inputs.dry_run == 'true' }}
```
kpchoy, line 197: **This guard now fails toward a real release.**

Good catch on the underlying bug: `inputs.dry_run` really is a boolean from the UI and a truthy string from the dispatch API, so `!inputs.dry_run` skipped the upload on every API dispatch regardless of intent. Reading `github.event.inputs.dry_run` is the right fix.

The pairing is what I'd change. `== 'true'` here and `!= 'true'` at `:196` and `:215` are exact complements, so one branch always runs, but anything other than the literal `true` takes the *destructive* one. A dispatch carrying `TRUE`, `1`, or `yes` skips the pack-and-preview and goes straight to upload plus `commit-tag-push`, producing a real tagged release for an operator who asked for a dry run. GitHub does not strictly validate boolean-typed inputs arriving through the REST API, which is the same gap the fix is working around. Note this direction is a regression from the old form, where a stray truthy string erred toward doing nothing.

Suggest validating the value once alongside the version inputs, then gating the destructive path positively:

```yaml
          case "$DRY_RUN" in
            true|false) ;;
            *)
              echo "::error::dry_run must be exactly true or false, got ${DRY_RUN}."
              exit 1
              ;;
          esac
```

with `DRY_RUN: ${{ github.event.inputs.dry_run }}` in that step's `env:`, and `:196` / `:215` changed to `== 'false'`. That way an unrecognized value fails loudly in the first ten seconds instead of quietly releasing.

kpchoy (commented):
Really nice iteration on this. The `--atomic` push with the `ls-remote` pre-flight, the origin-side reachability check, and the hand-rolled comparator with tests drawn from real history all landed well, and catching the `inputs` vs `github.event.inputs` boolean coercion yourself was a good spot.

.github/workflows/prepare-release.yml
luisrivas: Thanks, fixed

.github/workflows/prepare-release.yml
luisrivas: Fixed

.github/workflows/prepare-release.yml
luisrivas: Fixed

.github/workflows/release.yml
luisrivas: Fixed

kpchoy (approved):

charliesantos (approved):
Great work @luisrivas !

